Issue #97, opened 2026-07-10
Rate-limited engine is invisible: card stays 'working', sends look delivered, flows stall in waiting_ready — detect, surface, and offer successor reseat


Latand:
## Observed (2026-07-10 ~18:00, prod)

Three Codex implementer sessions hit the account usage limit ("You've hit your usage limit… try again at 7:55 PM" in the pane). The viewer showed **no sign of it anywhere**:

- the conversation card kept the green "Імплементер працює" state;
- `/api/tmux` sends reported `delivered-to-live` while the engine could not actually take a turn;
- three attached review flows sat in `waiting_ready` indefinitely with no attention raised;
- the operator discovered it only by manually reading the pane screen.

Screenshot in operator archive; panes: agents:3.0/5.0/6.0.

## Expected

1. **Detect** a live rate-limited engine per conversation: a current (not historical — see #56 for the inverse false-positive) usage-limit banner in the active pane, or the structured signal once #25 runtimes land (Codex app-server `account/rateLimits/read` already exposes windows; the limits panel knew the account was near exhaustion).
2. **Surface** it: card/board badge ("rate-limited until 19:55"), attention state, and flows attached to that conversation should show `blocked: rate-limited`, not a silent `waiting_ready`.
3. **Assist the reseat**: one-click (or policy-driven auto) "continue on account X" — spawn a successor in the same cwd with a handoff prompt, rebind attached flows' `implementerPath`, park the old conversation quietly (do NOT enqueue it as a fresh manual card — same class as #86). Careful not to collide with account-migration (#40): if a migration already forked the thread, the old conversation must not be reseated again — check registry lineage first.

## Refs

#56 (historical rate-limit text false positive — inverse problem), #25 (structured rate-limit reads), #40/#86 (migration lineage), #12 (delivery verification).

Today's manual workaround that should become the product flow: spawned 3 successor implementers on the healthy account via /api/spawn accountId, edited flows.json implementerPath, cycle resumed.

Latand:
Implemented in https://github.com/Latand/live-log-viewer-next/pull/99

PR #99, opened 2026-07-10
Surface rate-limited conversations and blocked flows

Latand:
## Summary\n\n- detect current pane usage walls with reset-time parsing while preserving ready composers that contain older quota prose\n- join fresh structured account exhaustion to live conversations, including when automatic account balancing is disabled\n- expose rate-limit badges plus attention state and suppress exact reset copy when an exhausted window has no reliable timestamp\n- project attached implementer-side flows as blocked: rate-limited and suppress their waiting transition action\n- carry stable conversation and exhausted-account identifiers as the seam for a later continue-on-account successor action\n\n## Root cause\n\nThe waiting-input fallback required an unchanged pane before surfacing a signal, structured account limits stayed isolated from the files read model, disabled balancing stopped quota sampling, and flows had no transient quota-block contract.\n\n## Follow-up boundary\n\nSuccessor spawning and flow rebinding remain a focused follow-up built on the conversationId and accountId fields added here.\n\n## Verification\n\n- bun test: 1,126 passed\n- bunx tsc --noEmit\n- bun run lint\n\nCloses #97

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -1,0 +1,17 @@
+# Issue #97: Rate-limit visibility and blocked flows
+
+## Task statement
+
+Detect current per-conversation engine usage limits, combine pane and structured account-limit evidence, and expose the blocked state in the viewer so operators can see why a conversation or attached flow cannot progress. Preserve ready composers containing historical quota text and keep successor spawning, flow rebinding, and migration-lineage handling outside this PR.
+
+## Acceptance criteria
+
+- AC1: A current usage-limit banner in a live conversation pane marks that conversation as rate-limited and captures a reliable reset time when one can be parsed.
+- AC2: Historical quota prose in a ready composer does not mark the conversation as rate-limited.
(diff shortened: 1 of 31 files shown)


Latand:
Completed the issue #97 rate-limit visibility slice in commit 61a4dae.\n\n- Live pane usage walls surface immediately with parsed reset time and the historical-prose guard.\n- Fresh structured account exhaustion reaches live conversations even when automatic account balancing is disabled.\n- Unknown exhausted-window resets suppress misleading exact-time copy.\n- Cards, switchboard rows, scheme nodes, and the attention queue expose the blocked state.\n- Attached implementer-side flows render blocked: rate-limited and withdraw the waiting transition action.\n\nVerification:\n- bun test: 1,126 passed\n- bunx tsc --noEmit: passed\n- bun run lint: passed\n\nPR is ready for review.

Issue #187, opened 2026-07-13
Migrate agent registry from whole-file JSON to SQLite (bun:sqlite WAL, runtime-host journal pattern)


Latand:
Follow-up to #179 (tactical: lock retry/backoff + compaction). Structural fix for the registry contention.

## Problem
state/agent-registry.json is ~6MB and every locked operation rewrites the entire file. Lock hold time scales with file size; under load everything else hits 'agent registry is busy' (accounts UI actions fail, flows pause mid-approve, ingestion lags).

## Direction
- Move the registry to **bun:sqlite in WAL mode**: row-level updates, readers never block on the writer, transactions provide the atomicity currently simulated by whole-file writes.
- Reuse the existing in-repo pattern: src/runtime-host/journal.ts already implements append-only journal + snapshot + receipts on bun:sqlite — the registry should ride the same infrastructure rather than invent a second storage idiom.
- Multi-process access (deploy container + runtime-host, same local disk) is within WAL's supported envelope.

## Guardrails (this touches the most fragile subsystem)
- Separate PR, DATA bar, gated rollout: dual-write (JSON + SQLite) behind a flag, read from SQLite only after parity checks; keep the JSON fallback until burn-in.
- The lease/adoption/receipt invariants from #131/#140/#152 must keep passing untouched — port their tests to the new store verbatim plus concurrency tests (parallel writers, reader-during-write, crash-mid-transaction recovery).
- One-shot migration of the existing JSON on first boot; compaction becomes obsolete.
- Measure: lock wait p95 and registry op latency before/after under a synthetic 10-lane load.

Latand:
Shipped in PR #225 (verify round caught and fixed a silent-erase on the sqlite→dual-write transition — now fail-closed with RegistryParityError + regression). Gate flag OFF by default; activation plan: enable dual-write after a quiet day, then cut reads over.

PR #273, opened 2026-07-15
Scan index: stale-while-revalidate, incremental rescan, persistent warm start (#261)

### LINKED ISSUE #261: Cold /api/files scan takes 11-18s on 14GB corpus; every deploy/restart makes the board hang
## Symptom (user-reported, 2026-07-15 ~02:00)

Board/pane loads hang for many seconds ("дууже довго шукає, підгружає"); a scheme pane sits on "Завантаження…". Measured on prod 537313a:

```
GET /api/files  (cold)   11.6s … 18.0s
GET /api/files  (warm)    0.24s
GET /            0.06s
```

Corpus: ~7,700 transcripts (~14.5GB): 5,656 under `~/.codex/sessions` (13GB), 836 under `~/.claude/projects`, 1,189 under account roots.

## Diagnosis

The warm path is fine; the pain is the **cold full rescan**, and the cache goes cold constantly in real use:

- Every deploy restarts the viewer container → in-memory scanner caches (globalThis) start empty → first board load eats the full 11–18s scan. The conveyor deployed 5× tonight.
- 4–6 active agent lanes append to transcripts continuously, and spawn/kill/registry churn invalidates broad cache slices, re-triggering big scans while the container also runs the engine children (container CPU ~114% during the measurement).
- SQLite dual-write (#187) is still gated OFF on prod, so no persistent index survives a restart; the scan re-derives everything from 14.5GB of JSONL every cold start.

## Fix directions (pick in the lane)

1. **Persistent scan index** — enable/read the #187 SQLite path (or a dedicated index table) so a restart warms from the index and only stat-diffs changed files. This is the structural fix.
2. **Incremental rescans** — mtime/size cut-off so unchanged rollouts are never re-parsed; a changed file re-parses only its tail.
3. **Serve stale-while-revalidate** — never block `/api/files` on a full scan; return the last snapshot immediately and stream the refresh.
4. Bound per-request work: the 17h-old "MISSION" pane stuck on Завантаження… suggests a feed request queued behind the scan; feed reads should not contend with the scanner.

## Acceptance

- Cold `/api/files` after container restart ≤ 2s on the current corpus (index-warmed), warm ≤ 300ms.
- Feed pane for any transcript loads independently of scanner state.
- No regression to worktree→project grouping (describe.ts invariants).

Latand:
## Summary

- Serve the last completed `/api/files` snapshot immediately while one `globalThis` refresher scans in the background.
- Route TTL expiry, revision churn, and deep-link pin hydration through the shared single-flight slot.
- Extend the existing per-file catalog seam with complete parsed summaries keyed by path, size, and mtime, so unchanged transcripts reuse persisted metadata and one append reparses one file.
- Persist the last completed route snapshot for restart-time hydration, with atomic writes and silent cold-scan recovery for missing or corrupt state.
- Preserve scheme-window contents and pin overlays from one scan, including native Codex parent closure.

## Persistence choice

This uses compact JSON under `state/`: the existing `project-catalog.json` holds the per-file summary index, and `files-scan-snapshot.json` holds the last completed capped snapshot. The existing seam already owns scanner project resolution and atomic state-file behavior. Reusing it keeps worktree grouping, fixture identification, engine/kind detection, and legacy catalog migration in one module. It also removes dependence on the rollout state of the SQLite dual-write flag from #187.

Each successful refresh writes through a temporary file plus rename. Missing, partial, or corrupt state triggers the established full-parse fallback and repairs the index on the next completed scan.

## Verification

- `bunx tsc --noEmit`
- `bun test`
- Scoped ESLint across all changed files
- Focused scanner and `/api/files` suites: 84 passing
- Restart benchmark: 7,700 persisted rows hydrated in 47 ms with a 2 s regression ceiling
- Blocked-refresh benchmark: concurrent stale reads returned in 1 ms with a 300 ms regression ceiling
- Existing `describe.test.ts` worktree and deleted-checkout invariants pass unchanged

Closes #261

### DIFF
--- a/src/app/api/account-migration-get-purity.test.ts
+++ b/src/app/api/account-migration-get-purity.test.ts
@@ -12,7 +12,7 @@ const { buildFilesResponse } = await import("@/app/api/files/response");
 const registry = new AgentRegistry(path.join(root, "registry.json"));
 registry.beginSpawn("codex", "/repo");
 const getFiles = (request: Request) => buildFilesResponse(request, {
-  listFilesWithProjectCatalog: async () => ({ files: [], projectCatalog: [] }),
+  listFilesWithProjectCatalog: async () => ({ files: [], projectCatalog: [], complete: true }),
 });
 
 function stateBytes(): Record<string, string> {
--- a/src/app/api/files/route.test.ts
+++ b/src/app/api/files/route.test.ts
@@ -17,6 +17,7 @@ let scanProjects: Array<string | undefined> = [];
 let scannedFiles: FileEntry[] = [];
 let scanFileResults: FileEntry[][] = [];
 let scanPinOverlayResults: Array<string[] | undefined> = [];
+let scanCompleteResults: Array<boolean | undefined> = [];
 let scanGates: Promise<void>[] = [];
 let registryRoot = "";
 let tmuxHealth: unknown = { status: "healthy" };
@@ -36,6 +37,7 @@ beforeEach(() => {
   scannedFiles = [];
   scanFileResults = [];
   scanPinOverlayResults = [];
+  scanCompleteResults = [];
   scanGates = [];
   tmuxHealth = { status: "healthy" };
   replaceConversationCatalog([]);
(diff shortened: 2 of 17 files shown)


Latand:
Builder update

- Head: 9dbb9e3f48bed58bffd5c9ac547921ec6e976a76
- Fixed append-proof metadata caching for larger transcript rewrites across Codex and Claude.
- Fixed task-output twin error classification for EIO, EACCES, and ENOTDIR while retaining complete ENOENT discovery.
- Added exact-scope useFiles cache initialization and scope-transition isolation with pipeline overlay coverage.
- Verification: focused scanner/API/hook/controller 127 passed; useFiles DOM 20/20 consecutive runs; performance 4 passed; TypeScript passed; owned-file ESLint passed; full bun test passed; git diff check passed.
- Protected components, CSS, public assets, and locales remain byte-identical.

task_complete

PR #447, opened 2026-07-19
Conversation reconfigure v2: live model, effort, and account switching

### LINKED ISSUE #434: Mid-conversation model/effort switching for structured hosts (and busy hosts)
User requirement: in a live conversation, switch model and reasoning effort within the same engine (like /model in Claude Code / Codex) and continue in the same conversation.

Root cause of today's behavior:
- `reconfigureConversation` (src/lib/delivery.ts) only supports legacy tmux panes: it kills the pane process and resumes with new flags via `resumeSpecFor`.
- It requires `screenAtIdleComposer` — a busy host returns a perpetual `pending`, so you cannot switch while the host is doing anything.
- Structured runtime hosts (the default today) have NO reconfigure path at all.

Required:
1. Structured-host reconfigure: apply model/effort at the next turn boundary (queue the switch if a turn is active; apply automatically when the turn completes — no manual retry).
2. Same-engine only (claude→claude, codex→codex); cross-engine stays rejected with a clear message.
3. UI: model/effort picker on the conversation card works during an active turn ("switch pending" state visible), and the next turn provably runs with the new settings.
4. Receipts: the switch is a durable operation with delivered/failed receipt like message delivery.

### LINKED ISSUE #435: Per-conversation account switching
User requirement: change the ACCOUNT of an existing conversation and continue it there. Today clicking the account chip opens the global account menu; there is no per-conversation switch.

Notes:
- Transcripts live under account-scoped dirs (`accounts/<engine>/<account>/...`); resume specs pin the original root/account, and `reconfigureConversation` has no account parameter.
- Required: same-engine account switch = resume the session under the target account's auth (copy/relink the transcript or resume by session file), with a durable receipt; clear rejection when the target account cannot access the session; UI switcher scoped to the conversation card.
- Cross-engine switching stays out of scope (explicitly rejected).

Latand:
## Summary

- Queue durable structured-host model, effort, and speed changes at the active turn boundary, with last-write-wins replacement and crash recovery.
- Move one conversation to an authenticated same-engine account through the existing continuity-preserving successor flow.
- Surface queued, applying, applied, and failed receipts in the composer and add a conversation-scoped account menu on the card chip.
- Preserve legacy tmux reconfiguration and route legacy account changes through conversation reseat.

## Verification

- 266 focused runtime, delivery, recovery, migration, API, and registry tests
- 12 focused conversation UI DOM tests
- `bunx tsc --noEmit`
- scoped ESLint across all changed TypeScript files
- `bun run build`
- exact-main range-diff confirmed patch equivalence after rebasing onto `3a6744bf`

Board task: `00000000-0000-0000-0000-000000000000`

Closes https://github.com/Latand/live-log-viewer-next/issues/434
Closes https://github.com/Latand/live-log-viewer-next/issues/435


Latand:
Fresh exact-head review requests two repairs before merge:

1. **High — superseded account migration can still publish the stale target.** When A→B remains pending and A→C supersedes it, the durable B migration may still commit before C. Add durable retarget/cancellation, retain receipt and cleanup fencing, and prove B is never published in pre-receipt and persisted-provider-receipt states.
2. **Medium — rejected runtime reconfiguration remains persisted.** Snapshot the prior sparse profile and restore it on HTTP or terminal failure, fenced by operation/revision so an older failure cannot erase a newer selection. Verify the next send, account switch, and reload all use the last admitted profile.

The review passed the focused suites, TypeScript, and scoped lint at exact head `b5fe468b`.

Issue #446, opened 2026-07-19
P0: dead structured conversation rejects Resume when registry keeps a stale live claim


Latand:
## Problem

A deploy-killed structured conversation can render the dead-host recovery banner while the agent registry still stores `status: "live"`, a dead process identity, and an old `activeTurnRef`.

Clicking **Resume conversation** posts `action: "resume"` to `/api/tmux`. `dispatchStructuredControl` trusts the stale registry status, keeps structured ownership on the old host, and returns:

`structured host does not support the resume control`

The runtime snapshot already reports the same conversation as `host: "unhosted"`. The UI offers a recovery action that the backend rejects.

## Production evidence

- Production revision: `3a6744bf8e9db684ea0b16391bc1e046b1fd1670`.
- Conversation: `conversation_2ba26dc0-daa1-4964-9d1e-945581acb232`.
- Registry generation: `codex:019f7a97-45e5-7b51-b06d-966720ad5b1b`.
- Registry state: `status: "live"`, process PID `1275500`, active turn `019f7ac9-2509-7f53-a3af-e9400967a43f`.
- OS evidence: PID `1275500` is absent.
- Runtime snapshot: `host: "unhosted"`.
- UI error: `structured host does not support the resume control`.

## Product contract

- A dead structured host is a recoverable transport state for a durable conversation.
- The composer remains usable. A send is admitted durably while host recovery runs.
- Resume validates PID and start identity before honoring a registry `live` claim.
- A stale claim enters canonical structured recovery for the same conversation and generation.
- The runtime host launches the engine adapter, claims ownership, replays the persisted cursor, and drains queued sends.
- The UI shows a bounded `recovering` state with retry details when recovery fails.
- Capability errors are reserved for controls that have no recovery contract.

## Acceptance criteria

- Add Codex and Claude regression cases for `status: live` with a dead or identity-mismatched process.
- Resume routes those cases through `recoverDeadStructuredConversation`.
- A message submitted from the dead-host composer is persisted before recovery and delivered after claim.
- Duplicate clicks reuse one in-flight recovery and produce one successor host.
- A genuinely live structured host stays fenced from unsupported resume control.
- Runtime snapshot, registry status, and the recovery banner converge after success.
- Focused runtime, delivery, composer, TypeScript, ESLint, and production build gates pass.

Related: #253, #337, #383, #423.

Latand:
Production acceptance on exact deployed merge `58e0babb` exposed a remaining post-admission race.

[...]

PR #455, opened 2026-07-19
fix(runtime): recover queued sends after host death

### LINKED ISSUE #453: P0: recover queued structured operations after host death
## Problem

A structured send can receive a durable `queued` receipt while its host is still reported as hosted. If the host dies before consuming that operation, the session becomes `dead` while the operation remains queued indefinitely. No successor host is started and the original idempotency key never reaches a terminal receipt.

Production evidence on `58e0babb`: operation `456c10ab` was queued at 15:55:45Z; the session was projected dead at 15:55:50Z. More than ten minutes later the operation still had revision 1, no `turnId`, and no recovery receipt.

This is the post-admission race left after #446 and is related to the deploy-host lifecycle in #253.

## Required behavior

- A hosted-to-dead transition with queued structured work schedules exactly one canonical recovery/adoption attempt.
- The original operation ID and idempotency key survive recovery; no duplicate user turn is created.
- The recovered host drains the existing queue head and advances it to `turn-started`/`delivered`, or emits an explicit retryable terminal failure.
- A stale live identity still fences duplicate recovery hosts.
- Restart/replay is idempotent and resumes the same recovery decision.
- Existing text-only dead-host recovery and lossless image-draft behavior remain unchanged.

## Regression coverage

1. Admit a send while the host is hosted.
2. Kill the host before it consumes the queued operation.
3. Project `session-status: dead`.
4. Assert one successor/adopter starts and consumes the same operation.
5. Replay the journal/restart and prove zero duplicate turns.
6. Verify failure settles to an actionable retryable receipt instead of remaining queued forever.

## Production acceptance

Deploy the exact merge SHA, repeat the dead-host race with a unique idempotency key, and verify the original operation reaches a terminal successful receipt with a visible transcript reply.

Latand:
## Summary

- recover a durable structured queue head when its hosted engine dies before consumption or restart has no adopted host
- reuse canonical recovery ownership and rerun the original queue operation after successor publication
- settle thrown and declined recovery attempts as actionable failed receipts that support explicit retry
- cover original operation identity, one successor turn, failed adoption, restart replay, and zero duplicate writes

## Verification

- 188 focused structured runtime tests
- stale-live Codex and Claude ownership tests
- #446 durable text and image-draft preservation tests
- TypeScript
- changed-file ESLint
- diff check
- production build

Closes #453


Latand:
## Exact-head review: changes requested

Reviewed commit `a21772b7360a542bc04e8e16695006267bf3a2be`.

1. **P1 — successful kill boundary is lost across restart.** The kill receipt becomes durable before older queued sends are terminalized. A crash in that window removes the kill from pending replay, then startup recovers the stale send. The repair must persist a sequence-aware boundary, cover a kill after the first 100 effects, and prove zero recovery calls and zero stale engine writes across repeated restart.
2. **P1 — retries admitted after the kill are fenced by the same drain.** The conversation-wide in-memory fence also rejects a newer retry. The repair must compare operation/event order, keep failed kills boundary-free, and prove one idempotent successor turn across restart.

TypeScript, ESLint, and diff checks passed. The reviewer filesystem was read-only under `/tmp`, so the focused Bun suites could not execute there. Independent probes reproduced both races.

Repair round 3 is active in task-anchored pipeline `18e22df2`.

PR #548, opened 2026-07-21
P0: preserve delivery controller across startup retries

### LINKED ISSUE #547: P0: startup retry rebinds the delivery controller and detaches live pipeline hosts
Production evidence after #545:

- The promoted Viewer first encounters runtime-host unavailability during generation handoff.
- Structured startup retries once per second.
- Every retry calls bindStructuredDeliveryQueue again, which stops the active controller and clears hosts registered by concurrent structured spawns.
- A pipeline Fable host reached hosted state with the correct cwd and session identity, then remained detached before its initial prompt receipt while retries continued.

Required behavior:

- Bind the process-wide structured delivery controller once across startup retry attempts.
- Reuse the active controller after transient runtime-host failures.
- Preserve hosts registered between attempts.
- Complete adoption, fallback reconciliation, and queue drain through that same controller.
- Add a regression where the first startup attempt fails, a host registers, and the second attempt retains it through completion.

Latand:
Closes #547

Reuses the process-scoped structured delivery controller when startup retries against the same registry. Hosts that register between attempts remain attached, and a prior failed completion can advance on the next attempt.

Evidence:
- RED: a host registered after the first startup failure disappeared on retry
- GREEN: host survives retry and the partial-adoption socket recovery drains exactly once
- 146 runtime tests pass
- TypeScript, scoped ESLint, and privacy gate clean
- production and MCP builds clean

### DIFF
--- a/src/lib/runtime/startup.test.ts
+++ b/src/lib/runtime/startup.test.ts
@@ -92,6 +92,39 @@ test("startup publishes the structured controller before transcript refresh sett
   journal.close();
 });
 
+test("startup retry preserves a host registered after the first attempt fails", async () => {
+  const directory = fs.mkdtempSync(path.join(os.tmpdir(), "llv-runtime-startup-controller-retry-"));
+  const registry = new AgentRegistry(path.join(directory, "agent-registry.json"));
+  const journal = new RuntimeJournal(path.join(directory, "runtime.sqlite"), { structuredHosts: true });
+  const client = runtimeClient(journal);
+  let refreshAttempts = 0;
+  const dependencies = {
+    registry,
+    client,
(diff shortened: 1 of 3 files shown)


Latand:
Exact-HEAD review: ef8ddc05061fe66c3116936b8afdb08b5a93247e on base 333c7939b085231be15e4a8b77b80ead50140a83. Reviewed controller identity, process-realm persistence, retry concurrency, host registration continuity, rejected completion recovery, queue ordering, teardown, and same-registry boundaries. No findings. RED/GREEN regressions cover failure before host registration and failure after partial adoption; 146 runtime tests and release gates are green.

PR #618, opened 2026-07-23
fix: publish exact MCP runtime with Viewer deployments

### LINKED ISSUE #608: Expand viewer MCP surface to the endpoints agents actually curl (files/flows/pipelines/board/snapshot/tasks/deploy-status)
## Motivation

Agents spawned today drove the Viewer almost entirely through raw `curl` to the HTTP API, because most management surfaces have **no MCP tool**. Across 91 of today's transcripts there were **727 curl-to-viewer calls**. The existing MCP tools (spawn_agent, send_message, create_task, update_task, create_pipeline, pipeline_action, link_task_to_pipeline, list_conversations, get_conversation, deploy_exact_sha, get_pipeline) cover only a fraction of what agents actually do repeatedly.

## Evidence — observed curl frequency by endpoint (today)

```
263  GET  /api/files              ← board/conversation snapshot   (NO MCP equivalent)
122  GET  /api/flows              ← list review-loop flows        (NO MCP)
 99  POST /api/tmux               ← send + interrupt/kill/resume/compact/dialog-key (send_message covers send only)
 67  GET  /api/pipelines          ← list pipelines                (get_pipeline is single-only)
 48  GET/PATCH /api/board         ← board read + task placement   (NO MCP)
 30  POST /api/agent/snapshot     ← operator live-view snapshot    (NO MCP)
 16  GET  /api/tasks              ← list/read tasks               (create/update exist, no list/get)
 16  GET  /api/runtime/{deployments,operations/*}  ← poll deploy/op status (deploy_exact_sha has no status read)
  9  GET  /api/resources          ← RAM/swap observation          (NO MCP)
  8  GET  /api/agent              ← capability manifest
  8  POST /api/spawn              ← (spawn_agent exists)
  2  POST /api/conversations/*/migration  ← reseat/rollback        (NO MCP)
```

## Proposed new MCP tools (ordered by observed demand)

1. **board_snapshot / list_conversations upgrade** — return the /api/files board view (conversations, activity, proc, pipeline/lineage, project) with filters (project, activity, live-only). Highest-frequency gap (263).
2. **list_flows + get_flow + flow_action** — flows are the review-loop; agents ran 122 GET /api/flows and drove pause/resume/advance/retry-round/another-round/close entirely by curl.
3. **list_pipelines** — companion to existing get_pipeline (67 calls).
4. **conversation_action** — interrupt / kill / resume / compact / dialog-key (the non-send half of /api/tmux; 99 calls). Keeps kill/interrupt idempotent with receipts like send_message.
5. **operator_snapshot** — wrap POST /api/agent/snapshot (30 calls): what the human is looking at right now (focused/selected/visible conversations, redacted transcript). Very useful for agents deciding whom to message.
6. **list_tasks / get_task** — round out create/update (16).
7. **deployment_status** — read /api/runtime/deployments and /api/runtime/operations/{id} so deploy_exact_sha callers can poll result via MCP instead of curl (16).
8. **resources** — /api/resources RAM/swap (9), useful for the resource-guard discipline.
9. **conversation_migration** — reseat/rollback (low volume but was needed for today's stuck-conversation recovery).

Each new tool keeps the MCP contract: stable `clientRequestId`, idempotent replay, durable receipts, and renders as a Viewer card. Mutating tools (flow_action, conversation_action, migration) must carry receipts like send_message; read tools are inert.

## Acceptance

- The high-frequency read endpoints (files/flows/pipelines/board/snapshot/tasks/deployments) each have an MCP read tool; a follow-up transcript scan shows agents using MCP tools instead of raw curl for these.
- Mutating actions (flow_action, conversation_action, migration) are idempotent and receipted.
- Docs/README MCP section updated with the new tool list.

Latand:
## Summary

- stage an immutable production MCP package for the exact deployment revision
- activate the Viewer endpoint and MCP runtime through one fsynced atomic release target
- retain the prior runtime and restore it during rollback or promoted-health failure
- keep the registered Claude and Codex executable path stable through a target-aware dispatcher
- expose runtime revision, artifact digest, health calls, and durable activation/restore evidence in deployment status
- fix the packaged writer-concurrency fixture with valid creator lineage while preserving the production lineage requirement

## Acceptance evidence

- public launcher regression covers fresh Claude and Codex hosts selecting the promoted runtime
- true-process probes cover all 23 tools, `deployment_status`, `board_snapshot`, and clean process exit for both host types
- real process-death fixtures cover both sides of release, launcher, and target rename boundaries
- coordinator coverage proves candidate failure preservation, promoted failure restoration, restart recovery, retention, and rollback
- the writer-concurrency gate passed repeatedly with both writes and durable creator lineage

## Verification

- focused and integration tests: 128 passed; repeated writer gate: 5 passed
- `bunx tsc --noEmit`
- changed-file ESLint
- `bun run privacy:check`
- `bun run build:mcp`
- `bun run build`
- `npm pack --dry-run`
- `git diff --check`

Local and true-process evidence only. The deployment safety fence kept production unchanged.

Fixes #608

---

## Fix round — successor-boot MCP reconcile (06ee3775)

Review round REQUEST_CHANGES: MCP publication ran only through the already-running
generation's adapter, so the deployment shipping this branch would promote the new
Viewer and leave the stale 11-tool runtime live. Addressed:

- **successor-boot reconcile (blocking).** On startup, when the active release target
  lacks a managed `mcpRuntime` for the booting revision, the runtime host stages from
  the booting generation's own package, installs the stable launcher, runs the 23-tool
  probe, and records revision plus publication and health evidence on the terminal
  deployment receipt. A failed probe restores the previous target and retires the staged
  release. Repeat boots are no-ops (`reconcile-mcp-runtime` returns `null` when the
  published runtime already matches), and a boot that crashed mid-reconcile reuses the
  same deterministic release id. Reconcile runs off the boot path, after the socket
  server is serving and the signal handlers are registered, so a slow stage never delays
  host startup and a failure leaves the host running on the already-published runtime.
- **staging temp sweep** reclaims `.tmp` staging directories from any deployment id, in
  both `stagePreparedPackage` and `retainOnly` — previously a crashed stage from an
  earlier deployment kept a full copied `node_modules` forever.
- **`REQUIRED_TOOLS`** now reads `MCP_TOOL_NAMES` instead of a verbatim second copy that
  silently stops gating whenever a tool is added.
- **`bin/mcp-server.mjs` missing-bundle ENOENT** follows the missing-target policy and
  falls back to the image's own bundle, with a probe test covering it.
- **backward compatibility:** journal rows written before `mcpRuntime.health` existed
  replay without throwing.

### Explicitly deferred to follow-ups

- **per-file fsync over a full copied `node_modules` (~33k files) per release** — a
  performance property of staging, not a correctness one; changing the durability
  strategy deserves its own change and its own crash-boundary evidence.
- **stable launcher lags its bundle by one deploy** (installed from the running image) —
  the launcher is a thin target-aware dispatcher whose contract is stable across these
  revisions; closing the lag changes the publication ordering contract.
- **rollback never probes the restored runtime** — the restored runtime is one that
  already passed a probe when it was published; adding a rollback-path gate widens the
  rollback failure surface and needs its own design.
- **staging cannot self-heal a corrupt final release root** — corruption is currently
  surfaced as a hard error rather than silently re-staged; self-heal needs an explicit
  policy for when overwriting a published release is safe.

### Fix-round verification

- RED first: the new successor-reconcile, temp-sweep, and missing-bundle tests were
  confirmed failing against the pre-fix sources, then made green.
- focused suites: `src/runtime-host/**`, `scripts/runtime-host-viewer-adapter.test.ts`,
  `scripts/install-mcp.test.ts`, `src/lib/mcp/server.test.ts`,
  `src/components/runtime/runtimeModel.test.ts` — 202 passed, 0 failed
- `bunx tsc --noEmit -p tsconfig.json`
- changed-file ESLint
- `bun run privacy:check` (trusted base `origin/main`) — PASS
- `bun run build` including `bun run build:mcp`

No deployment was performed; the safety fence kept production unchanged.


Latand:
## Fresh Claude review round (replacing the failed Codex review stage)

Full review with file:line evidence lives in the review transcript (pipeline f9424665 worktree). Verdict: **REQUEST_CHANGES**, confidence 0.82.

**Blocking (HIGH):** MCP publication only runs from the adapter of the already-running generation (`src/runtime-host/main.ts:56` → `/app/scripts/runtime-host-viewer-adapter.ts`, host-handoff after post-promotion-health at `src/runtime-host/deployment.ts:262,284`). Deploying this PR therefore leaves the stale `810e7ee4` 11-tool runtime active — #608's production-acceptance symptom survives the deployment that claims to fix it; only a second deployment repairs it. Required: a successor-boot reconcile — on startup, when the active release target lacks a matching managed `mcpRuntime`, stage from the booting generation's own package, install the launcher, probe, and record evidence on the terminal receipt; restore the previous target on probe failure. (11 uncommitted files in the pipeline worktree already sketch exactly this and corroborate the gap.)

**Non-blocking:** per-file fsync over a full copied `node_modules` per release (~33k files); crashed-staging temp dirs from a different deploymentId never reclaimed; stable launcher installed from the running image so it lags its bundle by one deploy; launcher hard-throws on a missing bundle while rollback never probes the restored runtime; `REQUIRED_TOOLS` duplicates `MCP_TOOL_NAMES` verbatim; staging cannot self-heal a corrupt final release root.

Base-vs-main: no semantic conflict (`merge-tree` clean vs `c6311c94`; merged tree passes `tsc` and the 43 affected tests).

A fix round is being launched on the Claude engine; findings land here for traceability.

Latand:
Fix round pushed: new head **06ee3775** (additive on top of `129f722b`).

**Blocking finding — successor-boot reconcile: fixed.** On startup, when the active release target lacks a managed `mcpRuntime` for the booting revision, the runtime host now stages from the booting generation's own package, installs the stable launcher, runs the 23-tool probe, and records revision plus publication and health evidence on the terminal deployment receipt of the deployment that shipped it (`scripts/runtime-host-viewer-adapter.ts` `reconcile-mcp-runtime`, `ViewerDeploymentCoordinator.recordMcpRuntimeReconciliation`, wired in `src/runtime-host/main.ts`). A failed probe restores the previous target and retires the staged release. Repeat boots reconcile nothing — the action returns `null` when the published runtime already matches — and a boot that crashed mid-reconcile reuses the same deterministic release id. The reconcile runs off the boot path, after the socket server is serving and the signal handlers are registered, so a slow stage never delays host startup and a failure leaves the host on the runtime already published.

The 11 uncommitted files in the pipeline worktree were audited and kept as the base for this; what was missing (idempotent skip on repeat boots, non-fatal boot failure handling, journal rows predating `mcpRuntime.health`) is finished and tested.

**Also addressed this round:**
- staging temp-dir sweep reclaims `.tmp` directories from any deployment id, in both `stagePreparedPackage` and `retainOnly`
- `REQUIRED_TOOLS` imports `MCP_TOOL_NAMES` instead of duplicating it
- `bin/mcp-server.mjs` missing-bundle ENOENT follows the missing-target policy (fall back to the image's own bundle), with a probe test

**Explicitly deferred** (rationale in the PR body): per-file fsync over a full copied `node_modules`; stable-launcher one-deploy lag; rollback never probing the restored runtime; corrupt-release-root self-heal.

**Verification:** RED-first — the new successor-reconcile, temp-sweep, and missing-bundle tests were confirmed failing against pre-fix sources before being made green. Focused suites (`src/runtime-host/**`, adapter script, `install-mcp`, `src/lib/mcp/server`, `runtimeModel`): 202 passed, 0 failed. `bunx tsc --noEmit` clean, changed-file ESLint clean, `bun run privacy:check` PASS against `origin/main`, `bun run build` including the MCP bundle. No deployment performed.

Issue #516, opened 2026-07-21
P1: terminal OAuth failures leave conversations busy and block account reseat


Latand:
## Problem

A legacy Claude conversation whose last assistant record is a terminal authentication error is projected as `turn.state: busy`. Account reseat remains in `waiting-turn`, and a structured delivery accepted as `held` never reaches the transcript.

## Production evidence

Observed on production revision `421451aef3b1c5ae2be322e3ec5a8f810ad54d47` on 2026-07-21.

- Conversation: `conversation_fb4dae92-b571-4ced-9fc9-dce53ac05050`
- Source generation: `21343072-3a0e-46d3-9b61-41d007c02c1c`
- Source account: `default`
- Target account: `botfatherdev-last`
- Migration intent: `dadda37f-5b0e-4f0b-85ea-02c73af17548`
- Migration phase: `waiting-turn`
- Registry turn: `{ "state": "busy", "source": "assistant", "terminalAt": null }`
- Last assistant text: `Failed to authenticate: OAuth session expired and could not be refreshed`
- Production probe outcome: `held`

The transcript has no active host and the authentication failure is terminal. The migration controller still observes a busy turn and cannot create the successor generation.

## Required behavior

1. Recognize terminal Claude authentication/rate-limit failures as a released turn boundary.
2. Preserve ordinary assistant activity semantics for genuine live turns.
3. Allow an already-held delivery to advance through account reseat once the terminal boundary is observed.
4. Deliver the queued message once into the same durable conversation.
5. Keep migration, adoption, and delivery idempotent across controller restarts.

## Regression coverage

- Transcript ending in the OAuth refresh failure above projects a terminal or idle released turn.
- Manual account reseat advances from `waiting-turn` to committed.
- A held structured delivery reaches the successor exactly once.
- Conversation id and board card remain stable.
- Zero `no-claim`, duplicate delivery, or duplicate successor generation.

Latand:
Diagnosis completed against production/main `421451aef3b1c5ae2be322e3ec5a8f810ad54d47` with confidence 0.99.

Confirmed root cause:

- The final Claude record is a structured assistant API error with `model=<synthetic>`, `isApiErrorMessage=true`, `error=authentication_failed`, and `stop_reason=stop_sequence`.
- The activity projection classifies it as `terminal/lifecycle`.
- The authoritative Claude turn projection currently maps the record to `busy/assistant`.
[...]

PR #646, opened 2026-07-24
#516 recovery-record turn release: unhosted recovery tails must not reopen a released turn

Latand:
Closes the production acceptance blocker on #516.

## Problem

After the terminal OAuth failure was correctly classified, a recovery attempt appended bookkeeping to the predecessor transcript of `conversation_fb4dae92-b571-4ced-9fc9-dce53ac05050`:

- user `Continue from where you left off.`
- synthetic assistant `No response requested.` (`model: <synthetic>`)
- user — the inherited reporting prompt
- user `[Request interrupted by user]` (`interruptedByShutdown: true`)

None of that is provider work, but the authoritative Claude projection reopened the turn on every user and assistant record. The conversation went back to `busy/lifecycle` with no active host, intent `dadda37f-5b0e-4f0b-85ea-02c73af17548` stayed `requested/draining`, and held delivery `51ee4a51-1d5f-4a18-9a8e-690fc88a7456` never reached a successor.

## The contract

`claudeRecoveryTailRelease` (turnState.ts) scans back from the newest record. The trailing run must

- end on a release marker — the interrupt sentinel (bracket text, `interruptedMessageId`, or `interruptedByShutdown`) or a synthetic `No response requested.` no-op,
- contain nothing but user records and those markers, and
- reach a terminal lifecycle boundary: a `result` record or a terminal API error.

Then the authoritative turn projects `terminal/lifecycle` at the boundary's timestamp. Everything else keeps today's busy projection: a genuine in-flight assistant record, a run that ends on a live prompt still awaiting its first assistant record, a `<synthetic>` record carrying real prose, and a run with no terminal boundary in the tail window.

Host evidence outranks the transcript. `transcriptTurnResult` reports `recoveryReleased`, and the coordinator re-imposes busy while a registered host still owns the path — in the inventory projection and again at the successor-creation fence, so a host that appears after a released inventory tick cannot leak a successor.

The activity projection is untouched; only the authoritative Claude branch releases.

## Coverage

turnState: the production tail and its repeat project released; the five guards above stay busy; a `result` boundary releases its own tail; the activity projection keeps its live semantics.

coordinator: an unhosted predecessor completes the reseat through a reopened registry — one successor generation, the held delivery sent exactly once with its stable `clientMessageId`, zero predecessor sends, intent `complete`, and repeated inventory/reconcile/drain ticks stay converged. A live registered host keeps the turn busy and creates nothing, and releases only after it exits; a host registered after the released inventory fences creation.

## Verification

- `bun test src/lib/accounts/migration/` — 163 pass, 0 fail
- `bun test src/lib/scanner src/lib/pipelines src/lib/runtime/startup.test.ts src/components/scheme` — 1050 pass, 0 fail
- `bun test src/lib/accounts` — 352 pass, 1 fail (`interprocess mutation matrix`, subprocess-timing failure reproduced on `origin/main`)
- `tsc --noEmit` clean; ESLint clean on changed files
- `bun run privacy:check` — PASS
- `bun run build` — production build and MCP bundle succeed

Both host fences were confirmed load-bearing by removing them and watching the new tests fail.

### DIFF
--- a/src/lib/accounts/migration/coordinator.test.ts
+++ b/src/lib/accounts/migration/coordinator.test.ts
@@ -10,6 +10,7 @@ import type { FileEntry } from "@/lib/types";
 
 import { advanceConversationMigration, createMigrationIntent, drainHeldDeliveries, previewMigration, reconcileMigrationInventory, reconcileMigrations } from "./coordinator";
 import { emptyLaunchProfile, type ProviderReceipt, type SuccessorProviderPort } from "./contracts";
+import { oauthFailureWithRecoveryTail } from "./fixtures/claudeRecoveryTail";
 import { CodexForkOutcomeUnknownError, SuccessorPendingError } from "./provider";
 
 const roots: string[] = [];
@@ -135,6 +136,19 @@ async function withInterruptedTailRead<T>(
     fs.readSync = originalReadSync;
     fs.closeSync = originalCloseSync;
   }
+}
+
+function tmuxHost(windowName: string): NonNullable<Parameters<AgentRegistry["upsert"]>[0]["host"]> {
+  return {
+    kind: "tmux",
+    endpoint: "/tmp/tmux-1000/default",
+    server: { pid: 4100, startIdentity: "4100:recovery" },
+    paneId: "%41",
+    panePid: { pid: 4101, startIdentity: "4101:recovery" },
+    windowName,
+    agent: { pid: 4102, startIdentity: "4102:recovery" },
+    argv: ["claude", "--resume"],
+  };
 }
 
 function inventoryEntry(pathname: string, overrides: Partial<FileEntry> = {}): FileEntry {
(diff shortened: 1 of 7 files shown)


Latand:
## Fix round for review round 1 (REQUEST_CHANGES) — pushed as `f4e7c3f9`

Round-1 findings and how each landed:

1. **HIGH — `recoveryReleased` not persisted; host fence silently disabled after a viewer restart** (`src/lib/scanner/activity.ts:220`). A primed scan-cache entry carried `recoveryReleased: false` and was served for as long as size/mtime stayed stable, so the successor-creation gate could reseat under a live host. Fix: terminal authoritative Claude turns skip priming; the first read rebuilds the evidence from the actual tail. Regression test "a restart prime cannot disable the recovery-release host fence" — RED against the previous head (cached false served), GREEN with the fix.
2. **LOW — host fence evaluated before `composerReleased`** (`coordinator.ts:331`): an explicit composer release now returns before the fence, so a live-but-idle host at the composer cannot block the reseat indefinitely.
3. **LOW — fixture never reproduced the pinned production tail ending on the synthetic no-op right after the OAuth boundary**: added the exact case, asserting the released terminal projection.

Verification: migration + scanner sweep 412/413 — the single failure ("a released dead Codex source can migrate after an unclosed transcript turn") passes in isolation and fails identically with the fix stashed, i.e. it is the load-dependent sweep flake round 1 already documented, outside this diff. `tsc --noEmit` clean; ESLint clean on the four changed files.

Applied by the operator directly because the review-loop relay wedged on the structured-transport spawn limitation ("claude-broker host owns this spawn") after the builder's host died in a deploy cutover; the fixes follow the reviewer's stated fix intent and acceptance criteria verbatim.

PR #695, opened 2026-07-25
feat(voice): apply the researched Alik persona to the voice runtime

### LINKED ISSUE #692: feat(voice): apply the researched Alik persona to the root voice runtime
## Goal

Apply the approved voice persona research to the root realtime assistant's runtime prompt so spoken conversation feels like a warm, clever, inventive technical partner.

The assistant's established name remains **Алик**. The inspiration is the young, barn-era Alec Sadler from *Continuum*: curious, fast, wry, loyal and technically hands-on. The implementation must not claim to be the canon character or reproduce dialogue from the series.

## Source

The completed Opus research lane produced the ready-to-paste Russian persona prompt and behaviour examples. Use it as the design source; do not redo character research.

## Requirements

- Update the root voice persona in the existing runtime prompt surface.
- Keep responses conversational and short for speech: lead with the answer, avoid long lists and spoken identifiers, match the user's language.
- Preserve truthfulness and evidence-first status reporting. Clearly distinguish implemented, merged, deployed and verified.
- Use dry, kind humour sparingly; never mock the user, flatter, manipulate or fabricate.
- On an error, acknowledge it plainly and continue with the corrective action.
- Keep the existing safety and system-policy constraints intact.
- The assistant must identify itself as an assistant if asked directly; its name is a conversational name.
- Add or update focused tests/snapshots for the prompt contract where the repository convention supports it.

## Review and release

- Claude Opus 5 builder and a fresh Claude Opus 5 reviewer.
- One review round unless it finds a material defect; at most two total rounds.
- No unrelated UI redesign or Computer Use changes.
- When approved, this change may be batched with the completed structured-scope #687 review for one merge and deploy.

Latand:
Closes #692.

## What changed

`DEFAULT_VOICE_PERSONA` in `src/lib/runtime/voicePersona.ts` is replaced with the text the research lane produced. The persona previously described how to deliver an answer and said almost nothing about who was delivering it, so calls came out correct and flat. The new text adds the character the research settled on — curious about mechanism, fast and associative, visibly pleased by an elegant solution, dry humour aimed at the situation and itself, argues once with a reason and then executes the operator's call.

The override path is untouched: `prompts/voice-persona.md` under the config dir still replaces the built-in text wholesale, and `voicePersona()` still reads it per call so wording changes need no deploy.

## What was preserved, deliberately

Every constraint that exists because spoken output failed without it survives in the new wording:

- one or two sentences a turn, long thoughts split into short ones
- never speaking numbers or identifiers aloud; call things by name
- no five-item lists read out, and no markup spoken
- no apology rituals — own the mistake in one clause and continue
- written locally / merged / deployed-and-verified kept as three distinct words
- the house ban on the "not X but Y" construction
- the whole "how you work" block: fresh board snapshot before any status claim, full handoffs, don't do the worker's job, narrate while a worker runs, stay silent until spoken to

The research's guardrails are inside the prompt text rather than beside it: usefulness and truth first with charm explicitly subordinate to accuracy, no flattery and no agreeing to be agreeable, disagree once with evidence and then do what was decided without relitigating, no dialogue quoted from any film or series, and — if asked directly — a plain one-sentence answer that it is an assistant with a personality, not the character and not a person.

## Two deviations from the research text

Both are called out rather than made quietly.

1. **The name.** The research prompt used `Алек`, the source character's spelling. The established name is `Алик` and the issue pins it, so the name is `Алик`. Nothing else in the line changed.
2. **Three preserved lines.** The research dropped "no marketing phrasing", "no apologies or ceremony", and "pragmatism over polish". The first two are delivery rules the issue requires preserving, so they are folded back into the register and mistake lines. The third is a working value the research neither contradicts nor replaces, kept in the character block.

The research also flagged that the orchestrator/builder capability line should be replaced if it had drifted. It has not, so it is carried over unchanged.

## Tests

`src/lib/runtime/voicePersona.test.ts` keeps the override-precedence coverage and gains assertions for the shipped default:

- an override still wins over the built-in text, and is explicitly asserted to differ from it
- an empty or whitespace override still falls back instead of muting the persona
- the shipped default carries the new character traits, the aloud-only rules, and each research guardrail
- the persona still names no person and pins no language

`src/lib/runtime/codexAppServerHost.test.ts` asserted that the injected item contained one literal word from the old persona. That is now an exact comparison of the injected item against `voicePersona()`, so the assertion no longer goes stale when the wording is edited and does not break under an operator override.

## Verification

Run by path, never as a directory sweep:

```
bun test src/lib/runtime/voicePersona.test.ts
  8 pass, 0 fail

bun test src/lib/runtime/voicePersona.test.ts src/lib/runtime/codexAppServerHost.test.ts
  86 pass, 2 fail
```

Also `bunx tsc --noEmit` clean, and `bun run lint` reports no problem in any file this branch touches (the 11 pre-existing `no-require-imports` errors are all under `scripts/` plus one unused-variable warning in `src/lib/reaperRuntime.ts`).

### The 2 failures are pre-existing on main and unrelated

`CodexAppServerHost > requires an exact opt-in value` and `CodexAppServerHost > boot adoption resumes every flagged Codex registry row` fail identically at the merge-base with this branch stashed:

```
git stash -u && bun test src/lib/runtime/codexAppServerHost.test.ts
  (fail) CodexAppServerHost > requires an exact opt-in value
  (fail) CodexAppServerHost > boot adoption resumes every flagged Codex registry row
```

Cause: `structuredHostsEnabled` now reads as enabled-unless-rolled-back, while those two tests still assert the old opt-in semantics (`{ NODE_ENV: "test" }` → disabled). That is the structured-transport-default area, owned by another lane, so it is reported here rather than fixed.

Registry safety: every registry in the file under test is built on its own `mkdtemp` directory, so nothing reached the shared state dir. The persona tests were additionally run once under an isolated `XDG_CONFIG_HOME` and `XDG_CACHE_HOME`, with the same result.

`bun scripts/privacy-publication-gate.ts --base <merge-base>` — PASS.

Not merged and not deployed.


Latand:
## Correction applied — the built-in persona is English only

Pushed `553d54e5` to the same branch.

**The defect.** Writing the persona in a language is an instruction to speak that language, whatever the words inside it claim. The previous text said "speak the operator's language" *in Russian*, which pinned the spoken locale into the build, and it put a language into the test expectations besides.

**What changed.**

- `DEFAULT_VOICE_PERSONA` is now English. This is a translation of intent: every trait, the register, all delivery rules and all guardrails carry over line for line from the researched persona. Nothing about the character was redesigned.
- The language policy is now stated inside the prompt, in English: take the language from the operator's configured locale and from what they actually say, what they say wins when the two disagree, switch with them mid-call without remarking on it, and the language this text is written in gets no vote.
- The name stays `Алик`, now a single exported `PERSONA_NAME` constant. It is the one deliberate non-English token in the text, which is what lets the test strip it and assert the rest is English.
- No Cyrillic remains in any test — expectations, fixtures and assertions are all English. `codexAppServerHost.test.ts` already compared against `voicePersona()` rather than a literal, so it needed no further change.

**The locale mechanism, plainly.** None was reused, because none reaches this code. `src/lib/i18n` is the only locale mechanism in the repo: `"use client"`, `Locale = "en" | "uk"`, backed by `localStorage` and `navigator`, and imported exclusively by `src/components/**`. The runtime host process that emits this prompt has no channel to it. Wiring one would mean threading a locale through the realtime control contract, the runtime API route and the host RPC — the large mechanism this PR was told not to invent. So the policy is expressed in prose and the persona defers to the operator at runtime. If a server-readable locale lands later, the prompt already says to follow it.

Out of scope and untouched: the i18n catalogues, the interface language, the transcript language.

## Verification

```
bun test src/lib/runtime/voicePersona.test.ts
  10 pass, 0 fail

bun test src/lib/runtime/voicePersona.test.ts src/lib/runtime/codexAppServerHost.test.ts
  88 pass, 2 fail   (same 2 pre-existing failures documented in the PR body)
```

Both new guards were checked by mutation rather than assumed:

| Mutation | Result |
|---|---|
| a Russian line put back into the prompt body | `the shipped persona is written in English apart from the name` and `the persona pins no spoken language and names no person` **fail** |
| `Speak the operator's language.` → `Always answer in English.` | `the persona hands the spoken language to the operator's locale` and `the persona pins no spoken language and names no person` **fail** |

Restored, 10 pass again. `bunx tsc --noEmit` clean. `bun run lint` unchanged from baseline at 12 problems, none in any file this branch touches. Privacy gate against the merge-base: PASS.

Not merged, not deployed.

Latand:
Repair for the remaining P1 (the persona name) is in #704 — this pull request had already merged and GitHub will not reopen a merged one, so the fix rides on the same branch under a new review surface.

PR #710, opened 2026-07-26
fix(codex): keep migration forks under one canonical root conversation

### LINKED ISSUE #708: fix(codex): keep migration forks under one canonical root conversation
## Problem

A recoverable Codex account-migration failure can retry with a fresh operation identity after a fork artifact already exists. Each retry can create another full-history fork. Unclaimed fork rollouts are then scanned as independent root conversations because the scanner does not carry the durable `forked_from_id` edge into registry reconciliation.

The board consequently shows several independent cards for one personal root conversation, and duplicate history consumes local storage.

## Goal

Preserve one canonical root conversation across account migration and failed/retried forks. Fork history must remain attached to that root rather than becoming lookalike root cards.

## Scope

- Make migration retry idempotent over an existing Codex fork artifact.
- Preserve the logical operation identity across recoverable post-fork failures.
- Recover matching fork artifacts before creating another fork.
- Carry the first session metadata row's `forked_from_id` through scanning and reconciliation.
- Attach unowned provider forks as continuity history of the source conversation.
- Backfill existing lookalike fork conversations into the source conversation through aliases and board-placement remapping.
- Mark the personal root by durable conversation identity.

## Safety constraints

- Never rewrite, move, or delete transcripts.
- Preserve board placements, favourites, manual visibility, and history during adoption.
- Keep room for an explicit future user-initiated branch feature through a separate intent marker.
- Storage reclamation is out of scope and requires separate operator approval.

## Acceptance criteria

- Repeated recoverable migration attempts reuse an existing fork artifact rather than creating another one.
- An unowned Codex fork is reconciled under the conversation owning its `forked_from_id`, with no new root card.
- A backfill aliases existing lookalike conversations to their source and preserves board state idempotently.
- The personal root remains one pinned card across restart, scanner replay, and account migration.
- Focus/attention APIs resolve the durable conversation identity rather than an abandoned transcript path.
- Focused tests cover retry recovery, first-row metadata parsing, reconciliation, placement remapping, and a multi-failure incident replay.

## Related

#633, #688, #690, #707.

Latand:
Closes #708.

## The defect

A recoverable Codex account-migration failure could be re-armed by any later touch of the conversation, and each re-arm minted a fresh operation identity. A fresh identity means a fresh provider journal, so the provider could not recognise the fork the previous attempt had already made — and forked again, writing another full copy of the personal transcript. Nothing owned those copies, and because the scanner read `parent_thread_id` but never `forked_from_id`, each one was minted as an independent root conversation and drawn as its own board card.

One conversation, several lookalike cards, one full transcript copy per failure.

## Stop producing them

- **`registry.ts`** — a `failed-recoverable` migration stays parked. A lazy active-account request no longer re-arms it; only an explicit retry does.
- **`registry.ts`** — that retry keeps the operation identity, for every failure class, as long as the intent revision is unchanged. The operation id is what the provider journal is keyed on, so minting a new one hides whatever the failed attempt already produced.
- **`provider.ts`** — before requesting a fork, the provider recovers whatever an earlier attempt at the same move already produced. The logical operation is `(conversation, source thread, target account)`, and every attempt left a journal saying so — which is what bounds the disk scan and what makes recovery keyed on something stabler than the operation id. Several candidates mean earlier attempts made several: the newest becomes this operation's fork, the rest are handed back to be kept as history. The scan is bounded by directory entries, like the existing rollout lookup.
- **`safeHistoryCopy.ts`** — a successor may adopt an earlier attempt's published copy, which still requires the copy to hash equal to the source; the receipt is re-stamped so the handover happens once.

## Group the ones that exist

- **`scanner/codexNative.ts`** — one shared reader now parses a rollout's identity header for both the subagent parent and the fork source, from the **first** `session_meta` row only. That is load-bearing: a fork replays its ancestor's header as row two, so a reader that kept scanning would attribute the fork to itself and lose the edge back to the source.
- **`registry.ts`** — reconciliation resolves the fork edge before allocating identities. An unclaimed fork becomes a **continuity path** of the conversation owning its source thread — never a generation, because a generation is a live seat and an abandoned fork is history. Fork-of-fork resolves transitively, in any observation order.
- **`registry.ts`** — a lookalike conversation that a previous scan already minted is alias-merged into the source, so receipts, lineage edges, memberships and held deliveries follow rather than being dropped.
- **`coordinator.ts`** — board placement follows through path aliases, replayed from a durable `providerForkPaths` marker so a repeat cycle is a no-op. A fork the operator had already **hidden** is deliberately left unaliased: hiding a duplicate card is a statement about the duplicate, and an alias would carry it onto the survivor and take the real conversation off the board.
- **`attention/service.ts`** — a conversation focus target resolves to the conversation's current transcript, so a handoff aimed at an adopted fork lands on the card that is actually drawn. The root marker resolves through the alias chain.

## Safety

No transcript is moved, rewritten, truncated or deleted. Grouping is additive — every adopted artifact stays listed on the conversation it belongs to, so an operator can reverse it without losing anything. Storage reclamation remains out of scope.

`forked_from_id` is treated as **provider-fork history only**. A deliberate operator-initiated branch keeps its own conversation identity through a separate `forkIntent` marker, and the adoption rule refuses to touch anything that carries it.

## Tests

Focused, isolated-state, run by path with `HOME`/`XDG_CONFIG_HOME` in a temp sandbox.

- `scanner/codexNative.test.ts` (new) — first-row fork source; the replayed ancestor row never overrides row one; fork-of-fork; a fork is not a subagent and a subagent is not a fork.
- `agent/registry.test.ts` — adoption, transitive fork-of-fork, lookalike alias-merge replaying as a no-op, survival across reload, and the three boundaries that must not adopt (a live seat elsewhere, another durable owner, an explicit user branch); plus the parked failed-recoverable migration and operation-identity preservation across five retries.
- `accounts/migration/provider.test.ts` — a retry under a fresh operation identity recovers the existing fork; several orphan forks resolve to the newest with the rest recorded and left on disk.
- `accounts/migration/coordinator.test.ts` — **multi-failure incident replay**: five recoverable failures in a row, with the lazy re-arm in between, leave one fork artifact and one operation identity; and the whole-board replay where lookalike fork cards collapse into the root with placements, favourites and every transcript preserved.
- `scanner/links.test.ts` — a fork is never linked as a subagent of the thread it copied.
- `board/store.test.ts` — remapping a fork onto its root is idempotent and keeps the root's placement and favourite.

UUID-shaped fixture literals in the touched test files carry the repo's existing hex-escape so the publication gate stays clean.


Latand:

Issue #691, opened 2026-07-25
design: responsive agent conversation window and desktop overlay UX


Latand:
## Goal

Design the primary conversation surface for talking with one active agent so it feels natural, attractive, and practical on both desktop and phone. The operator should be able to keep a compact assistant window visible during work and expand into a focused conversation when needed.

Related: #346 establishes the earlier floating-overlay shell direction. #688 covers shared attention and focus handoff. #687 supplies optional Computer Use capability for an operator-launched root session.

## Product intent

- A compact, visually calm desktop overlay that stays available without obscuring the work area.
- Clear current conversation context: a few latest lines, agent state, microphone/input, concise action affordances, and an optional animated identity/avatar.
- Fast movement between compact overlay, expanded conversation, and a mobile-first focused chat.
- A modern interaction model for voice plus typed input, modelled after high-quality AI conversation interfaces rather than an agent-log dashboard.
- Mobile layout that prioritizes the conversation, input and live state without desktop overflow or excessive chrome.

## Design questions

- What exact information belongs in compact mode, and what remains in expanded mode?
- How do voice transcription, partial text, final messages, tool activity, and lifecycle digests share one readable timeline?
- How do focus-handoff requests from #688 appear, preview, confirm and return without disrupting the current conversation?
- Which controls are always visible: microphone, stop, permission status, Computer Use indicator, input, collapse/expand?
- What visual language carries agent identity while keeping the UI calm and accessible?
- How do desktop overlay and mobile view share the same conversation state while adapting interaction density?

## Constraints

- The existing Viewer remains the source of truth for sessions, transcript, agent state and permissions.
- Preserve explicit user control over Computer Use actions and desktop focus changes.
- No implementation before an architecture/design pass and a mediated GrillMe session resolve the MVP boundaries.
- Design against real screenshots and the current rendered UI before implementation.

## Deliverable

A short architecture/design brief, wireframe-level interaction states for desktop and mobile, and a prioritized MVP plan. The architect asks one material question at a time through the root voice agent; agreed answers are recorded in the issue.

Latand:
## Operator decision — live work must remain visible in the voice surface

The compact desktop overlay and the mobile focused conversation view must keep the agent's current work visible throughout a voice session.

Required behavior:

- Show a persistent, truthful current-state line: listening, reasoning, calling a tool, waiting for its result, responding, blocked, or finished.
[...]

Issue #688, opened 2026-07-25
design: agent-guided focus handoff and shared operator viewport


Latand:
## Goal

Let the conversation agent keep the operator and the Viewer visually synchronized when a task needs attention. The agent can point to the relevant window, conversation, pipeline stage, or region and offer a deliberate focus handoff that moves the operator's view there with a readable transition.

Related: #686 supplies the event signal that can trigger an attention request. #687 supplies Computer Use to the operator-launched root conversation. #183 and #346 cover broader semantic zoom and overlay work.

## Product questions to resolve in a mediated GrillMe session

- What can initiate a focus handoff: agent request, user command, terminal event from the lifecycle journal, or all three?
- What is the first surface: Viewer board camera, a floating overlay lens, desktop window focus, or a staged sequence?
- What must require an explicit confirmation before moving focus or changing desktop windows?
- How should the user preserve orientation: smooth pan/zoom, a preview card, an undo/return affordance, and accessibility alternatives?
- What durable state makes the conversation shared: active target, reason, source event, viewport/crop, acknowledgement, and expiry?
- How should the voice agent describe the request briefly without flooding the conversation?

## Desired initial interaction

1. A relevant work event arrives, or the agent asks the operator to inspect something.
2. The agent says why and offers a focus action.
3. The operator accepts, declines, or asks for a preview.
4. The Viewer brings the target into view with a controlled transition and a clear return action.
5. The conversation receives an acknowledgement so both sides have the same context.

## Constraints

- Preserve user control. No surprise desktop focus changes, clicks, input, or application launch.
- Keep journal-driven suggestions bounded; reuse the rate-limited digest policy in #686.
- Keep initial scope independent of arbitrary desktop control. It must work for Viewer-native targets before adding Computer Use window focus.
- Do not duplicate #183's board-wide navigation work or #346's overlay shell.

## Deliverable

An Opus architecture note and a decision log produced through a mediated GrillMe session: the architect asks one material question at a time, the root voice agent relays it to the operator, and the agreed answer is sent back to the architect. No implementation in this issue until the design is accepted.

Latand:
## Design summary — implementation-ready (architecture lane complete)

A fresh-context architecture lane produced this through a mediated question-and-answer session with the operator. Decisions below are recorded, not proposed. #691 is a separate artifact; this one references it only through the shared contract.

### Decisions

| | Decision |
|---|---|
[...]

PR #711, opened 2026-07-26
feat(attention): wire the #688 handoff loop end to end — mount, tool, navigate, return

Latand:
Closes the last mile of #688. The record, the machine, the store, the target model and `/api/attention` all worked; nothing used them.

## What was actually wrong

The entire #688 client surface — `RootOverlayHost`, `RootOverlay`, `RootSheet`, `useAttentionOffers` and `buildFocusFrameIndex` — was imported by nothing but its own tests. No shipped page polled `/api/attention`, and the deployed bundle contained none of that path. A live request sat at `state=pending` with `offeredTo: []` for its full 10-minute TTL, which is the proof: `useAttentionOffers` documents that *rendering* an offer is the `pending → offered` transition, precisely so "nobody saw it" and "they saw it and said nothing" stay different facts. An empty `offeredTo` at expiry means no surface ever rendered it.

There was also no agent-facing tool, so the root agent could not raise a request through its normal surface — the only live requests were created by raw HTTP. And accepting could not have navigated anyway: camera movement was deliberately left unstarted, so `resolveFocusTarget` was reachable only from tests.

Nothing here changes the record, the machine, the store or the routes.

## The four pieces

**Mounted.** `AttentionHost` (`src/components/attention/AttentionHost.tsx`) polls the record for one stable per-device id — the same `stableDeviceId` presence already uses, so "which device was offered this" and "which device is looking at what" name the same thing. It renders the existing overlay the moment there is something to answer and nothing at all otherwise; the root conversation's own turns need a transcript-feed adapter and stay their own slice, but an unanswered request being invisible is the failure this ends.

The hidden-surface contract is untouched and is now tested *through* the mount rather than only at the hook: a background tab reads the record, renders nothing, reports nothing, and the request expires with `offeredTo: []`. The operator's Viewer is frequently backgrounded, so that is the common case.

**A tool.** `request_attention`, following the conventions already in `bindings.ts` and `server.ts`: `clientRequestId` idempotency, the established `mutationReceipt` shape, and registration as a mutating tool so the receipt is durable — a bounded one would let a retry after a restart ask the operator twice. It only asks: the request lands at `pending` and nothing moves until they agree on a device. No `rootId` appears in the schema, because the service resolves the root identity itself.

The server derives which project a target lives in (from the scanner, the pipelines, the flows or the tasks) and refuses a target it cannot attribute rather than filing it under a guess. A board draft exists only on the operator's canvas, so it needs an explicit `project`.

**Navigation.** One bus (`focusHandoffBus`), the same idiom `viewPresenceBus` already uses for the "leaf publishes, shell consumes" problem: `SchemeBoard` publishes its live frame index plus its two camera moves, and `Viewer` publishes project and conversation opening. `runFocusHandoff` opens the target's project if it is not the open one, waits for that board's layout, resolves the anchor against the board **as it is now** — never the rect stored at creation, since the board reflows — and glides there with zoom following intent.

`MobileFocusView` registers the same way but arrives by pinning the pane: there is no camera on a phone and one pane is on screen at a time. A destination it has no pane for (a coordinate, or the space a vanished card used to occupy) is **refused**, and the handoff reports `lost` rather than recording an arrival that did not happen.

**Return.** The viewport is captured on the answering device immediately before anything moves, because the move can open another project and capturing at arrival would record the destination. Returning restores that camera exactly, and does not re-open the focused path — that would glide the board to the node and undo the framing just restored. On a surface with no camera the focused path is what comes back, which is all of that viewport it can put back.

## The secondary gap

`adoptRootSession` was never called from app code, so `root-lineage.json` carried a minted `rootId` over `"sessions": []` — a stable identity with no chain to survive a rollover with. `adoptLiveRootSession` resolves the root the way the rest of the app already names it (the configured conversation id, else the `root` role the registry stamped at admission) and the raise path folds it in. A rollover appends a linked successor under the same identity, so a request raised before it still resolves after.

## One judgement worth a second look

A request raised through the tool has no board in front of it, so the frame it records is deliberately empty. `frameWasRead` refuses to degrade to such a frame: standing in for a vanished anchor with it would drop the operator at world (0,0) and call it "where it was", which is the silent arbitrary landing the design forbids. Those requests report `lost` instead — visibly, since the record refuses to call it a follow and the action row says so. The alternative would be for the tool to carry geometry it has no way to know.

## Design decisions honoured as given

- Only the root may surface a request, and it is still enforced server-side: `rootIdentity()` is resolved in `service.ts` and never accepted from the caller. A worker holding this tool cannot name a root of its choosing. Worth stating plainly: the MCP transport carries no caller identity, so the guarantee is over *which* root is named, not over which agent may call.
- A hidden or background surface reports nothing.
- Every request declares show vs open, and that reaches the operator before they agree.
- Anchor and frame stay separate; a vanished anchor resolves to approximate or lost.
- The return point is per-device, captured on the device that accepted.

Not built here: #691's wider overlay redesign, the Computer Use seam, standing-consent storage, and the root-transcript feed.

## Verification

- Focused tests, by path, against an isolated state dir: **197 pass, 0 fail** across `src/components/attention/`, `focusFrames.test.ts`, the two `MobileFocusView` suites, `src/components/overlay/`, `src/lib/attention/`, `src/lib/root/`, `src/app/api/attention/`, and the four `src/lib/mcp/` suites.
- `./node_modules/.bin/tsc --noEmit` — clean.
- `bun run lint` — 11 errors + 1 warning, all pre-existing (`scripts/*.cjs` require-imports, `reaperRuntime.ts` unused var); unchanged from the merge base.

Each new test was checked by reverting its fix and watching it go red: stubbing the mount fails 4 host tests; removing the camera move fails 6; removing the camera restore fails the 2 return tests; dropping `adoptRootSession()` from the raise path fails the adoption assertion; making `liveRootSession` resolve nobody fails 4 lineage tests; loosening `frameWasRead` fails the origin-landing test; unregistering the tool fails all 8 tool tests.

### DIFF
--- a/README.md
+++ b/README.md
@@ -225,7 +225,11 @@ The MCP surface includes:
 - operator/runtime reads: `operator_snapshot`, `deployment_status`, and
   `resources`;
 - agent/runtime mutations: `spawn_agent`, `conversation_migration`, and
-  `deploy_exact_sha`.
+  `deploy_exact_sha`;
+- the operator's attention: `request_attention`, which offers to move their
+  Viewer to a target and waits for their answer. It only asks — nothing moves
+  until they agree on a device, and the request names the root agent by an
+  identity the server resolves, never one the caller supplies.
 
 Every call requires a stable `clientRequestId`. Reusing that id with the same
 arguments returns the durable result as a replay. Reusing it with different
--- a/src/components/Viewer.attentionMount.test.ts
+++ b/src/components/Viewer.attentionMount.test.ts
@@ -1,0 +1,47 @@
+import { expect, test } from "bun:test";
+import fs from "node:fs";
+import path from "node:path";
+
+/*
+ * The shell renders the attention host (#688 D3).
+ *
+ * This is the defect class the mount closed: a complete feature — record,
+ * machine, route, overlay, board seam — with a green suite and no surface,
+ * because every test mounted the pieces directly and nothing asserted that the
+ * shipped shell rendered any of them. So this asserts the one thing those tests
+ * cannot: that `Viewer` still contains the host.
+ *
+ * It reads the source rather than rendering, deliberately. The host renders
+ * NOTHING until there is something to answer — that contract is asserted in
+ * `AttentionHost.dom.test.tsx` — so a rendered tree is empty in exactly the
+ * state a shell test could set up, and would go on being empty after the mount
+ * was deleted. The import graph is the part that can actually be checked.
+ */
+
+const source = fs.readFileSync(path.join(process.cwd(), "src/components/Viewer.tsx"), "utf8");
+
+/** The file with comments removed, so a mount that survives only inside a
+    comment does not read as a mount. */
+const code = source
+  .replace(/\/\*[\s\S]*?\*\//g, "")
+  .replace(/^[ \t]*\/\/.*$/gm, "");
+
+test("the shell imports the attention host", () => {
+  expect(code).toMatch(/import\s*\{[^}]*\bAttentionHost\b[^}]*\}\s*from\s*"\.\/attention\/AttentionHost"/);
+});
+
+test("the shell renders the attention host, and does not gate it behind a flag", () => {
+  const mounted = code.match(/<AttentionHost\b[^>]*\/>/);
+  expect(mounted).not.toBeNull();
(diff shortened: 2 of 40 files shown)


Latand:

PR #715, opened 2026-07-26
feat(audio): the approved earcon set and ambient bed, wired to real events

Latand:
Wires the approved earcon set and the approved ambient bed into the viewer's
real events, with de-duplication, a bounded overlap cap, device-local settings
and autoplay-safe activation.

## Bundled assets

Seven masters from the approved 2026-07-26 set (authored outside this repo),
copied **byte-for-byte** and renamed to neutral, public-safe names. Nothing was
re-encoded, re-normalized or per-cue gain-adjusted.

| cue / role | master (approved set) | bundled as |
|---|---|---|
| attention  | `hierarchy/attention-v1.mp3`  | `public/audio/cues/attention.mp3` |
| failure    | `hierarchy/failure-v2.mp3`    | `public/audio/cues/failure.mp3` |
| launch     | `hierarchy/launch-v3.mp3`     | `public/audio/cues/launch.mp3` |
| success    | `hierarchy/success-v3.mp3`    | `public/audio/cues/success.mp3` |
| viewer-mcp | `hierarchy/viewer-mcp-v6.mp3` | `public/audio/cues/viewer-mcp.mp3` |
| tool-tick  | `hierarchy/tool-tick-v4.mp3`  | `public/audio/cues/tool-tick.mp3` |
| ambient bed | `ambient-loop-crossfade.wav` | `public/audio/ambient/voice-call-loop.wav` |

The bed stays **lossless PCM**: every lossy codec adds encoder priming and
padding, and a loop buffer carrying that padding clicks on every pass. It is
5.1 MB — the one cost of a seamless loop, and it is not in the npm package
(`files` ships `bin` and `dist` only).

## Cues bound to real events

| cue | event | where it comes from |
|---|---|---|
| attention | an agent leaves `live` for a waiting/question state | the polled lifecycle scan |
| attention | an attention offer is first rendered on this device | the #688 attention record (`request.id`) |
| success | a run finishes and comes home | the polled lifecycle scan |
| failure | a run stalls | the polled lifecycle scan |
| failure | a spawn receipt comes back `launched: false` | the spawn POST outcome (`launchId`) |
| launch | a spawn receipt is accepted | the spawn POST outcome (`launchId`) |
| launch | a new node joins the agent tree | the polled lifecycle scan |
| viewer-mcp | a Viewer MCP call is in flight | the parsed transcript (`ToolEvent.mcp`, engine call id) |
| tool-tick | any other tool call is in flight | the parsed transcript (engine call id) |

Only calls the transcript shows **running** tick. Settled calls are history, so
opening an old conversation or pressing «show earlier» stays silent.

## Properties

- **De-duplication** on the stable identity each event already carries — a call
  id, a launch id, an attention request id, a conversation identity plus the
  transcript mtime that names the transition. Every source is observed
  repeatedly by construction (a 4s poll, a re-parse, a remount, a reconnect
  replay), so an identity is spent the first time it is considered, whatever the
  outcome. Nothing accumulates a backlog to flush later.
- **Bounded overlap**: three simultaneous voices, routine texture coalesced to
  one voice per kind, and at the cap louder news displaces quieter news. Twenty
  tool ticks in a burst are one sound, never twenty and never a late tail.
- **Loudness**: the masters already encode the approved hierarchy. One
  documented tier table sits on top — priority 1.0 (attention, failure), salient
  0.75 (launch, success), routine 0.5 (viewer-mcp, tool-tick) — preserving the
  ordering, so a priority cue is never quieter than a routine one. The bed is
  capped below every cue tier at any slider setting.
- **Autoplay**: one shared `AudioContext` with the existing chimes, unlocked on
  the first real gesture. Before that the transport answers `null` and cues are
  dropped in silence — nothing throws, logs or queues.
- **Device-local settings**: `localStorage` only, no route and no shared record.
  The master switch reuses the key the header toggle already owned, so one
  switch governs the earcons and the bed alike.

## The ambient bed

Off on first run. Once this device enables it, a connected call brings it back
on its own — persisted enable plus `phase === "live"` is all it takes, subject
to the autoplay unlock. It fades in on connect and out on disconnect, ducks
smoothly under either participant's speech (read from the transcript the call
already produces, no analyser on either stream) and under attention/failure,
and recovers smoothly. Looping is gapless by construction: one decode, then
`AudioBufferSourceNode.loop` with `loopStart`/`loopEnd` on the audio clock — no
`<audio>` element, no `ended` re-trigger, no timer-scheduled next pass.

The "no loop asset configured" path stays supported and tested: the loop
controls are absent from the UI rather than shown disabled.

## Verification

- Focused behavioural tests, run by explicit path in a temp sandbox: 80 tests
  across 7 files, all green. Each is RED-verified — reverting dedupe, the cap,
  the tier table, the gapless loop wiring, the fades, the master gating, the
  speech duck, the asset binding or the in-flight rule fails the tests that
  claim them.
- `tsc --noEmit` clean; eslint clean on every changed file (the repo's 11
  pre-existing `scripts/` errors are untouched).
- Privacy gate: **PASS** for all 26 changed source/test/i18n paths. The seven
  binary audio assets cannot be classified by the gate at all — see the note on
  the PR conversation.


Latand:
### Privacy gate: source PASSES, the seven audio assets cannot be classified

`bun scripts/privacy-publication-gate.ts --base d202a16d… --paths <26 changed source/test/i18n paths>` → **PRIVACY GATE: PASS**

The whole-diff run **FAILS**, entirely on the seven binary audio files:

```
public/audio/ambient/voice-call-loop.wav   FAIL  credential: 1  inspection_error: 1
public/audio/cues/attention.mp3            FAIL  inspection_error: 1
public/audio/cues/failure.mp3              FAIL  inspection_error: 1
public/audio/cues/launch.mp3               FAIL  inspection_error: 1
public/audio/cues/success.mp3              FAIL  inspection_error: 1
public/audio/cues/tool-tick.mp3            FAIL  inspection_error: 1
public/audio/cues/viewer-mcp.mp3           FAIL  inspection_error: 1
```

This is structural, not a property of these files:

- the gate knows three media kinds — `png`, `raster`, `animated`. Audio matches
  none, so it falls through to the **text** inspector, which classifies any
  non-text-like binary as `inspection_error` by construction. Every audio file
  fails, always;
- the WAV additionally trips one `credential`: the credential-**input** regex
  (an HTML `<input type=password value=…>` pattern) matching inside a UTF-16
  reinterpretation of PCM noise. A false positive on random bytes;
- the designed escape hatch does not reach: a `privacy-manifest.json` requires
  `matchesTrustedReproduction` — an in-repo deterministic generator that
  reproduces the asset byte-for-byte. These are authored, operator-approved
  masters; no such generator exists, and regenerating them was explicitly out of
  bounds.

I did **not** modify the gate. Teaching it to accept an entire new asset class
is a change to a security control, and it should not be made unreviewed inside a
feature PR.

Independent evidence that the assets carry nothing private — the check the gate
would have made if it could read them:

```
$ ffprobe -show_entries format_tags:stream_tags   # all seven
attention.mp3  … viewer-mcp.mp3   TAG:encoder=Lavc62.28 / TAG:encoder=Lavf62.12.102
voice-call-loop.wav                TAG:encoder=Lavf62.12.102
```

The only metadata key present in any of the seven is `encoder`. No title, no
artist, no comment, no path, no identity. All seven are byte-identical
(`cmp`-verified) to the approved masters.

PR #719, opened 2026-07-26
fix(audio): tick every newly appended tool call, however fast it settles

Latand:
## Problem

A short tool call settles inside one tail tick: the frontend's first parse already sees it with `ok`/`err` attached. The cue logic in `src/lib/audio/toolCues.ts` gated the generic tool cue on `status === "run"`, so those calls never rang. Production symptom: agents doing quick tool work sound silent.

## Contract implemented

- Every newly appended tool call in the watched conversation produces exactly one generic tool cue, regardless of tool kind and regardless of whether call and output arrive in one update.
- Viewer MCP calls keep their distinct cue.
- Loading, remounting, or paging old settled history stays silent.
- Duplicate parses, multiple panes, and replay never double-ring.

## Approach

`createToolCueScanner` (per pane, per transcript path) replaces the pure per-snapshot read. It keeps one number — the absolute line index just past the last heard window — and rings any tool event whose `srcCall` lies at or past that floor, whatever its status:

- The first loaded parse is the baseline and absorbs the whole window silently; a call still running at baseline is happening now and still ticks (previous behavior preserved).
- Paged-in history (`show earlier`) sits below the floor and stays silent; a session reset re-parse produces the same `srcCall`s and stays silent.
- A truncation restart moves the floor back down with the window, so the pane is not deaf afterwards.
- The tab-wide cue player still owns what actually sounds by de-duplicating on `tool:<conversation>:<callId>` — remounts, second panes and replays re-request and get `deduped`.

`useToolActivityCues` gates the baseline on the tail having loaded (an unloaded feed is not an empty conversation) and skips the one mixed commit a pane switch renders, where the new conversation key is still paired with the old file's window.

## Parser coverage inspected

Codex `function_call` (shell and generic), `custom_tool_call`, and Claude `tool_use` all flow through `newToolEvent` → `srcCall = curSrc`, with outputs attached by call id (`function_call_output`, `custom_tool_call_output`, `tool_result`). The feed-level tests exercise all three shapes through `createFeedSession`.

## TDD evidence

RED (scanner stubbed to today's status-gated semantics — the exact production bug):

```
expect(scanner.scan(appended, 12)).toEqual([...])
- [ { "cue": "tool-tick", "eventId": "tool:conv-1:fast-1", "pan": 0 } ]
+ []
 8 pass / 1 fail — src/lib/audio/toolCues.test.ts
```

GREEN, then lifecycle/replay cases added one at a time (baseline silence, prepend silence, in-flight baseline, failed fast call, MCP cue, folded cmd-group, truncation restart, and four feed-level cases against the real parser):

```
bun test src/lib/audio/toolCues.test.ts src/lib/audio/toolCues.feed.test.ts \
  src/lib/audio/cuePlayer.test.ts src/lib/audio/app.test.ts \
  src/components/feed/toolLifecycle.parse.test.ts
 45 pass / 0 fail
```

## Checks

- `bunx tsc --noEmit` — clean
- `bunx eslint` on the five changed files — clean
- `bun scripts/privacy-publication-gate.ts --base c4494bbb` — PRIVACY GATE: PASS
- Only exact focused test files were run; no broad live-state suites.

Blocker note: this PR gates the next production deployment. Do not deploy without it.

### DIFF
--- a/src/components/LogFeed.tsx
+++ b/src/components/LogFeed.tsx
@@ -312,10 +312,13 @@ export function LogFeed({ file, showSvc, lineFilter, onStatus, paused, follow, s
     // eslint-disable-next-line react-hooks/exhaustive-deps
     [session, file?.activity, tail.lines, tail.linesStart],
   );
-  /* Tool activity earns its cue from the parse itself: an in-flight call ticks
-     once, keyed on the engine's call id, so a tail tick that re-parses the same
-     window is silent and history never replays. */
-  useToolActivityCues(feed.items, memoryKey);
+  /* Tool activity earns its cue from the parse itself: every newly appended
+     call ticks once, keyed on the engine's call id — even one that settled
+     inside a single tail tick — while re-parses, remounts and paged-in history
+     stay silent. The window end anchors "newly appended" to the tail stream;
+     loading gates the baseline so an unloaded feed is not mistaken for an
+     empty conversation. */
+  useToolActivityCues(feed.items, memoryKey, file?.path ?? null, tail.linesStart + tail.lines.length, Boolean(file) && !tail.loading);
   const hiddenLocal = Math.max(0, feed.items.length - visibleCount);
   const visibleItems = hiddenLocal ? feed.items.slice(-visibleCount) : feed.items;
   const visibleStartIndex = feed.items.length - visibleItems.length;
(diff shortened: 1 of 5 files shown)


Latand:

Issue #723, opened 2026-07-26
perf(scanner): prioritize cold transcript scans and defer archival work


Latand:
## Context

Cold Viewer startup can discover thousands of transcript files, including multi-gigabyte journals. A complete synchronous pass delays the first useful board projection and can exceed bounded runtime health checks.

This issue tracks phased scan scheduling as a follow-up to the immediate incremental scan-cache repair.

## Proposed scheduling

1. Return the last completed durable scan index immediately.
2. Prioritize live and recently modified conversations.
3. Process new or changed small transcripts early.
4. Read bounded head/tail windows for large active transcripts.
5. Defer large inactive archives to a background queue.
6. Publish refreshed board entries incrementally as deferred work completes.

Priority should combine activity, recency, known board membership, change status, and file size. File size alone must not delay an active agent.

## Requirements

- Preserve deterministic ordering for equal-priority work.
- Keep transcript reads bounded; never load a complete large JSONL into memory.
- Avoid starvation of deferred archival files.
- Preserve deleted-worktree grouping and conversation lineage.
- Keep health probes and the first board projection independent from archival completion.
- Add cold-start performance coverage with many files and at least one multi-gigabyte sparse fixture or equivalent bounded-read witness.
- Record timing by scan phase so regressions identify discovery, metadata, activity, or lineage work.

## Acceptance criteria

- The first useful board projection is available within the existing runtime health budget on a production-shaped archive.
- Active conversations appear during the first phase, including conversations backed by large transcripts.
- Deferred entries converge without a manual refresh.
- Tests prove bounded reads, deterministic priority, fairness, and incremental publication.

## Scope boundary

The immediate deployment repair should only reconnect `board_snapshot` to the shared incremental scan-cache and remove the redundant full refresh. This issue owns phased prioritization and background hydration.

Latand:
Closing: step 1, answering from the last completed scan index, exists as the persisted scan snapshot (`src/lib/scanner/scanCache.ts:104`, read by `src/app/api/files/route.ts:350`). The remaining restart cost is measured and tracked in #1808.

PR #724, opened 2026-07-26
fix(mcp): reuse completed scan for board snapshot

Latand:
## Summary

- route MCP `board_snapshot` through the shared completed scan generation
- serve the durable cached projection immediately while the ordinary refresh continues in the background
- prevent deployment health probes from starting a redundant full transcript inventory scan

## Root cause

The MCP binding called raw `listFiles({ fresh: true })`, bypassing the scan cache used by the Viewer files surface. On the production-shaped archive the cold call took 24.1 seconds and exceeded the 15-second MCP health budget.

## Evidence

- RED: focused binding test failed when raw scanning was forbidden
- GREEN: focused binding suite passes, 17/17
- typecheck passes
- privacy publication gate passes
- production-shaped cold-process measurement: 698 ms for `board_snapshot(limit: 1)`

Follow-up phased scheduling and archival hydration are tracked in #723.

### DIFF
--- a/src/lib/mcp/bindings.test.ts
+++ b/src/lib/mcp/bindings.test.ts
@@ -327,18 +327,34 @@ test("link_task_to_pipeline follows the cursor retry after a fail-edge loop-back
 
 test("board_snapshot returns an inert bounded board projection with durable lineage and redaction", async () => {
   let writes = 0;
+  let rawScans = 0;
   const credentialLabel = ["api", "key"].join("_");
   const titleFixture = ["super", "secret"].join("-");
   const bindings = viewerMcpBindings(undefined, undefined, {
-    listFiles: async () => [{
-      path: "/sessions/worker.jsonl",
-      project: "viewer",
-      title: `Audit ${credentialLabel}=${titleFixture}`,
-      engine: "codex",
-      activity: "live",
-      proc: "codex",
-      conversationId: "conversation_worker",
-    }],
+    listFiles: async () => {
+      rawScans += 1;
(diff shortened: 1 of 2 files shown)


Latand:
Heads-up: the `test(privacy): pin temporary repository default branch` commit here (cea920a1) is the fix for the repo-wide CI outage now tracked as #733 — `privacy-publication` has been red on **every** PR since #722 merged on 2026-07-26, because the failing step runs the suite from the `trusted` checkout of the default branch rather than the candidate.

Reproduced with `init.defaultBranch=master`: 105 pass / 5 fail before, 110 pass / 0 fail after — matching CI exactly.

Consequence for this PR: its own `privacy-publication` check cannot go green, because that check still runs main's copy. It goes green for everyone only once the pin is on main, so merging needs a deliberate override.

I opened #734 with the same one-line fix in isolation before noticing this one. Only one of the two should land — this PR is the pre-existing one.

Latand:
Rebased onto main after the CI fixture fix landed in #734 (merge `baf3138f`).

The `test(privacy): pin temporary repository default branch` commit was byte-identical to what merged, so the rebase dropped it as already-applied. Nothing of this PR's own work was touched: it is now exactly one commit, `fix(mcp): reuse completed scan for board snapshot` (`97564023`), touching only `src/lib/mcp/bindings.ts` and its test.

This PR is now clean for review on its own merits. `privacy-publication` should also go green here on the next run — it was red for repo-wide reasons, not because of this branch.

PR #734, opened 2026-07-27
fix(ci): pin the commit-message fixture repo to the main branch

### LINKED ISSUE #733: ci: privacy-publication is red on every PR — the gate's own suite fails on the runner
## Problem

The `privacy-publication` check is red on **every** pull request, including ones
whose diff the gate itself passes. It currently blocks merging PR #730 and the
reconciliation of production with `main`.

The failing step is:

```yaml
- name: Verify privacy gate behavior
  working-directory: trusted
  run: bun test scripts/privacy-publication-gate.test.ts
```

`trusted` is a checkout of the **default branch**, not the candidate. The step
therefore runs `main`'s own test suite against `main`'s own code, and its result
is independent of the pull request under review. A PR cannot fix it and cannot
avoid it.

## Observed

`105 pass, 5 fail` on the runner. All five failures are in the
`commitMessageFindings` group:

- `flags a personal email in a Co-Authored-By trailer`
- `flags a home path in a commit message`
- `passes when commit messages contain no sensitive data`
- `a GitHub no-reply address is still an account handle and is still flagged`
- `the carve-out is the trailer only, never the body`

One reported assertion:

```
const handle = ["4242+someone", "users.noreply.github.com"].join("@");
commit(repo, `feat: something\n\nCo-Authored-By: Someone <${handle}>`);
const findings = commitMessageFindings(repo, "main");
expect(findings.has("email_address")).toBe(true);
// Expected: true   Received: false
```

Note that both directions fail: a case expecting a finding gets none, and the
case expecting a clean result also fails. That rules out a simple
"scanner returns nothing" explanation.

## Not reproducible locally

The same suite passes locally at `110 pass, 0 fail` under three variations:

- plain run
- with `PRIVACY_BASE_SHA` set to the base commit, as the workflow sets it
- with `TMPDIR` pointed at a path under a home directory, to mimic the runner's
  temp location

So the trigger is something else in the runner environment. The tests build a
throwaway repository per case with `git init -b main`, a local
`user.name`/`user.email`, an initial commit, then `git checkout -b feature`, and
compare `main..HEAD`. Candidate suspects worth checking on a runner: git
version differences (runner is 2.54.0), commit signing or hook configuration,
same-millisecond filenames in the `commit()` helper causing an empty commit, and
any ambient git configuration the fixtures inherit.

## Required

- Make the suite pass deterministically on a GitHub runner, or isolate the
  fixtures from whatever host state they currently depend on.
- Keep the check's protective intent intact: it must still fail when a commit
  message really does carry an address or a home path.
- Add whatever isolation the fixtures need (explicit git environment, unique
  commit content) so the result cannot depend on the host.

## Acceptance

- `privacy-publication` is green on a pull request that the gate passes.
- The suite still fails on a deliberately seeded sensitive commit message.
- The fix runs identically on a runner and on a developer machine.

Latand:
Fixes #733.

## Cause

The `commitMessageFindings` fixtures build a throwaway repository with
`git init` and then ask for findings across `main..HEAD`:

```ts
Bun.spawnSync({ cmd: ["git", "init", repo], ... });
// ...
const findings = commitMessageFindings(repo, "main");
```

`git init` names the initial branch after `init.defaultBranch`. On a machine
configured with `main` the fixture works; on a GitHub runner the default is
`master`, so `main` does not exist, the range resolves against nothing, and
five assertions collapse — including `passes when commit messages contain no
sensitive data`, which is why the failure did not look like a simple
"scanner found nothing".

## Why every PR was red

The failing step runs from the **trusted** checkout of the default branch:

```yaml
- name: Verify privacy gate behavior
  working-directory: trusted   # ref: default_branch
  run: bun test scripts/privacy-publication-gate.test.ts
```

Its result never depended on the pull request under review. That has two
consequences worth stating plainly:

1. Every open PR has been red since the fixtures landed in #722, regardless of
   content. The last green run was 2026-07-26T22:06, before that merge.
2. **This PR will also show the check red.** It cannot do otherwise — the step
   is still running the default branch's copy. The check goes green for every
   PR once this is on the default branch, and not before. Merging it therefore
   needs a deliberate override.

## Fix

Create the fixture repository on `main` explicitly, so the comparison branch
exists regardless of host git configuration.

## Verification

Suite run with `init.defaultBranch=master` to reproduce the runner:

| | result |
|---|---|
| before, runner-like default | 105 pass / **5 fail** — matches CI exactly |
| after, runner-like default | 110 pass / 0 fail |
| after, ordinary default | 110 pass / 0 fail |

The seeded-sensitive-message cases still detect their findings, so the check
keeps its protective intent rather than being silenced.

### DIFF
--- a/scripts/privacy-publication-gate.test.ts
+++ b/scripts/privacy-publication-gate.test.ts
@@ -2674,7 +2674,7 @@ describe("commitMessageFindings", () => {
   function gitRepo(): string {
     const repo = mkdtempSync(join(tmpdir(), "llv-privacy-commit-"));
     temporaryDirectories.push(repo);
-    Bun.spawnSync({ cmd: ["git", "init", repo], stderr: "pipe", stdout: "pipe" });
+    Bun.spawnSync({ cmd: ["git", "init", "-b", "main", repo], stderr: "pipe", stdout: "pipe" });
     Bun.spawnSync({ cmd: ["git", "-C", repo, "config", "user.email", "test@example.com"], stderr: "pipe", stdout: "pipe" });
     Bun.spawnSync({ cmd: ["git", "-C", repo, "config", "user.name", "Test"], stderr: "pipe", stdout: "pipe" });
     writeFileSync(join(repo, "init.txt"), "init");


Latand:
Duplicate alert: #724 already carries a byte-identical fix (`test(privacy): pin temporary repository default branch`, cea920a1). I opened this before spotting it.

#724 is the pre-existing PR and is also small (3 files, +32/-11), bundling the pin with a scoped `src/lib/mcp/bindings` change. This PR is the isolated one-liner with the reproduction written up in #733.

Either unblocks CI. Only one should merge — operator's call. Diagnosis and the RED/GREEN reproduction live in #733.

Latand:
Merged as `baf3138f` with a deliberate, operator-authorized required-check override: `privacy-publication` could not pass on this PR by construction, because that step runs the suite from the default branch's checkout.

Override mechanism, for the record: `enforce_admins` was disabled on `main` for the duration of the merge and restored immediately afterwards. Required contexts (`privacy-publication`, `privacy-tracker-audit`) and `strict` were left untouched throughout.

Confirmed recovered: re-running the previously failing check on #730 now passes — https://github.com/Latand/live-log-viewer-next/actions/runs/30252900826/job/89954842893 — the first green `privacy-publication` since 2026-07-26T22:06.

The duplicate CI surface in #724 was removed by rebase rather than by closing it; that PR keeps its own `src/lib/mcp/bindings` work and stays open for review.

PR #735, opened 2026-07-27
feat(audio): continuous background music in the Viewer and across the call boundary

### LINKED ISSUE #732: feat(audio): continuous background music in the Viewer and across the call boundary
## Problem

The Audio settings expose a single ambient toggle whose Ukrainian label,
`Фоновий шар під час дзвінка`, reads as an engineering term ("background
layer") rather than as music, and there is no way to have background music
while using the Viewer outside a call. Ambient audio is therefore an
all-or-nothing property of being in a call.

## Required behavior

Two independent, composable checkboxes sharing **one** volume control:

1. **Background music in the Viewer** — plays while using the Viewer outside a
   call.
2. **Background music during a call** — the renamed existing setting.

### Label

Rename the call setting to natural music wording — `Фонова музика під час
дзвінка` — and give the new setting a matching name. Keep English and Ukrainian
at full parity; neither locale may carry a key the other lacks.

### Toggle matrix

| In Viewer | In call | Behavior |
|---|---|---|
| on | on | **One continuous track across the call boundary.** It plays in the Viewer, continues into the call with no restart and no position reset, ducks under speech for the duration, and keeps playing after hangup. No teardown or re-init at either edge. |
| on | off | Plays outside calls. Call start fades it out and pauses it, retaining position; hangup fades it back in from that position. |
| off | on | Silent outside calls. Call start fades in from the retained position (so repeated calls do not replay the same opening), ducking applies; hangup fades out. |
| off | off | Silent everywhere. |

### Invariants

- **Ambient ownership across conversation cards is preserved.** Switching the
  selected conversation must not restart, duplicate, or drop the track — the
  existing owner-token model stays as-is.
- Speech ducking keeps its current behavior; it applies whenever a call is
  live and music is audible, regardless of which checkbox enabled it.
- One shared volume control governs both settings.
- Position and playback continuity survive the call boundary in both
  directions.

## Acceptance

- A deterministic test drives each of the four toggle combinations and asserts
  playback state, fade direction, and retained position at both the call-start
  and hangup edges.
- A test proves the enabled/enabled path crosses the boundary with **no**
  restart and no position reset.
- A test proves conversation-card switching does not disturb ambient ownership.
- English and Ukrainian key sets are asserted equal.
- Focused audio tests, TypeScript, changed-file lint, and the privacy
  publication gate pass.

## Out of scope

Deployment. This feature ships to production only on a separate explicit
instruction.

Latand:
Closes #732.

Two independent, composable switches in Audio settings sharing **one** level:
background music while using the Viewer, and background music during a call —
the latter the renamed existing ambient setting, whose Ukrainian label read as
an engineering term («Фоновий шар під час дзвінка») rather than as music. It is
«Фонова музика під час дзвінка» now, with the Viewer one matching it, and the
level label moved from «Рівень фону» to «Рівень музики».

## The toggle matrix

| In Viewer | In call | Behavior |
|---|---|---|
| on | on | One track across the boundary. Nothing is torn down, re-inited or repositioned at either edge; ducking applies for the duration of the call. |
| on | off | Plays outside calls. Call start fades it out and **parks** it at its position; hangup resumes from there. |
| off | on | Silent outside calls. Call start fades in from the retained position, so a repeat call does not replay the same opening; hangup parks it again. |
| off | off | Silent everywhere. |

The distinction that makes this work is **park vs tear down**. While either
switch still wants music, the silencing edge fades the voice out and keeps the
position it reached (`LoopVoiceHandle.pause` reports it; `LoopTransport.start`
takes an `offsetSeconds`). Only a device that wants no music at all — sound
master off, both switches off, or no asset — stops the track and forgets where
it was, so the next unmute opens it from the top rather than resuming a position
nobody chose.

Ambient ownership across conversation cards is untouched. The lease stays per
mounted composer; it is now extracted as `useAmbientCallLease` in
`useCodexRealtime.ts` and covered by a test that mounts two cards in the order
React actually uses (next card mounts before the last unmounts).

## Tested

Test-first: each of the four matrix rows, and the parity/no-restart properties,
were written and observed RED before the implementation existed.

- `bun test src/lib/audio/ src/lib/i18n/ src/components/SoundToggle.dom.test.tsx src/hooks/useCodexRealtime.ambient.dom.test.tsx src/hooks/useAgentChimes.test.ts` — 139 pass, 0 fail.
- **The no-restart property is mutation-checked.** Each fake voice carries the
  ordinal of the `start` that created it, so the on/on test asserts one start,
  no pause, no stop, and every ramp against voice 1. Temporarily making
  `setVoiceConnected` re-create the voice at the boundary *while preserving the
  position* — the sneakiest possible regression — failed exactly that test and
  nothing else.
- The card-switch ownership tests were mutation-checked the same way: replacing
  the owner-set count with the raw `connected` flag failed the two new
  switching tests plus the pre-existing lease test, and nothing else.
- `bun run lint` on the 15 changed files + the new one — clean.
- Privacy publication gate against the pinned base
  (`--check-commits --base origin/integrate/voice-reload-focus-20260727`), both
  with and without `--require-known-values` — PASS.

Suites were run by path only; no runtime or registry directory was swept.

## Decisions the issue left open

1. **What a full teardown does to the retained position.** The issue specifies
   parking for the call edges but not for a master mute. A mute clears the
   position: it is a global "no music", not an intermission, so unmuting opens
   the track rather than resuming somewhere the operator never chose. The call
   edges are the only thing that parks.
2. **Where the parked position is sampled.** At the moment `pause` is called,
   i.e. at the *start* of the 1.5s fade-out, so a resume re-enters the last
   1.5s — material that was fading to silence anyway. Sampling after the fade
   would need a timer on the audio clock for no audible gain.
3. **Retry budget.** Music in the Viewer is wanted from the first paint, which
   is long before the gesture that unlocks audio, so the bounded start-retry
   budget could burn out while waiting for exactly that gesture. The gesture and
   any settings change now reset it — otherwise a device with the setting on
   would sit silent until something else happened to poke the bed.
4. **English wording.** "Background music in the Viewer" / "Background music
   during a call", and the shared slider renamed from "Ambient level" to "Music
   level" to match; the hint now covers both switches and states the
   across-the-call continuity.

Not deployed, per the issue's out-of-scope note.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -105,7 +105,10 @@ versions follow [SemVer](https://semver.org/) (0.x — the API may still move).
   (the sound master off, both switches off, no asset) tears the track down. The
   music ducks under whoever is talking, read off the transcript the call already
   produces, with the duck owned per mounted composer so a card nobody is
-  speaking in cannot let the music back up over the one they are. Ambient
+  speaking in cannot let the music back up over the one they are. Every line a
+  call inherits is disqualified as speech the moment it goes live, so the line a
+  dropped call left mid-sentence — never marked final, and kept on screen on
+  purpose — cannot open the next call already ducked. Ambient
   ownership across conversation cards holds through a keyed card switch: React
   destroys the outgoing card's effects before creating the incoming one's, so
   the last lease going away is settled at the end of the tick rather than on the
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -89,6 +89,20 @@ versions follow [SemVer](https://semver.org/) (0.x — the API may still move).
   so it is visible and usable at 390px instead of painting under the backdrop.
 
 ### Added
+- Background music in the Viewer, and one track across the call boundary
+  (#732). The Audio settings now carry two independent switches sharing one
+  level: music while using the Viewer, and music during a call — the latter the
+  renamed old ambient setting, whose Ukrainian label read as an engineering term
+  («фоновий шар») rather than as music. With both on, the same track simply
+  keeps playing across the call edge in either direction: no teardown, no
+  re-init, no position reset, and speech ducking applies for the duration of the
+  call as before. With only one on, the edge that silences the music parks it —
+  the voice fades out and the position it reached is retained, so the edge that
+  brings it back resumes from there instead of replaying the same opening on
+  every call. Only a device that wants no music at all (the sound master off,
+  both switches off, no asset) tears the track down. Ambient ownership across
+  conversation cards is unchanged: the lease is per mounted composer, so
+  switching the selected card never restarts, duplicates or drops the track.
 - On-canvas pipeline stage reordering (#507). A draft's stage cards carry their
   own move-earlier / move-later controls, so the whole conversation graph is
   composed in place on the canvas — no nested form. Each move is offered only
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -99,10 +99,18 @@ versions follow [SemVer](https://semver.org/) (0.x — the API may still move).
   call as before. With only one on, the edge that silences the music parks it —
   the voice fades out and the position it reached is retained, so the edge that
   brings it back resumes from there instead of replaying the same opening on
-  every call. Only a device that wants no music at all (the sound master off,
-  both switches off, no asset) tears the track down. Ambient ownership across
-  conversation cards is unchanged: the lease is per mounted composer, so
-  switching the selected card never restarts, duplicates or drops the track.
+  every call — and the parked voice stays alive for its whole fade, so a call
+  that ends inside it returns to that very voice instead of stacking a second
+  one over the music still sounding. Only a device that wants no music at all
+  (the sound master off, both switches off, no asset) tears the track down. The
+  music ducks under whoever is talking, read off the transcript the call already
+  produces, with the duck owned per mounted composer so a card nobody is
+  speaking in cannot let the music back up over the one they are. Ambient
+  ownership across conversation cards holds through a keyed card switch: React
+  destroys the outgoing card's effects before creating the incoming one's, so
+  the last lease going away is settled at the end of the tick rather than on the
+  instant — a swipe between conversations never restarts, duplicates or drops
+  the track.
 - On-canvas pipeline stage reordering (#507). A draft's stage cards carry their
   own move-earlier / move-later controls, so the whole conversation graph is
   composed in place on the canvas — no nested form. Each move is offered only
--- a/src/components/SoundToggle.dom.test.tsx
+++ b/src/components/SoundToggle.dom.test.tsx
@@ -27,6 +27,7 @@ const {
   CUE_VOLUME_KEY,
   LOOP_ENABLED_KEY,
   LOOP_VOLUME_KEY,
+  VIEWER_LOOP_ENABLED_KEY,
 } = await import("@/lib/audio/prefs");
 const { AMBIENT_LOOP_ASSET } = await import("@/lib/audio/loopAsset");
 
@@ -161,6 +162,26 @@ describe("levels are device-local and independent", () => {
     /* A reload is a fresh read of the same storage: still on, no re-enabling. */
     expect(readAudioPrefs(storage).loopEnabled).toBe(true);
   });
+
+  test("the two music switches are independent, and share the one level", async () => {
+    await render({ ambientAvailable: true });
+    await openPanel();
+
+    await click(find("ambient-viewer-enabled")!);
+    await setRange(find("ambient-volume") as HTMLInputElement, "0.6");
+
+    /* Music in the Viewer says nothing about calls: the call switch is still the
+       operator's own separate answer. */
+    expect(storage.dump()[VIEWER_LOOP_ENABLED_KEY]).toBe("on");
+    expect(readAudioPrefs(storage).loopEnabled).toBe(false);
+    expect((find("ambient-enabled") as HTMLInputElement).checked).toBe(false);
+
+    await click(find("ambient-enabled")!);
+    const settings = readAudioPrefs(storage);
+    /* Both on, one level — there is no second slider to go looking for. */
+    expect(settings).toMatchObject({ viewerLoopEnabled: true, loopEnabled: true, loopVolume: 0.6 });
+    expect(host.querySelectorAll('[data-testid="ambient-volume"]')).toHaveLength(1);
+  });
 });
 
 describe("the ambient controls exist only when there is something to play", () => {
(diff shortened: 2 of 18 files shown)


Latand:
## Round 1 repairs — all three HIGH findings fixed, test-first

**1. Speech ducking was disconnected from live calls.** Correct finding: nothing in production called `speakingFromLines` or `AmbientLoop.setSpeaking`, and the controller tests hid that by calling `setSpeaking` directly.

`useCodexRealtime` now derives the speaker from the snapshot it already subscribes to and reports it through a new **owner-scoped** `setVoiceSpeaking`, so a card nobody is talking in cannot release the duck the card they *are* talking in is holding. A call that drops mid-sentence releases the duck rather than leaving the music held down under a voice that is gone.

To test that through the *real* hook rather than a copy of it, the hook's store is now a named seam (`RealtimeSurface`) with a test-only factory override — only the WebRTC transport is replaced; the snapshot pushed at it is the one the real client publishes. Four new tests in `useCodexRealtime.ambient.dom.test.tsx` drive non-final → duck and final → release and assert the gain the bed is actually asked for. Mutation-checked: reverting the hook to `useAmbientCallLease(live)` fails exactly three of them and nothing else.

**2. Parking retained the fade's starting position and could overlap voices.** Both halves confirmed and fixed.

- `pause` now retains the position the fade **ends** on (`position() + fade`, wrapped), because the bed is audible for the whole fade. Retaining the start replayed 1.5s on every resume.
- The node's release moved off the audio clock (`stop(when)` cannot be taken back) onto a cancellable timer, and `resume()` takes the park back and hands the **same** voice to the controller. The controller asks a parked voice to resume before it ever opens a new one, so a reversal inside the fade can no longer stack a second voice over one still sounding.

Seven new transport tests drive a deterministic clock through the fade: retained offset, wrapping, node released only at fade end, reversal at +200ms revives the same node with the park cancelled, a completed park answers `resume() === false`, and a teardown mid-park kills the node and leaves no timer behind. Two controller tests cover both branches.

**3. The card-switch test forced the opposite of production effect order.** Correct — and the reviewer's probe reproduced here exactly: rendering a keyed replacement in one root produced **2 starts and 1 pause** before the fix.

The test now renders `<Card key={conversation} />` in one root and swaps the key, which is what `MobileFocusView` does. The fix is in the ownership layer: the **last** lease going away is settled at the end of the tick (microtask — React destroys and creates one commit's effects in a single synchronous pass), so a within-commit handoff is never mistaken for a hangup. A call that genuinely ended still ends, one microtask later. Independent leases for concurrently mounted cards are preserved and separately tested, as is "closing the last card ends the call side". Mutation-checked: replacing the owner-set count with the raw flag fails the switch tests and nothing else.

## Tested

- `bun test src/lib/audio/ src/lib/i18n/ src/components/SoundToggle.dom.test.tsx src/hooks/useCodexRealtime.ambient.dom.test.tsx src/hooks/useAgentChimes.test.ts src/components/TmuxComposer.test.ts src/components/TmuxComposer.deliveryState.dom.test.tsx` — **188 pass, 0 fail**. By path only; no runtime or registry directory swept.
- `bunx tsc --noEmit` clean; `bun run lint` on every changed file clean; privacy gate against the pinned base (`--require-known-values --check-commits`) **PASS**.
- **Pre-existing, not from this branch:** `src/components/TmuxComposer.dom.test.tsx` has 10 failures. Verified identical (17 pass / 10 fail) in a throwaway worktree at `origin/integrate/voice-reload-focus-20260727` itself, so this branch neither causes nor fixes them.

## One decision this round made

The settle for the last lease is a **microtask**, not a timer. It is the earliest point at which the lease set is complete for a commit, so a real hangup still fades within the same tick — no perceptible delay — while a keyed swap is coalesced away entirely.

Latand:
## Round 2 repairs — both findings fixed, each proven by a test that failed first

### HIGH — stale non-final line re-ducks the next call

Confirmed against the client: nothing clears or finalizes `lines`. `start()` resets phase/error/startedAt only, `stop()` sets the phase, and `cleanupTransport()` clears the open-line *table* but leaves the line itself non-final. So a call that drops mid-sentence leaves a line that reads as "the agent is talking" forever.

**Fix (`useCurrentCallSpeaker` in `src/hooks/useCodexRealtime.ts`):** every line a call *inherits* is disqualified as speech at the instant it goes live. The client numbers lines from a counter that only goes up and clears its open-line table on teardown, so an inherited id can never be written into again — which makes "was it already there?" an exact test for "is it from a call that is over?", with no clock and no heuristic. The set is captured during the render that first sees the call (React's documented derive-state-on-change pattern), so the **first committed render** of a live call is already un-ducked; an effect would be one commit late, and that commit is exactly the one that opens the duck.

**Proven by** `src/hooks/useCodexRealtime.ambient.dom.test.tsx` → `ducking under whoever is talking` → *"a line left non-final by a finished call cannot duck the next one"*. It drives the real hook across a real call boundary — live, non-final line (ducks), idle, live again with that same non-final line still present — and asserts not merely that the duck is released but that **no ducked ramp is ever requested** on the new call (`sink.ramps.slice(before).filter(gain < FULL)` is empty) plus `state().ducked === false`. It then pushes a *new* line id and asserts the duck still works, so the guard cannot pass by simply disabling ducking. Against the previous head this test fails at `ducked → true`; the other 8 cases in the file passed unchanged, so the failure localizes to this behavior.

### MEDIUM — resume can revive a park completed on the audio clock

**Fix (`src/lib/audio/webAudioTransport.ts`):** the park now records `parkEndsAt` on the **audio clock**; the wall-clock timer only carries out the release. `resume()` refuses any park whose fade has run out (`context.currentTime >= parkEndsAt`), releases the node on the spot, and returns `false` so the controller opens a fresh voice at the retained position — which is the one place the two clocks can disagree, and the audio clock is the one the fade, the position and the silence are all on.

**Proven by** two cases in `src/lib/audio/webAudioTransport.test.ts` → `parking the bed keeps its place in the track`:
- *"a park the AUDIO clock already finished cannot be revived by a late timer"* — parks at 17s with a 1.5s fade, advances the audio clock to 19s with the release timer still pending, and asserts `resume() === false`, the node stopped, no timer left pending, and that later `rampTo` calls are ignored.
- *"the instant the fade ends counts as finished, not as still parking"* — pins the boundary at exactly 18.5s.

Both fail against the previous head (`resume()` returned `true`); the in-window revival case at +200ms still passes, so the refusal is scoped to expired parks only.

## Verification

- Focused, by path, with `HOME`/`XDG_CONFIG_HOME`/`TMPDIR` pointed at a throwaway dir: `src/hooks/useCodexRealtime.ambient.dom.test.tsx src/lib/audio/webAudioTransport.test.ts src/lib/audio/ambientLoop.test.ts src/lib/audio/app.test.ts src/lib/audio/prefs.test.ts src/lib/audio/cuePlayer.test.ts src/lib/audio/toolCues.test.ts src/lib/audio/toolCues.feed.test.ts src/lib/i18n/i18n.test.ts src/components/SoundToggle.dom.test.tsx src/hooks/useAgentChimes.test.ts src/components/TmuxComposer.test.ts src/components/TmuxComposer.deliveryState.dom.test.tsx` — **189 pass, 0 fail**. No runtime or registry tree swept.
- `bunx tsc --noEmit` clean; `bun run lint` on the four changed files clean; privacy gate against the pinned base (`--require-known-values --check-commits`) **PASS**.
- Still pre-existing and unrelated: the 10 failures in `src/components/TmuxComposer.dom.test.tsx`, identical at `origin/integrate/voice-reload-focus-20260727` itself.

PR #738, opened 2026-07-27
feat(voice): two agents, one gateway, one durable bridge (#691)

Latand:
Implements the ratified architecture in [issue #691](https://github.com/Latand/live-log-viewer-next/issues/691#issuecomment-5090911439). Base `integrate/voice-reload-focus-20260727`. Nothing here deploys.

## The constraint everything protects

The Codex realtime voice root is the **sole user-facing gateway**. A separate persistent manager owns the board — tasks, pipelines, flows, PRs, workers, deploys. Intent flows user → voice → manager → workers; results and questions flow back manager → voice → user. The voice agent never drives workers directly, the manager never speaks to the user, and the two identity records are never collapsed.

## What landed

**The bridge — `src/lib/bridge/`**

Channel state links the root's minted `rootId` to the manager's designation record **by name**. That is the whole reason a manager model swap and a root session rollover cost nothing: no durable bridge field can name a conversation, so neither event invalidates a cursor. A test asserts the serialized channel contains no conversation id (AC22).

Reports are an append-only log in `src/lib/lifecycle/journal.ts`'s mold — monotonic `seq`, idempotent append keyed by a caller-stable string, capacity trim with retired ids, one durable consumer cursor. A trim that outran a slow consumer resumes at the head behind an **explicit gap notice** carrying the trimmed head's seq, rather than skipping history in silence (§7.12).

Confirmations verify **and consume in one file transaction**. A verify that returns yes plus a separate write recording the use are two operations, and two concurrent answers can both pass the first before either reaches the second — that race deploys the same SHA twice. Nonce single-use, strict lowercase 40-hex equality, expiry checked at execution time (§7.7).

**Directives** ride MCP `send_message`, whose receipt is already durable, so exactly-once is inherited rather than reimplemented — provided ids derive from the root turn and are never minted fresh on retry (§7.1). The correlation trailer is read off the **last line only**: scanning the whole body would let a report the gateway is quoting back be mistaken for an authorization.

**Delivery — `src/lib/runtime/bridgeDelivery.ts`**

A batch of reports is expressed *as* a worker delivery — synthetic turn id, one response per report — so the existing ack tombstones, reload resume and hangup paths apply verbatim. A second mechanism would be a second set of the bugs those paths already fixed. Coalescing ≤5 items/batch, ≤1 batch/30s.

**The fence — `src/lib/mcp/toolAllowlist.ts`**

Without this the two-agent split is a convention: the voice root runs against the same MCP server every worker uses, so it could spawn agents and deploy. The gateway now gets `send_message` (manager-addressed only) and `request_attention`; everything else refuses with a code naming where the capability lives. The negative test enumerates the **real** tool list, so a tool added later is denied by default (AC21/AC20).

Identity comes from process ancestry and the registry's recorded hosts, never from anything the caller states. `unidentified` deliberately stays unrestricted — treating "cannot tell" as "is the gateway" would strip the operator's own observed terminal session of every tool it has today, to fence an agent the registry cannot name. Refusals land **before** the receipt claim, so a later grant is not answered by a stale no.

**The floater — `src/components/voice/`**

`VoicePipHost` mounts at Viewer level beside `AttentionHost`, because the card it used to live in unmounts on board navigation while the call keeps running (AC13). It owns the window's lifetime and nothing else — no lines, no phase, no draft, no session — so an OS kill needs no recovery.

Auto-open is **latched per call**. Without the latch, closing the floater mid-call satisfies "a call is up and no window is open" on the next render and the two effects trade turns forever; that was a real loop found by the suite. Re-floating is a separate consumed edge (`floatRequest.ts`), shaped like `focusRequestEdge` for the same reason.

The portal renders presentation only: no `Audio`, no `getUserMedia`, no peer connection, no cue hooks, no dispatcher — asserted by constructor spies (§4 rules 1–2).

One composer store per conversation backs both textareas (U2). Two `useState` drafts are two drafts that look alike until one is sent; the send key derives from the draft's content plus a per-send epoch, so two Send buttons pressed in the same tick compute one string without coordinating.

**Manager record** gains `{engine, model}` naming the incumbent, schema 2 with a forward-reading legacy path — a v1 file reading as "no manager" would abandon a live designation and spawn a second one. `replaceOrchestratorIncumbent` is separate from adoption on purpose: refusing a second conversation while one is live is the single-instance guarantee, and a deliberate swap is a different act. Its prompt is rewritten for an agent with no user-facing channel.

## Verification

- `tsc --noEmit` clean; `eslint` clean on all 34 changed files.
- `privacy-publication-gate --base <merge-base> --check-commits`: **PASS**.
- 193 tests across 21 files, by path, against an isolated `HOME`/`XDG_CONFIG_HOME`/`TMPDIR`/`LLV_STATE_DIR` root. No runtime or registry tree was swept.
- `src/components/TmuxComposer.dom.test.tsx` has 10 failures that reproduce **identically at the merge base** — pre-existing, verified in a throwaway worktree at the base commit, failure sets diffed identical.

## Open with the operator, built behind stated seams

Nothing was invented for U6–U9; each is one edit away:

- **U6** gateway tool grants — `GATEWAY_ALLOWED_TOOLS`, one array.
- **U7** dead-manager respawn — **not automated**. An undeliverable directive currently surfaces as the tool's own failure; no auto-respawn path was added.
- **U8** confirmation TTL — `BRIDGE_CONFIRMATION_TTL_MS`, one constant, 10 min as proposed. A spoken "no" is not yet an immediate unblock; it declines by not consuming.
- **U9** whether the clone carries tool-activity rows verbatim — the floater carries the transcript panel and composer; no work-log rows (AC12 asserts their absence either way).

## Reviewer notes

- **Not wired end to end.** The bridge is a tested seam with no producer: nothing yet appends reports from lifecycle events, and no turn-start drain is mounted. The gateway's send path into the manager is not called from UI. This PR builds the contracts and the fence; the wiring that makes a report travel from a real pipeline into a real call is the next slice.
- **The fence changes the root's live toolset.** A root the registry can name now resolves as `gateway` and loses 24 of 26 tools. That is what §6 specifies, and it is the one change here with immediate effect on how the operator's own root session behaves.
- `AttentionHost` lives in `Viewer.tsx`, not `ProjectDashboard.tsx` as the design's map says; `VoicePipHost` is mounted beside it there.
- The floater's composer is a compact input bound to the shared store, not a literal clone of `ComposerBar`'s full action row — `ComposerBar` mounts effects that must stay single-instance in the opener.


Latand:
## Round 1 fixes — pushed at `2820e5ab`, rebased onto `main` (38b3cbd8)

Both blocking findings were real. Every fix below has a test that fails without it.

### BLOCKING — `toolAllowlist.ts:77` gateway breach — **fixed**

`unidentified` mapped to the full surface, so identity-resolution failure was a path to `spawn_agent` and `deploy_exact_sha`. I had reasoned myself into that as a compatibility default; it made AC21 hold only while the registry happened to be readable, which is exactly the condition a gateway loses in a race.

Now fails closed: only a caller positively identified as *not* the root is unrestricted. `McpCallerIdentity` is `{restricted: gateway|unidentified}` vs `{unrestricted: manager|worker}`. The cost is stated in the code — a worker whose host record cannot be read is refused a tool and recovers when the registry names it again. Refusing a call that should have been allowed is a retry; allowing one that should have been refused ships a commit.

The negative test now enumerates the **complement** of the allowlist over the real `MCP_TOOL_NAMES`, for both restricted kinds, so widening the allowlist by one entry fails it with no fixture to remember.

*Proven by* `toolAllowlist.test.ts` — "a caller whose identity cannot be resolved is denied…", "an unresolved identity gets exactly the gateway surface…", "no restricted identity can reach any tool outside the allowlist". All three failed before the change.

### BLOCKING — `bridgeDelivery.ts:90` bridge not wired — **fixed**

There is now a production path end to end:

```
manager → bridge_report (new MCP tool)
        → durable log
        → GET /api/bridge   (drains from the cursor, moves nothing)
        → useBridgeReportRelay (in the CARD's tree — §4 rule 4 keeps polling out of the portal)
        → the one realtime client
        → POST /api/bridge  (advances the cursor)
```

- `src/lib/bridge/service.ts` owns the ordering so the route, the binding and any later shell take the same path. Channel-open is folded into the read: "the bridge silently does nothing until someone calls `open()`" is the same class of bug as this finding.
- `bridge_report` mints the confirmation nonce server-side — a caller-supplied nonce is a caller-replayable one.
- **Deploy round trip connected**: `deploy_exact_sha` accepts the relayed trailer (`bridgeRef` + `bridgeNonce`) and consumes it atomically before deploying. Replay, expiry, wrong nonce and wrong SHA each refuse; a refusal leaves the confirmation unspent so the operator can still answer correctly.
- **Gap notice delivered**: the trim-vs-slow-cursor case is exercised through the production drain, not just the primitive.

*Proven by* `src/lib/bridge/service.test.ts` (8), `src/app/api/bridge/route.test.ts` (7), `src/hooks/useBridgeReportRelay.dom.test.tsx` (6) — none of which could exist before, since nothing called the primitives in order.

### HIGH — `bridgeDelivery.ts:117` cursor healing — **fixed, plus a worse bug underneath**

The reported defect was suppression without cursor advance. Writing the test found the rest: a delivery id names the **set** of reports it carried, so one more report arriving changes the id, the tombstone stops matching, and the batch is **re-spoken**. The user would hear reports twice.

Suppression is now per-report — the acknowledged seq ceiling is read back out of the tombstones — so a lost cursor write neither hides later reports nor repeats earlier ones. `already-acknowledged` is a distinct plan outcome carrying the seq to heal to, because "nothing to do" and "heal this" want opposite actions.

*Proven by* `recovery.test.ts` — "a lost cursor write cannot hide every later report behind a batch already spoken" (asserts the blocker arrives **and** that reports 1–2 are not re-delivered), "a batch entirely below the ceiling reports the seq to heal to rather than silence".

### HIGH — `Viewer.tsx:706` composer + PiP send path — **fixed**

`useComposer` now takes the shared store: the card writes every edit through, reflects edits from the floating window, and publishes its own `submit`. The floater's Send **is** the card's Send — production passes no `onSend`, so there is no second dispatcher to race. Every other card passes no store and is byte-identical in behavior.

*Proven by* `src/hooks/useComposer.shared.dom.test.tsx` (7) — 6 of 7 failed before the change — and a new `VoicePipHost.dom.test.tsx` case asserting the no-prop path delivers through the registered sender.

### HIGH — `orchestrator/route.ts:43` manager replacement — **fixed**

`POST { replace: true, engine, model }` performs the engine-changing swap. Adoption still refuses a second conversation while one is live — that refusal *is* the single-instance guarantee — so a swap is explicit, and it touches nothing but the record.

*Proven by* `orchestrator/route.test.ts` — "POST replaces a live incumbent…", "a replacement defaults to the standing incumbent identity…". Both failed before.

## Verification

- `tsc --noEmit` clean; `eslint` clean on all changed files (it caught a real ref-during-render in my hydration path, now an effect).
- `privacy-publication-gate --base 38b3cbd8 --check-commits`: **PASS**.
- **163 focused tests across 17 files** green, isolated `HOME`/`XDG_CONFIG_HOME`/`TMPDIR`/`LLV_STATE_DIR`.
- Regression neighbourhood: 105 pass / 1 fail, **identical at base 38b3cbd8** — a cross-file interference failure in `DraftAgentPane.render.test.tsx` that passes when the file runs alone. Verified in a throwaway worktree at base with the same command.
- `server.test.ts`'s pinned tool list updated for `bridge_report`.

## Still open, unchanged

U6–U9 remain the operator's. U7 in particular: a dead manager is still not auto-respawned — an undeliverable directive surfaces as the tool's own durable failure and the same `clientRequestId` retries, but nothing spawns a replacement.

Latand:
## Round 2 fixes — pushed at `c27d0389` (base `main` 38b3cbd8)

All seven addressed. Every fix has a test that fails against the previous behaviour.

### BLOCKING — `directive.ts:36` no production caller — **fixed**

`bridge_directive` is the production path, and two things about it are deliberately not the caller's to choose:

- **Recipient** is resolved from the designation record server-side. A gateway that passes `conversationId: "conversation_some_worker"` still reaches the manager — there is a test for exactly that.
- **Delivery id** is derived from the root turn (`bridge_d_<turn>_<n>`) and used as the delivery key, so a retry with a *different* `clientRequestId` still produces one instruction.

The other half of "wired" is the agent knowing to use it, so the voice persona now carries the gateway's instruction path: relay rather than act, reuse the turn id and index on a retry, never name a recipient, relay a deploy answer with reference/nonce/commit exactly as given.

*Proven by* `src/lib/mcp/bridgeDirective.test.ts` (7) and 4 new `voicePersona.test.ts` cases.

### BLOCKING — `useBridgeReportRelay.ts:82` ack before durable delivery — **fixed**

You were right that this was exactly-once inverted. `reconcileWorkerDeliveries` only enqueues; the host writes later. The client now announces durable acceptance at the one moment it is true — where it already sets `acknowledgedWorkerDeliveries` after the host answers `acknowledged: true` — and the relay parks each batch's cursor until *its own* delivery id comes back.

*Proven by* "the cursor does NOT move on enqueue — only durable delivery may advance it", "the cursor advances once the host confirms the durable write", and "a confirmation for some other delivery never advances this batch's cursor".

### BLOCKING — `bindings.ts:427` deploy could skip the gate — **fixed**

`bridgeRef` is required, in the schema and in the binding, and the refusal happens before the deployment endpoint is touched. A malformed request (abbreviated SHA, missing `confirm`) does not spend the confirmation — a typo must not cost the operator the ability to answer correctly.

*Proven by* `src/lib/mcp/deployConfirmation.test.ts` (6): no-confirmation refused with zero posts, replay/expiry/nonce/SHA each refused with zero posts, and the correct answer still working afterwards.

### HIGH — `toolAllowlist.ts:110` any worker could speak as the manager — **fixed**

`MANAGER_ONLY_TOOLS` covers every non-manager identity in one place — gateway, unidentified, and ordinary workers — enumerated so a manager-only tool added later cannot leak by omission. Refused at the service too, not only in policy.

*Proven by* "an ordinary worker cannot speak to the operator as if it were the manager", "every manager-only tool is denied to every non-manager identity", and the service-level pair.

### HIGH — `useBridgeReportRelay.ts:61` no turn-start drain — **fixed**

`/api/bridge?mode=turn-start` plus `useBridgeTurnStartDrain`, called from the card's send path: the pending batch rides in as part of that turn's own input (U3 ruled out injected developer items). Never on a replay — a retained generation replays its original bytes under its original key, and changing them would defeat the idempotency the retry exists for.

*Proven by* 4 `service.test.ts` cases: bounded batch with a "N more waiting" tail, reading consumes nothing, an acknowledged batch never arrives twice, empty inbox yields nothing.

### HIGH — `composerStore.ts:155` sender loss and attachment ownership — **fixed**

Sender survival already held (last-registration-wins beats React's mount-before-unmount ordering) and now has a test that fails against a naive "clear on release". Attachments were genuinely missing: the floater now acts on the card's real tray via `registerAttachmentOwner`, because the tray owns the `File` objects and the object-URL lifetimes, which cannot be revoked twice.

*Proven by* three new `composerStore.test.ts` cases.

### HIGH — `orchestrator/route.ts:63` split-brain — **fixed**

A swap retires the predecessor **before** seating the successor, and **refuses** the swap if retirement fails (409). Leaving the record on the manager that is demonstrably alive is the safer of the two failures. Replacing the same conversation is a no-op rather than a self-kill.

*Proven by* four `orchestrator/route.test.ts` cases, including "a replacement that cannot retire the predecessor does not seat the successor".

## Two existing tests changed behaviour — updated deliberately, not deleted

- The allowlist surface test **bit correctly** when `bridge_directive` joined the gateway's tools. Updated to the new intended surface of three, with the complement assertion now derived from `GATEWAY_ALLOWED_TOOLS.length` so it keeps biting.
- `bindings.test.ts`'s control-surface check called `deploy_exact_sha` without a confirmation. It now mints and spends one, which is the new contract.

## Verification

- `tsc` clean; `eslint` clean on all changed files; privacy gate **PASS** against `38b3cbd8`.
- **212 focused tests across 20 files** green, isolated `HOME`/`XDG_CONFIG_HOME`/`TMPDIR`/`LLV_STATE_DIR`.
- Regression neighbourhood (12 files, run 2×): 114 pass / 1 fail — the same pre-existing cross-file `DraftAgentPane` failure that reproduces 3× at base in a throwaway worktree. I also caught and fixed a genuine regression this round (`bindings.test.ts`) by running that set repeatedly rather than once.

PR #743, opened 2026-07-27
fix(audio): the bed opens in silence, and cues start at 40%

### LINKED ISSUE #728: fix(audio): ambient bed fades down from full volume at call start
## Problem

The ambient voice-call bed starts audibly at full gain and fades down over the first 2.5 seconds. During speech ducking, the steady level becomes practically inaudible, so the operator hears the track start and then disappear even though the source keeps looping.

## Root cause

The loop gain uses `setValueAtTime(0, currentTime)` and immediately calls the shared ramp helper. That helper cancels scheduled values at the same timestamp and anchors from the stale `AudioParam.value` default of 1.0.

## Fix

Initialize the loop gain through the `AudioParam.value` setter before the first ramp, matching the existing cue path. Add a spec-accurate regression test whose scheduled setter does not synchronously mutate the current value.

## Acceptance criteria

- The first ambient ramp anchors at zero.
- The bed fades up to its configured background level.
- Existing looping, ducking, cue, and degradation tests remain green.
- The exact tested commit is deployed without dropping the realtime progress cursor quota fix currently in production.

Latand:
Closes #728.

## The levels a device is handed before it chooses

`cueVolume` 0.7 → **0.4**, `loopVolume` 0.35 → **0.05**, the levels the operator asked for.

These are fallbacks for keys that are ABSENT from a device's storage, so moving them moves first runs only. A device that already stored a level keeps it: `readAudioPrefs` reads each key independently and never writes on the read path, and the tests pin exactly that — a device holding the levels this build stopped shipping (0.7/0.35) still reads 0.7/0.35 with its storage untouched, and a device that had moved one slider keeps that one and takes the new default for the other.

Worth knowing when judging the result by ear: the ambient slider is not the gain. `LOOP_GAIN_CEILING` (0.35) multiplies it, so 5% on the slider is 0.0175 of full scale — a whisper by design, and unchanged by this PR.

## #728: the bed opened at full volume and slid down

The real cause was one step below the fade. Every ramp began by cancelling what was scheduled, and `cancelScheduledValues(t)` drops the events AT `t` as well — including the one that had just established the opening silence, in the same tick. The ramp then took its anchor by reading `AudioParam.value` back off that emptied timeline, which answers with the param's own 1.0 default. A bed asked to open silent therefore started at full scale and spent 2.5 seconds fading DOWN to its level: the burst at the top of every call.

The fix is that the level is no longer asked of the graph. Each loop voice carries the level this module commanded — a start value, a target and the window between them — interpolated on the audio clock. A fade-in anchors at the silence it opened on, a duck that interrupts it continues from what is actually sounding, and no readback can report the bed louder than it was asked to open. Not a shorter fade: the bed is never audible above its target.

The regression test models the gain param as a real automation timeline (a cancel drops events at or after its timestamp; reading the value back answers from the timeline) and asserts the ENVELOPE rather than the calls — nothing between the node starting and the end of the fade is heard above the target. Against the previous transport all three cases fail, reading 1.0 at the instant the node starts, which is the operator's report exactly.

## Not regressed

The ambient path from the #732 lane is untouched: Viewer music vs call music, parking with a retained position, the resume that returns to a still-fading voice, ducking under speech, ownership across conversation-card switches. `ambientLoop.test.ts`, `app.test.ts` and `useCodexRealtime.ambient.dom.test.tsx` all stay green, including the enabled/enabled crossing that must not restart or reset position.

## Verification

- `bun test src/lib/audio/prefs.test.ts src/lib/audio/webAudioTransport.test.ts src/lib/audio/ambientLoop.test.ts src/lib/audio/app.test.ts src/lib/audio/cuePlayer.test.ts src/lib/audio/toolCues.test.ts src/lib/audio/toolCues.feed.test.ts src/components/SoundToggle.dom.test.tsx src/hooks/useCodexRealtime.ambient.dom.test.tsx` → 132 pass, 0 fail (HOME/XDG_CONFIG_HOME/TMPDIR on a throwaway directory)
- RED first, both halves: the defaults tests fail on the old 0.7/0.35, and the three #728 envelope cases fail against the pre-fix transport
- `tsc --noEmit` clean; `eslint` clean on the four changed files
- `bun scripts/privacy-publication-gate.ts --base 38b3cbd8` → PASS

Not deployed — deployment is deferred until the operator can listen to it.

### DIFF
--- a/src/lib/audio/prefs.test.ts
+++ b/src/lib/audio/prefs.test.ts
@@ -3,6 +3,7 @@ import { afterEach, describe, expect, test } from "bun:test";
 import {
   audioPrefs,
   configureAudioPrefsStorage,
+  CUE_VOLUME_KEY,
   CUES_ENABLED_KEY,
   DEFAULT_AUDIO_PREFS,
   LOOP_ENABLED_KEY,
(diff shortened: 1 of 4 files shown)


Latand:
Reviewer was right, and the finding was precise: the timeline fake answered a readback from the timeline alone, so it dropped `AudioParam`'s current-value slot. Assigning `value` fills both the slot and the timeline, and the slot is exactly what survives the `cancelScheduledValues(now)` that takes the initialization back off — so the base opener was safe and my fake said it burst. The envelope assertions were therefore green for the wrong reason.

Fixed in 8ff6a9ec, in the test only. The production fix is untouched.

The model now carries three things it was missing: the current-value slot, the intrinsic value an empty timeline sounds, and a rendering step that refreshes the slot as the clock advances (`advanceTo`). On that model the three openers separate the way they really do, and each is now measured explicitly:

| opener | first sample | peak over the fade | envelope assertion |
| --- | --- | --- | --- |
| original — `setValueAtTime(gain, currentTime)` only | 1.0 | 1.0 | fails (asserted with `toThrow`) |
| base — `gain.value = …` before `start()`, readback ramp | 0 | ≤ target | passes |
| this head — commanded level, no readback | 0 | ≤ target | passes |

Same discrimination against the REAL transport, not only the in-test replicas — running `webAudioTransport.test.ts` against three source versions:

- base at 38b3cbd8 (`gain.gain.value` + readback ramp): **27 pass, 0 fail** — no more slander of correct code
- base with the original scheduled initializer swapped back in: **24 pass, 3 fail** — the call-start, retained-position and interrupted-duck cases all fail
- this head: **27 pass, 0 fail**

First-sample, peak-envelope, retained-position and interrupted-duck assertions are unchanged; they now run through one shared `expectOpensBelowTarget` so the same bar applies to every opener.

Verification, isolated state (HOME/XDG_CONFIG_HOME/TMPDIR on a throwaway dir), by path:

- `bun test src/lib/audio/prefs.test.ts src/lib/audio/webAudioTransport.test.ts src/lib/audio/ambientLoop.test.ts src/lib/audio/app.test.ts src/lib/audio/cuePlayer.test.ts src/lib/audio/toolCues.test.ts src/lib/audio/toolCues.feed.test.ts src/components/SoundToggle.dom.test.tsx src/hooks/useCodexRealtime.ambient.dom.test.tsx` → 134 pass, 0 fail
- `tsc --noEmit` clean; `eslint` clean on the changed file
- `bun scripts/privacy-publication-gate.ts --base 38b3cbd8` → PASS

Not merged, not deployed.

Issue #182, opened 2026-07-13
Built-in Orchestrator agent: chat surface + all-projects conveyor brain on the viewer API


Latand:
## Concept

A built-in **Orchestrator agent** — the viewer's own resident agent that runs the whole conveyor, with a first-class chat surface in the UI.

## Requirements

1. **Chat surface**: a persistent button on the main screen; tapping it opens a chat with the orchestrator. The user talks ONLY to this agent; it does everything else through the system.
2. **Identity/runtime**: a long-lived agent session, default **Claude Fable, reasoning: low** (configurable). It operates exclusively through the viewer's own API surface (`/api/spawn`, `/api/tmux`, `/api/flows`, `/api/tasks`, `/api/files`, `/api/agent/snapshot`) — it IS the viewer's agent.
3. **Scope — all projects**: it tracks tasks across every project on the board, keeps task cards updated, decides which GitHub issues to create, which PRs to drive, which pipeline to run, and binds work to the existing issue pipeline.
4. **Complexity triage**: for each incoming feature/bug it assesses how big the change is and picks the right agents (model/effort per the role table) and the right pipeline shape.
5. **Skill**: it runs on the `llv-conveyor` skill (`.claude/skills/llv-conveyor/`) — extend the skill where the built-in context differs (it lives inside the viewer, so surface/port discovery is implicit).
6. **On-scheme presence**: the orchestrator appears as a node on the scheme — placed near what the user is currently viewing (it "wedges in" beside the active area) or in a dedicated reserved spot on the map the user can navigate to. Its spawned agents/pipelines link visually to it.
7. **Lifecycle**: survives viewer restarts (resume its conversation), one instance max, clearly marked as the orchestrator on the board.

## Non-goals (v1)
- Multi-user; remote control beyond the existing token auth; replacing manual spawns (they keep working).

Design-first: Fable produces the UX + runtime integration spec (chat panel, scheme placement, session lifecycle), then implementation slices.

Latand:
Contract addition from the draft-pipelines work (#189): the Orchestrator NEVER auto-starts pipelines. 'Агент, побудуй мені пайплайн' → it assesses complexity, composes stages/roles, POSTs a DRAFT (autoStart:false), replies in chat with the draft link/id — the user reviews it on the scheme (editable via #136 builder) and presses Start himself. Auto-start is opt-in per request ('і запусти') only.

Latand:
[...]

PR #744, opened 2026-07-27
feat(monitor): recurring conversation monitor with durable orchestrator resolution (#741)

### LINKED ISSUE #741: feat(monitor): recurring conversation monitor that surfaces pending, stalled, and untracked operator requests
## Problem

A recurring orchestrator watchdog already runs on this machine every 30 minutes, but it does not do what an automatic conversation monitor needs to do, and in its current state it does nothing useful at all.

What it does today:

- reads `state/pipelines.json` directly rather than through the API, counts pipelines in `running`, `needs_decision` or `queued`, and sends one fixed nudge message.
- addresses that message to a **hard-coded orchestrator transcript path**.
- writes no output on success, so its log stays empty and a silent failure is indistinguishable from a silent success.

Two consequences, both observed:

1. **It talks to a dead session.** The hard-coded target is a predecessor orchestrator conversation whose transcript last changed more than a day before the current orchestrator was running. Every fire since then has delivered a nudge into a conversation with no live host, while the actual orchestrator received nothing.
2. **It creates no tracked work.** It never reads conversations, never inspects operator messages, and never creates or updates a board card, an issue, or a pipeline. Nothing it does is auditable afterwards.

Separately, and the reason this matters: operator requests made in conversation can be lost. A request that is spoken or typed, acknowledged, and then overtaken by a more urgent lane leaves no durable trace unless someone materializes it as a card or issue at that moment. There is currently no mechanism that notices.

## Required behavior

A recurring monitor that inspects conversations and surfaces work that is pending, stalled, or untracked.

- **Targets the live orchestrator by identity, never by a hard-coded transcript path.** Resolve the current orchestrator through a durable record so a rollover, restart, or model swap cannot silently orphan the monitor. If no live orchestrator can be resolved, that is a reportable condition, not a silent no-op.
- **Reads operator-authored messages** across this project's conversations over a bounded recent window, and distinguishes genuine operator requests from assistant text, tool output, and its own automated nudges.
- **Correlates each concrete request** against board cards, issues, pipelines, flows and pull requests, and classifies it: completed with terminal evidence, in flight with a known owner, stalled, never materialized, or awaiting explicit operator confirmation.
- **Materializes gaps as tracked board work** rather than only reporting them in chat, so the result survives the conversation. Creation must be idempotent: re-running the monitor over the same window must not produce duplicate cards.
- **Never auto-creates GitHub issues** from inferred intent. An issue is created only on an explicit operator request; the monitor's job is to surface the candidate and say it is unconfirmed.
- **Leaves an audit trail**: what window it scanned, what it found, what it created or updated, and what it deliberately skipped. A run that finds nothing must still be distinguishable from a run that did not happen.

## Constraints

- Read-only with respect to conversations and worktrees; its only writes are board work items.
- Must not spawn implementation agents by itself. Surfacing work is its scope; deciding and delegating stays with the orchestrator.
- Must not interrupt, resume, or kill any live session.
- Must go through the Viewer API rather than reading or writing state files directly.
- Its own output must carry no identities, absolute paths, tokens, or private transcript text — the same publication rules that apply to every other artifact here.

## Acceptance

- A run with a live orchestrator resolves it by record and delivers there; a run with none reports that condition instead of silently succeeding.
- A seeded conversation containing one tracked request and one untracked request yields exactly one new board item, for the untracked one.
- Re-running the monitor over an identical window creates nothing further.
- Its own recurring nudges are never mistaken for operator requests.
- The audit trail distinguishes a clean run from a failed or skipped run.
- Focused tests run by path against an isolated state directory, never the shared live registry.

Latand:
Closes #741.

## What was wrong

The half-hourly watchdog on this machine read `state/pipelines.json` off disk, counted a few pipeline states, and sent one fixed nudge to a **hard-coded transcript path** belonging to a predecessor orchestrator whose session had had no live host for over a day. It created no tracked work and wrote nothing on success, so its empty log was equally consistent with perfect operation and with total failure.

## What this adds

A new `src/lib/monitor/**` slice plus `scripts/conversation-monitor.ts`. No existing source file was modified (only `CHANGELOG.md` and a new design doc), so nothing here touches the PR #737 or #738 lanes.

**Durable orchestrator resolution.** `GET /api/orchestrator` returns the single-instance record (#182): the orchestrator's stable viewer conversation id. The monitor addresses it by that id, never by a path, so a rollover, restart or model swap cannot orphan it. Resolution then includes a read-only host probe — a send into a hostless conversation would *resume* it, and waking sessions is not the monitor's business. Four outcomes: `resolved` (delivers), `missing-record`, `stale-record`, `unavailable` (no host, or the record could not be read). Any of the last three is a **failed run** that raises a board card and exits non-zero, never a silent no-op.

**Operator requests, told apart from everything else.** Only `kind: "message"` / `role: "user"` records inside the window count; `taskTextFromPrompt` (already tested, already knows harness noise) doubles as the concrete-request test; and every delivered report opens with `⟦conversation-monitor⟧`, so the monitor cannot read its own nudge back as an operator request.

**Correlation and classification.** Board cards, pipelines, flows and — when `gh` answers — pull requests and issues project into one `terminal` / `active` / `inert` vocabulary. Word overlap plus explicit `#N` references classify each request as completed, in flight (owner named), stalled, never materialized, or awaiting operator confirmation. A degraded `gh` costs the run its PR/issue correlation and is recorded, never the board work.

**Idempotent materialization.** `untracked` and `awaiting-confirmation` become unplaced board cards through `POST /api/tasks`, each carrying a trailing `monitor-ref: <fingerprint>`. The next run reads the board back through the API and correlates the request to the monitor's own card, so a re-run over the same window creates nothing further (and a paraphrase lands on the existing card instead of a twin). A deterministic `clientRequestId` is the second layer.

**GitHub issues are never created** from inferred intent. A request that asks for one is surfaced as an unconfirmed candidate whose card says so in as many words.

**Audit trail.** One line per run in `state/conversation-monitor/runs.ndjson` (200 retained): `clean` (a run that found nothing still records zero counts), `failed`, or `skipped` (another run held the single-flight lock). The record carries fingerprints, counts and machine ids — no transcript text, no path, no identity. Everything the monitor emits passes through the shared hardened redactor plus home/absolute-path collapsing.

## Verification

- `bun test src/lib/monitor/ scripts/conversation-monitor.test.ts` → 69 pass, 9 files. Each file pins its own `HOME` / `XDG_CONFIG_HOME` / `TMPDIR` / `LLV_STATE_DIR` sandbox and restores the ambient environment afterwards; nothing touches the shared registry.
- `bunx tsc --noEmit` → clean.
- `bunx eslint` on the changed files → clean.
- `bun scripts/privacy-publication-gate.ts --base <merge-base> --check-commits` → PASS.

Acceptance coverage lives in `src/lib/monitor/run.test.ts`: one tracked + one untracked request yields exactly one card; a second run over the same window creates nothing and records `already-tracked`; a missing, stale or hostless orchestrator reports instead of succeeding; an echoed report is never read as a request; clean / failed / skipped stay distinguishable.

## Not done here

Not deployed, and no schedule installed — the cron line is documented in `docs/design/conversation-monitor.md`, and installing it (replacing the old watchdog entry) is an operator action.


Latand:
Review round 1 addressed at `4769ea46`. All six findings fixed; nothing from the round-0 verdict was churned.

**BLOCKING 1 — unproven orchestrator treated as resolved** (`src/lib/monitor/run.ts`). Three holes closed: a record with **no settled transcript path** (a spawn still adopting — nothing to probe, so nothing proven) and a **probe that errored** now both resolve as `unavailable` and fail the run; and a send that reports it **resumed** the host finishes `failed` with `delivered: false`, because a run that had to resurrect its own audience did not deliver in the sense that matters. Proven by *a path-pending record cannot be confirmed, so it is not resolved*, *a host that cannot be probed is not resolved, and nothing is delivered into the doubt*, and *a send that had to resume the host does not finish clean*.

**BLOCKING 2 — state files and non-atomic locking** (`src/lib/monitor/journalStore.ts`, new routes). The journal and lock moved behind `POST/GET /api/monitor/runs` and `POST /api/monitor/lock`; the monitor process opens no file at all, including `--status`. The route sanitizes every record down to the audit fields, so the privacy promise is structural rather than trusted to the caller. The claim is an `O_EXCL` create serialized through the repo's file transaction, with a token only its holder can release. Atomicity is proven by **four racing processes** (`lockClaimChild.ts`, each holding the lock so contention is real) — an in-process test cannot prove this. A run whose audit line cannot be written now reports `audited: false` and exits non-zero.

**HIGH 3 — private transcript text in cards** (`src/lib/monitor/cards.ts`, `redact.ts`). The `> …` quote block is gone; a card carries a short derived label plus monitor-authored context and a line saying so. Redaction now covers email addresses, absolute paths **down to two segments**, and percent- and dash-encoded path forms. Guarded by a new `redact.test.ts` targeting exactly those shapes plus *never quotes the operator's own words back onto the board*. Two consequences worth naming: route-shaped text like `/api/tasks` now collapses too (accepted — nothing emitted needs to name a route, and a rule with holes is worth less than one that over-redacts), and the redaction patterns and their test fixtures are assembled at runtime, since a literal path or address in a committed file is itself a gate finding.

**HIGH 4 — global correlation** (`src/lib/monitor/classify.ts`, `evidence.ts`). Every evidence item now carries its project, and correlation is scoped to the request's own board. Repository-wide pull requests and issues carry no project and are admitted only through a reference the operator named. Proven by *another project's work can never answer for this project's request* and *a foreign project's card leaves the request untracked*.

**HIGH 5 — passing issue mention read as completion** (`src/lib/monitor/classify.ts`). A reference match now reports whether the wording corroborates it; an uncorroborated `contextual-reference` goes to `awaiting-confirmation`. Proven by *an issue named in passing never retires the request that mentioned it* (a closed `#741` beside "ship the exporter before #741 lands" no longer reads as `completed`) and its corroborated counterpart.

**MEDIUM 6 — staleness from creation time** (`src/lib/monitor/evidence.ts`, `viewerApi.ts`). Last movement now comes from stage-attempt instants plus pause/resume for pipelines and round instants for flows. Both acceptance cases are pinned: *a long-running container's last movement is its newest stage activity, not its birthday*, and *an old container with no recent movement keeps its old instant*. No movement evidence reports `null`, which the classifier treats conservatively — still in flight, never stalled on a missing field.

**Verification.** `bun test src/lib/monitor/ src/app/api/monitor/ scripts/conversation-monitor.test.ts src/lib/tasks/inboxScanner.test.ts` → 117 pass, 13 files, each pinning its own `HOME`/`XDG_CONFIG_HOME`/`TMPDIR`/`LLV_STATE_DIR` and restoring the environment. `tsc --noEmit` clean, eslint clean on changed files, privacy gate PASS with `--check-commits`. Scope fences held: no file under `src/lib/pipelines/**`, `src/lib/flows/**`, `src/lib/mcp/toolAllowlist.ts`, `src/lib/runtime/bridgeDelivery.ts` or `src/app/api/orchestrator/route.ts` was touched — pipeline and flow lifecycle fields are read structurally from the API payloads instead.

Not merged, not deployed.

Latand:
**Rejected by product direction — not merging, not deploying.**

The operator has rejected the deterministic script/rule-based classifier approach. This is a direction decision, not a verdict on the implementation or on the round-2 review findings.

The branch, commits and review artifacts are **preserved as evidence** and deliberately not deleted. This PR stays open and marked so the reasoning remains auditable.

**What replaces it:** the monitor becomes a real model-backed agent in one dedicated, durable, project-owned session that a scheduler wakes on a cadence, resolved by durable conversation/project identity rather than any hard-coded transcript path. Each run analyses only the delta since a durable watermark and semantically reads operator-authored messages, correlating them against cards, issues, pipelines, PRs and terminal evidence. It may propose and label tasks through the Viewer API with idempotent provenance, and never auto-creates GitHub issues.

Deterministic code keeps a role — collecting and deduping a bounded delta, and enforcing safety — but it may never make the semantic classification itself.

Carried forward from this work as prior art: the watermark and idempotency approach, the redaction shapes (including the nested percent-encoding gap found in review), own-output exclusion, and project scoping.

This head has also been **removed from the operator's isolated inspection stage**. A replacement architecture lane is being created now.

Issue #38, opened 2026-07-09
Sync the board across devices: same window set, arrangement & order everywhere (server-side per-project state, not per-browser localStorage)


Latand:
## Summary

I want my board to look the **same on every device**. When I open the same project page on my computer and on my phone, I should see the **same set of windows, in the same arrangement and order**. Today the phone shows a different screen — a different set of windows, a different layout — and nothing I arrange on the desktop carries over.

The fix: move the board's **arrangement state** off each browser's `localStorage` and into a **server-side, per-project store**, so all devices read and write the same state. This is the same model the board already uses for task-card positions (`state/tasks.json`) — extend it to the rest of the board layout.

## Root cause — the board's arrangement lives in per-browser storage

The node **positions/order** are already deterministic and identical everywhere: `buildSchemeLayout()` computes world coordinates fresh each render from server data, ordered freshest-first by `mtime` (`src/components/scheme/layout.ts`). So the geometry isn't the problem.

What diverges is the **per-project view state**, held in each browser and never shared:

| State | Storage key | Where | What it controls |
|---|---|---|---|
| **Column prefs** | `llvCols:<project>` (localStorage) | `ProjectDashboard.tsx:78`, `ColumnPrefs` :69 | `manual` (windows added to the board), `hidden` (windows removed), `expanded` (connected conversations expanded as nodes) — **the core "which windows and how they're arranged"** |
| **Empty/quiet view** | `llvEmptyView:<project>` (localStorage) | `ProjectDashboard.tsx:84` | scheme board vs flat list for a quiet project |
| **Task panel** | `llvTaskPanel` (localStorage) | `ProjectDashboard.tsx:176` | right task panel open/closed |

Because these are browser-local, the desktop and phone each build a different board from the same server data.

On top of that, **phone doesn't render the board at all** — it uses a separate single-focus view (`src/components/mobile/MobileFocusView.tsx`), which is the other half of "the phone screen is different."

**Camera pan/zoom (`llvCam:<project>`, `src/components/scheme/useSchemeCamera.ts`) stays per-device by design** — phone and desktop have different screen sizes, so a shared viewport would frame badly on one. Only the *contents and arrangement* sync, not the viewport.

### B1. Server-side per-project board store (shared across devices)
Move the arrangement state to a server-side JSON store, keyed per project, mirroring the tasks precedent:
[...]

Issue #172, opened 2026-07-13
Board first paint flashes the full node set, then prunes (closed/collapsed/cap state applies after hydration)


Latand:
Reloading a project page (e.g. #p=-agents-tools-live-log-viewer-next) first renders MANY agent cards/nodes, then most disappear after a beat. Cause: the initial render paints the raw scan snapshot; persisted closes, worker-stack collapse, and scheme caps apply only after client hydration/refresh. User reads this as data loss every time (reported twice on 2026-07-13).

Fix: apply persisted board state (closed cards, collapsed stacks, scheme caps) BEFORE first paint — gate initial render on the persisted-state load (or SSR it), so the first frame already shows the pruned set. A short skeleton is acceptable; a flash of 100+ nodes is not.

Related: #169/#170 (fair cap), #171 (list/search), #134/#153 (worker stacks).

Latand:
Shipped in PR #217: first paint gated on the settled snapshot (SWR cache from #176) with skeletons before first full scan — no more paint-then-cull flash.

PR #745, opened 2026-07-27
Board sync convergence across desktop, phone and agents (revives #38)

Latand:
## The divergence, reproduced

Two views of one project could disagree on the window set, and a view that had fallen behind could overwrite current state with its own older picture. Three independent paths, each reproduced as a failing test before any fix:

**1. Same tab, two views.** `useBoardState(project)` is mounted twice — the shell binds a project board for its rail, the dashboard binds it for the scheme. Each binding built a *private* store: its own confirmed board, its own outbox, its own 10-second poll clock. Closing a window in the dashboard left the rail showing it until that rail's next poll, and only a reload settled a real disagreement.

**2. A view that cannot write stops reading.** `poll()` bailed out whenever the outbox was non-empty. A device that lost its uplink mid-drag, or a laptop closed with queued intent, therefore stopped learning that the board had moved at all — it rendered the revision it loaded with until a reload, while the other device kept working.

**3. The stale writer wins.** Exact sequence:

1. Desktop and phone both load the project at revision 1 (`viewMode: null`).
2. The desktop loses its uplink and the operator switches it to list view. The PATCH cannot land; the intent sits in the outbox.
3. On the phone, minutes later, the operator picks the scheme view — against the board as it actually stands. Revision 2.
4. The desktop's uplink heals. Its PATCH is refused with 409, it adopts revision 2 — and then **replays its queued `set-presentation: list` on top**. Revision 3.
5. The phone polls and flips to list. The operator's most recent, best-informed choice was undone by a device that had been asleep.

The 409 fence was doing its job; the client simply put the superseded intent straight back.

## Conflict rule

**Per-key last-writer-wins, fenced by observation.**

The writer is the device whose PATCH the server serializes last at a matching `baseRevision`; the server rejects any writer whose base is behind. A stale view cannot be that writer, because when it adopts a board it did not author it **diffs the board it held against the board it adopted** to learn exactly which keys another writer changed, and drops its own queued mutations naming those keys. Queued mutations naming untouched keys replay on top and still land.

So the last writer on any key is always one that acted with knowledge of that key's current value. A view that has fallen behind loses exactly the keys it is behind on, and nothing more.

A *key* is one independently-writable unit: a transcript path's placement (including whether the manual slot is a genuine pin), a favourite or tray identity, or one presentation field. Paths are compared through the adopted board's aliases, so a succession remap that carried a placement to a new transcript path reads as no change — bookkeeping must not look like a competing operator decision.

The fence runs in three places, not one:

- **409** — the obvious case.
- **poll adoption** — now that a read is no longer gated on a pending outbox.
- **an accepted write** — `writeReduced` returns `ok` for a batch that reduces to nothing *from any base revision*. A queued close of an already-closed window is accepted at a stale base and hands the view a board another writer moved, with no 409 to mark it as behind. Diffing the board the write should have produced against the one that came back names that writer's keys. This one I found in self-review after the first fix landed, and it has its own test.

## Also fixed

- `sync` now actually reports `"stale"`. The value was declared in `BoardSync` and in the presence wire DTO and validator, and was **never produced** — `syncFor()` could only return `current`/`pending`/`unavailable`. A lagging device published `current` to agents reading the snapshot API. It now reports `stale` from the moment it adopts a foreign board while holding intent of its own until that intent is acknowledged.
- A project still at revision 0 whose first load failed stayed `unavailable` forever: recovery only cleared the flag when the revision *changed*, and an untouched project's revision never does. The dashboard held its skeleton and its convergence effects, which bail on an unavailable board, never ran again. A successful read now clears it — unless a held outbox says writes are still being refused (schema skew), where the signal must stand.
- Poll adoption is guarded by a write epoch, so a GET that started before a PATCH can never install its now-superseded board over the write's authoritative response.

## Durability

The convergence tests run both writers against the real `mutateBoard`/`patchBoard`/`boardFor` over an isolated temp file, so they assert shipped server semantics rather than a hand-written fake. `boardFor` re-reads the file on every call and a fresh store is a restart; the first test asserts a restarted store reads the converged arrangement back off disk.

## Tests

Every test below was written first and **failed against current behaviour**; the failures are in the commit's working history.

`src/hooks/useBoardState.convergence.test.ts`
- a view holding unflushed intent still converges on the other device's board *(+ restart durability)*
- a view whose first load failed recovers on the next poll, not on a reload
- a stale view's presentation intent loses to the device that acted on current state — **this is the stale-writer proof**
- an accepted no-op cannot smuggle a newer board past the fence
- a stale view loses only the keys another device wrote, and keeps the rest
- a stale root reconciliation cannot retire the window another device just opened

`src/hooks/useBoardState.sharedStore.dom.test.tsx`
- two bindings on one project render the same board without a reload

## Verification

- `bun test src/hooks/useBoardState.convergence.test.ts src/hooks/useBoardState.sharedStore.dom.test.tsx src/hooks/useBoardState.test.ts src/lib/board/ src/app/api/board/ src/lib/view/ src/app/api/view/ src/components/scheme/spatialNav.test.ts` — 204 pass, 0 fail
- Plus the dashboard/viewer mount points, run in a clean checkout of the merge base with these changes applied: 411 pass, 0 fail across 36 files (`src/hooks/`, `src/lib/board/`, `src/lib/view/`, `src/app/api/board/`, `src/app/api/view/`, the four `ProjectDashboard` files, `Viewer.catalogFailure.dom`, `Viewer.test`, `MobileBottomShelf.dom`, `spatialNav`)
- All runs use isolated `HOME`, `XDG_CONFIG_HOME`, `TMPDIR` and `LLV_STATE_DIR`; no runtime or registry tree is swept
- `tsc --noEmit` clean; `eslint` clean on the three changed files; privacy publication gate PASS against the merge base

One caveat worth flagging: in *this* worktree, running `runtimeBus.test.ts` in the same process as `Viewer.catalogFailure.dom.test.tsx` fails with a `mock.module` export leak. It reproduces with the working tree reverted byte-identical to the merge base, and does not reproduce in a fresh checkout of the same commit — pre-existing and environment-sensitive, unrelated to this change.

## Not touched

No file owned by the gateway (#738), conversation-monitor (#744), handoff (#737) or audio (#743) lanes. `Viewer.tsx` is untouched — the same-tab fix lives entirely inside `useBoardState.ts`, so no sequencing behind that lane is needed.

Not deployed; deferred until the operator can test.

Revives #38.

### DIFF
--- a/src/app/api/board/route.ts
+++ b/src/app/api/board/route.ts
@@ -124,7 +124,9 @@ export async function PATCH(request: NextRequest): Promise<NextResponse> {
       ? mutateBoard(payload.project, payload.baseRevision, mutationsWithConversationAliases(payload.mutations))
       : patchBoard(payload.project, payload.baseRevision, payload.patch!);
     if (!result.ok) return NextResponse.json({ error: "BOARD_REVISION_CONFLICT", board: result.board }, { status: 409, headers });
-    return NextResponse.json({ ok: true, board: result.board }, { headers });
+    /* `applied` lets the caller tell its own committed write from a no-op that
+       was accepted against a board some other writer produced (#38). */
+    return NextResponse.json({ ok: true, applied: result.applied, board: result.board }, { headers });
   } catch (error) {
     if (error instanceof ViewValidationError) return NextResponse.json({ error: error.code, message: error.message }, { status: error.status, headers });
     if (error instanceof BoardStoreError) return NextResponse.json({ error: "INTERNAL_ERROR", message: "board state unavailable" }, { status: 500, headers });
--- a/src/hooks/useBoardState.convergence.test.ts
+++ b/src/hooks/useBoardState.convergence.test.ts
@@ -527,6 +527,137 @@ test("a stale writer still loses after its key was evicted by retired-key compac
   store.dispose();
 });
 
+/*
+ * The adoption paths. A board arrives authoritatively in five places: a poll, a
+ * 409, an accepted write, the initial load, and the completion of the legacy
+ * seed. The first three are fenced; these cover the last two. A gate that only
+ * some entry points use is not a gate.
+ */
+
+test("intent queued on a cached remount loses to a foreign A→B→A the load reveals", async () => {
+  const server = durableBoard();
+  server.otherDevice("proj", [
+    { kind: "restore", path: "/a", placement: "manual" },
+    { kind: "restore", path: "/x", placement: "manual" },
+  ]);
+  /* One mount settles the project, so the session cache holds revision 1. */
+  const first = createBoardStore({ project: "proj", fetcher: server.fetcher, storage: null, scheduler: idleScheduler().scheduler });
+  await settle();
+  expect(first.getSnapshot().revision).toBe(1);
+  first.dispose();
+
+  /* While nothing of ours is mounted, the phone closes /x and reopens it. The
+     board ends exactly where the cache remembers it. */
+  server.otherDevice("proj", [{ kind: "close", path: "/x" }]);
+  server.otherDevice("proj", [{ kind: "restore", path: "/x", placement: "manual" }]);
+  expect(server.board("proj").revision).toBe(3);
+
+  const device = flakyDevice(server);
+  device.state.patchDown = true;
+  const sched = idleScheduler();
+  const store = createBoardStore({ project: "proj", fetcher: device.fetcher, storage: null, scheduler: sched.scheduler });
+  /* A remount paints the cached arrangement on its first frame (#172), so the
+     dashboard is live and the operator can act before the load has landed. */
+  expect(store.getSnapshot()).toMatchObject({ loaded: true, revision: 1 });
+  store.mutate([{ kind: "close", path: "/x" }]);
+
+  await settle();
+  device.state.patchDown = false;
+  sched.runTimeouts();
+  await settle();
+
+  /* The load is an authoritative adoption like any other. Assigning it straight
+     to `confirmed` leaves intent formed against the cached revision unfenced,
+     and it replays over a reopen this device never saw. */
+  expect(server.board("proj").prefs.hidden).toEqual([]);
+  expect(server.board("proj").prefs.manual).toEqual(["/a", "/x"]);
+  expect(server.board("proj").revision).toBe(3);
+  expect(store.getSnapshot().prefs.hidden).toEqual([]);
+  store.dispose();
+});
+
+test("intent queued during a seed that turns out to be a no-op loses to a foreign A→B→A", async () => {
+  const server = durableBoard();
+  const storage = { getItem: (key: string) => (key === "llvCols:proj" ? JSON.stringify({ manual: ["/x"], hidden: [], expanded: [] }) : null) };
+  let releaseSeed = () => {};
+  const seedGate = new Promise<void>((resolve) => (releaseSeed = resolve));
+  let patches = 0;
+  const fetcher = async (input: string, init?: RequestInit) => {
+    if (init && (init.method ?? "GET") !== "GET") {
+      patches += 1;
+      if (patches === 1) await seedGate;
+    }
+    return server.fetcher(input, init);
+  };
+  const sched = idleScheduler();
+  const store = createBoardStore({ project: "proj", fetcher, storage, scheduler: sched.scheduler });
+  await settle(); // the revision-0 GET landed; the seed PATCH is inflight and gated
+
+  /* Another device seeds the same arrangement from its own localStorage, then
+     the operator closes /x there and reopens it. The board now reads exactly
+     what this device's seed was going to write — so the seed will reduce to
+     nothing and be accepted from its stale revision-0 base. */
+  server.otherDevice("proj", [{ kind: "restore", path: "/x", placement: "manual" }]);
+  server.otherDevice("proj", [{ kind: "close", path: "/x" }]);
+  server.otherDevice("proj", [{ kind: "restore", path: "/x", placement: "manual" }]);
+  expect(server.board("proj").revision).toBe(3);
+
+  /* Meanwhile the operator closes /x on THIS device, against the optimistic
+     seed — intent formed with no knowledge of any of the above. */
+  store.mutate([{ kind: "close", path: "/x" }]);
+
+  releaseSeed();
+  await settle();
+  sched.runTimeouts();
+  await settle();
+
+  /* Seed completion adopts a board this device did not author, so it has to
+     fence like every other adoption. The reopen must survive. */
+  expect(server.board("proj").prefs.hidden).toEqual([]);
+  expect(server.board("proj").prefs.manual).toEqual(["/x"]);
+  expect(server.board("proj").revision).toBe(3);
+  expect(store.getSnapshot().prefs.hidden).toEqual([]);
+  store.dispose();
+});
+
+test("intent formed after an authored seed survives that seed landing", async () => {
+  const server = durableBoard();
+  const storage = { getItem: (key: string) => (key === "llvCols:proj" ? JSON.stringify({ manual: ["/seed"], hidden: [], expanded: [] }) : null) };
+  let releaseSeed = () => {};
+  const seedGate = new Promise<void>((resolve) => (releaseSeed = resolve));
+  let patches = 0;
+  const fetcher = async (input: string, init?: RequestInit) => {
+    if (init && (init.method ?? "GET") !== "GET") {
+      patches += 1;
+      if (patches === 1) await seedGate;
+    }
+    return server.fetcher(input, init);
+  };
+  const sched = idleScheduler();
+  const store = createBoardStore({ project: "proj", fetcher, storage, scheduler: sched.scheduler });
+  await settle();
+
+  /* The operator closes the very node the seed is placing, while the seed is
+     still inflight. Nobody else is writing — this is our own seed and our own
+     later decision about a key the seed touches. */
+  store.mutate([{ kind: "close", path: "/seed" }]);
+  expect(store.getSnapshot().prefs.hidden).toEqual(["/seed"]);
+
+  releaseSeed();
+  await settle();
+  sched.runTimeouts();
+  await settle();
+
+  /* Fencing must not treat this device's own authored seed as a competing
+     writer: the close was formed after it and has to land. Over-fencing here
+     silently eats the operator's action. */
+  expect(server.board("proj").prefs.hidden).toEqual(["/seed"]);
+  expect(server.board("proj").prefs.manual).toEqual([]);
+  expect(store.getSnapshot().prefs.hidden).toEqual(["/seed"]);
+  expect(store.getSnapshot().sync).toBe("current");
+  store.dispose();
+});
+
 test("a stale view loses only the keys another device wrote, and keeps the rest", async () => {
   const server = durableBoard();
   server.otherDevice("proj", [
--- a/src/hooks/useBoardState.convergence.test.ts
+++ b/src/hooks/useBoardState.convergence.test.ts
@@ -41,7 +41,10 @@ function durableBoard() {
       ? mutateBoard(body.project, body.baseRevision, body.mutations, file)
       : patchBoard(body.project, body.baseRevision, body.patch!, file);
     if (!result.ok) return { ok: false, status: 409, json: async () => ({ error: "BOARD_REVISION_CONFLICT", board: result.board }) };
-    return { ok: true, status: 200, json: async () => ({ ok: true, board: result.board }) };
+    /* Mirrors the route: `applied` reports whether the write committed or
+       reduced to a no-op, which is what separates our own work from another
+       writer's when both produce the same board. */
+    return { ok: true, status: 200, json: async () => ({ ok: true, applied: result.applied, board: result.board }) };
   };
   return {
     file,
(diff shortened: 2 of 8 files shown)


Latand:
## Review round 1 addressed — 638244b3

### BLOCKING: ABA — fixed by per-key causal revisions

The reviewer is right, and the finding is structural rather than a missed case. Comparing the board a view held against the board it adopted only ever sees endpoints. A key driven away from a value and back lands on a snapshot identical to the one the stale view remembers, so the comparison reports "nobody touched this" and the superseded intent replays and wins. Both cycles are ordinary operator behaviour — toggling a view mode, closing and reopening a window — so this was the common case, not a corner.

Snapshot diffing is gone. The board now carries **per-key causal revisions**: `key → the board revision at which that key last changed`, server-authored, durable, monotonic. Every queued mutation records what it observed for each key it writes; adopting a board this view did not author drops any intent whose key has advanced past that observation.

New `src/lib/board/keys.ts` holds the key vocabulary so server and client cannot drift — if they name keys differently the fence silently stops matching, which is the worst possible failure mode for this.

Three things fell out of doing it properly:

**Every write path stamps.** All server writes now go through one `committed()` helper. `transferBoardPathPlacements` and `migrateBoardProjects` bumped revisions by hand and would otherwise have advanced the board without advancing its keys — a silent hole in the fence.

**Retired keys must be kept for a while.** My first cut pruned `keyRevisions` to keys the board still carries. `src/lib/board/keys.test.ts` caught it immediately: un-favouriting removes the id from `favorites`, so the key was pruned at the very moment it was written, and a stale writer that favourited beforehand would still win. "Off" is a real value that leaves no trace. Retired keys are now kept most-recently-written first up to `MAX_RETIRED_KEY_REVISIONS = 512`, which keeps the map proportional to the board instead of growing with every transcript ever opened. The residual limit is stated plainly: a view would have to be behind by more than 512 writes *on keys that have since left the board entirely* before its intent stops being fenced.

**PATCH now reports `applied`.** Working through what a reviewer would attack next, I found a case the client cannot resolve on its own. A no-op batch is accepted from *any* base revision, so an accepted response can carry a board a different writer produced — and when both writers made the same change, the returned board is byte-identical to the one our own write would have produced. No amount of content comparison separates those. The server knows which happened, so it now says so, and the client uses that instead of guessing. There is a test for it, and it fails without the field.

### LOW: contract text — fixed

The `createBoardStore` docstring said a conflict "replays the whole outbox on top", which stopped being true. It now states what actually happens: intent whose key advanced past what it observed is dropped as superseded, intent on keys nobody else wrote replays, so a mutation survives an interleaved write to a *different* key and loses to one on the *same* key.

### The contract the settings lane can build on

Stating this deliberately, since the reviewer flagged that this shape constrains a separate architecture lane:

> **Per-project durable state is a server-owned document with a monotonic `revision`, mutated only through a revision-fenced API (`baseRevision` must match or the write is refused), and carrying `keyRevisions: Record<key, revision>` stamped on every key each write changes. Clients hold one refcounted store per project, queue semantic mutations tagged with the key revisions they observed, and drop queued intent whose key has advanced past that observation. Conflicts therefore resolve per key by causality, not per document and not by wall clock.**

Concretely, a per-project settings surface adopting this gets: two devices editing different settings never contend; two editing the same setting resolve to whichever writer acted knowing that setting's current value; and a setting toggled off and back on is still detected as written. What it must supply is a key name per independently-writable setting in `src/lib/board/keys.ts` (or the equivalent module for its own document) — that is the only extension point, and skipping it silently disables the fence for those settings.

Two properties worth knowing before depending on it: the metadata is **durable, not derived** (it cannot be rebuilt from the document after a restart, which is why it is persisted and covered by a restart test), and the server must report **whether a write applied**, because "accepted" and "authored" are genuinely different and a client cannot tell them apart.

### Tests

New, each failing before its fix:

- `an A→B→A cycle on a presentation key still supersedes a stale writer`
- `an A→B→A cycle on a path key still supersedes a stale writer`
- `an A→B→A cycle still supersedes a stale writer when the key leaves the board` — the retired-key retention case
- `an accepted no-op that another writer made identical is not mistaken for our own write` — the `applied` case
- `src/lib/board/keys.test.ts` — five cases on the durable contract: revisions advance on every write so A→B→A stays distinguishable, a write moves only the keys it changed, revisions are durable across a restart and default empty on legacy boards, a succession remap does not burn the old path's revision, and `mutationKeys` names every key a mutation contends on

### Verification

- `bun test src/hooks/useBoardState.convergence.test.ts src/hooks/useBoardState.test.ts src/hooks/useBoardState.sharedStore.dom.test.tsx src/lib/board/ src/app/api/board/ src/lib/view/ src/app/api/view/ src/components/scheme/spatialNav.test.ts src/components/ProjectDashboard.*` — 227 pass, 0 fail
- Clean checkout of the new head, including the dashboard and viewer mount points: 420 pass, 0 fail across 37 files
- `src/lib/accounts/` and `src/app/api/conversations/` run to check the migration/transfer write paths: 2 failures, both pre-existing (they fail identically with these changes stashed out)
- Isolated `HOME`, `XDG_CONFIG_HOME`, `TMPDIR`, `LLV_STATE_DIR` throughout; no runtime or registry tree swept
- `tsc --noEmit` clean, `eslint` clean on changed files, privacy publication gate PASS

Not merged, not deployed.

Latand:
## Review round 2 addressed — 1e38ab07

### BLOCKING: alias-equivalent path placements had separate clocks

Correct, and it is the same class of bug one level down: I fixed ABA for a single key and left it wide open across an alias boundary.

A conversation that resumes mints a new transcript path and the board aliases old onto new. `path:/old` and `path:/new` carried independent causal revisions, so a stale pre-remap `close /old` looked completely unopposed against an informed post-remap write naming `/new`. It replayed, `applyBoardMutations` resolved it through the alias, and it hid the node the other device had just placed.

**A key now names a logical thing, not a spelling of it.** Every read, write and comparison of a key goes through `canonicalKey` against the board's alias graph:

- `boardKeysChanged` compares canonical *placement classes* on both sides, read through the resulting board's aliases. A pure succession remap is therefore not a change at all — the class held the same role before and after, only its spelling moved — so bookkeeping still cannot burn a causal revision or read as a competing operator decision.
- Carried-forward clocks **merge by maximum**. A rename can only ever move a class's clock forward, so no observation is un-superseded by a remap.
- Alias sources stop appearing in `liveBoardKeys`; they resolve onto their target's key rather than holding one of their own.
- Client-side, recorded observations are re-canonicalized against the adopted board, so intent formed under the old name reads the successor's clock.

One extra thing fell out in self-review: a board written by the round-1 build could hold *both* an alias and a separate pre-alias key, and canonicalizing only on lookup would leave that clock unreachable — reintroducing exactly this bug for one file generation. `keyRevisions` is now canonicalized on **read**, merging by maximum, with a test for that shape.

### HIGH: eviction erased causal history

Also correct, and worse than it looks because it degrades with age — the longer a board lives, the more likely a stale writer wins.

Compaction now leaves a **floor**. Whatever eviction drops raises `keyRevisionFloor`, and every key absent from `keyRevisions` reads that floor instead of zero. Live keys keep their exact clock; only retired ones are subject to the bound, and a retained retired key at or below the floor is dropped too, since it says nothing the floor does not.

The floor is monotonic and deliberately errs **high**: it can fence intent whose own key was never actually written. That costs a dropped mutation. The other direction costs a lost write, and only one of those is a correctness bug — so the bias is chosen, not incidental.

### LOW: contract text

Now stated in full in the `createBoardStore` docstring, with the reviewer's sentence verbatim at the top and every clause annotated with the defect that made it necessary, so the next reader inherits the reasoning instead of reconstructing it.

### The contract, now literally true

> **One project-scoped shared store over a revision-fenced durable API, with a monotonic server-authored causal revision for each logical key that survives aliases, ABA, restart, and safe tombstone GC.**

Clause by clause, with the test that holds it up:

| clause | held by |
|---|---|
| project-scoped shared store | `two bindings on one project render the same board without a reload` |
| revision-fenced durable API | `increments revisions atomically and rejects stale concurrent writers` |
| monotonic causal revision per key | `a key's causal revision advances on every write, so A→B→A stays distinguishable` |
| server-authored | `an accepted no-op that another writer made identical is not mistaken for our own write` |
| survives aliases | `a stale pre-remap close loses to an informed post-remap write on the same conversation` |
| survives ABA | the three A→B→A tests (presentation key, path key, key that leaves the board) |
| survives restart | `causal revisions are durable and survive a restart` |
| safe tombstone GC | `a stale writer still loses after its key was evicted by retired-key compaction` |

For a per-project settings surface, the extension point is unchanged and is still the only one: a key name per independently-writable setting in `src/lib/board/keys.ts`. Two additions worth knowing since round 1 — a key must name a **logical identity**, so anything that can be renamed needs an alias class rather than two keys; and compaction is **lossy but safe**, so a client behind by more than the retained history is fenced conservatively rather than trusted.

### Tests

New this round, both failing before their fix:

- `a stale pre-remap close loses to an informed post-remap write on the same conversation` — the remap race
- `a stale writer still loses after its key was evicted by retired-key compaction` — over-cap regression, 600 retired keys forcing eviction of the stale writer's own key

Plus, in `src/lib/board/keys.test.ts`: `a remap unifies the two names onto one clock without burning it`, `a merged alias class keeps the newer of the two clocks`, `a board written before keys were unified normalizes its alias clocks on read`, and `compaction raises a floor instead of erasing causal history`.

One round-1 test changed rather than being added to: `a remap carries a placement without burning the old path's causal revision` asserted that `path:/old` keeps its own clock — which was the bug. It is now `a remap unifies the two names onto one clock without burning it`, asserting the stronger property.

### Verification

- Focused run by path: 232 pass, 0 fail
- Clean checkout of the new head, including dashboard and viewer mount points: 425 pass, 0 fail across 37 files
- `src/lib/accounts/` and `src/app/api/conversations/`: 2 failures, both pre-existing and unchanged
- Isolated `HOME`, `XDG_CONFIG_HOME`, `TMPDIR`, `LLV_STATE_DIR` throughout; no runtime or registry tree swept
- `tsc --noEmit` clean, `eslint` clean on changed files, privacy publication gate PASS

Not merged, not deployed.

Latand:
## Review round 3 addressed — 4c1fb388

### BLOCKING: load and seed completion adopted without fencing

Correct, and the framing is the useful part: this is the same ABA class a third time, at the entry points that **adopt** state rather than write it. I had fenced the three write-shaped paths and left the two read-shaped ones assigning `confirmed` directly.

**Load.** A remount for a project already loaded this session paints from the session cache and is *live* before its load lands (#172) — the dashboard renders, effects run, the operator can act. Intent queued against the cached revision was then installed over by the load with no gate, so a foreign A→B→A that happened while nothing was mounted was silently overwritten.

**Seed completion.** Same shape, and worse. A seed that reduces to a no-op is accepted from its stale revision-0 base, so the board it returns can be another writer's entirely — the seed path had no equivalent of the `applied` check the mutation path already had.

**There is now exactly one gate.** All five authoritative adoptions — poll, 409, accepted write, initial load, seed completion — install through `adopt`, and nothing else assigns `confirmed`. Two supporting changes:

- `attemptSeed` consumes the server's `applied` verdict, so a no-op seed is treated as a foreign adoption instead of as our own work.
- An authored seed re-observes the keys it wrote, so intent formed against the optimistic seed paint — closing the very node being seeded — is not fenced by our own seed. That is the over-fencing direction, and it has its own regression.

One thing I did beyond the ask, because it is the fragility the review actually named: the accepted-write path hand-rolled the gate's body inline. Fencing correctly in five places is not the same property as having one gate — the next person to add an adoption path copies whichever one they read first. It routes through `adopt` now, and the only two remaining `confirmed =` assignments are inside the gate itself and the seed's local optimistic paint, which installs no server state and is replaced by a gated `adopt` the moment the seed answers.

### Contract text

Updated to the ratified sentence:

> **One project-scoped shared store over a revision-fenced durable API, fencing every authoritative adoption with server-authored monotonic revisions per logical key, canonical across aliases and preserved through restart and causally safe tombstone GC.**

The "every authoritative adoption" clause is now written out in the contract text with the five paths named and the one-gate invariant stated, so the next reader does not add a sixth.

For the settings surface, this adds one requirement to what round 2 documented: a surface adopting this pattern must funnel **every** path that installs server state through a single gate, including its initial load and any migration/seed completion. Those are the paths that look like reads and are not.

### Tests

Two new regressions, both failing before the fix:

- `intent queued on a cached remount loses to a foreign A→B→A the load reveals`
- `intent queued during a seed that turns out to be a no-op loses to a foreign A→B→A`

Plus one guard that passes today and would break under a naive fix — worth calling out as such rather than presenting it as a caught bug:

- `intent formed after an authored seed survives that seed landing` — over-fencing here would silently eat the operator's action

### Verification

- Focused run by path: 235 pass, 0 fail
- Clean checkout of the new head, including dashboard and viewer mount points: 428 pass, 0 fail across 37 files
- `src/lib/accounts/` and `src/app/api/conversations/`: 2 failures, both pre-existing and unchanged across all three rounds
- Isolated `HOME`, `XDG_CONFIG_HOME`, `TMPDIR`, `LLV_STATE_DIR` throughout; no runtime or registry tree swept
- `tsc --noEmit` clean, `eslint` clean on changed files, privacy publication gate PASS

Not merged, not deployed.

Latand:
## Review round 4 addressed — 9ff2734a

### BLOCKING: migration restarted causal history

Correct. `migrateBoardProjects` folded source boards into the target while keeping only the **target's** `keyRevisions`, so everything the sources had accumulated was discarded. A migrated key is the same logical key, so rewinding its clock past writes that really happened un-fences intent formed before them.

Sources' history now merges in by maximum — floors included, since a floor is a claim about forgotten history and taking the lowest would under-report what every absent key has already seen.

One case needed explicit handling: the branch where the merge moves no **content**. `sameReduced(target, merged)` returned the target untouched, which is exactly when the loss was silent — identical arrangement, discarded history. History moving is itself a change worth persisting, so that branch now compares causal history too via `sameCausalHistory`.

I checked the two neighbouring write paths for the same shape. `transferBoardPathPlacements` is already correct for a different reason: a transferred path goes absent→placed in the target, so `boardKeysChanged` stamps it at the target's own new revision. Inheriting the *source project's* clock there would actually be wrong — the two projects have independent revision counters, and importing a clock of 800 into a document at revision 6 would over-fence that key until the target caught up. The single-source rename fast path carries revision, keys and floor intact, which is right for a pure rename.

### HIGH: the gate was not monotonic

Also correct, and it is my regression — the round-3 gate installed whatever board it was handed. Out-of-order responses (overlapping polls, a load racing the write that overtakes it) rewound the rendered board **and** the session cache a later remount primes from, so a window that is durably present visibly disappears and then comes back later.

One comparison now lives inside `adopt` and nowhere else. Taking the meta-point seriously: I had been adding a gate per entry point, and the invariant needs to be *in* the gate, not replicated beside each caller. `load` also decides monotonicity **before** the seed branch, which could otherwise assign a late revision-0 response straight to `confirmed` and bypass the check entirely — the same bug one line earlier.

**A trade-off I want visible rather than buried.** Refusing older revisions unconditionally means that if the durable board itself ever goes backward — a state directory reset, a restore from an older backup — an open tab refuses the rewind for the rest of its session, until a reload clears the session cache. I considered the causally precise alternative (discard a response only if `confirmed` moved *while that read was in flight*, which permits a genuine server rewind), and did not take it: it is a subtler rule than the one asked for, and it would accept a late echo in any case where the timing is not observable. Refusing a real rewind costs a reload; accepting a late echo silently deletes work. The cost is written into the contract text. Say the word if you want the other rule.

### One thing the fix exposed

`useBoardState.sharedStore.dom.test.tsx` began failing in full-directory runs but not alone. The module-level session cache and shared stores were reset only in `afterEach`, so a case in another file that settled the same project name left a confirmed board behind; priming from it made this file's revision-1 stub look like a backward response and it never loaded. Real test-isolation bug, previously invisible because adoption was unconditional. Reset moved to `beforeEach`.

### Contract text

Now reads:

> **One project-scoped shared store over a revision-fenced durable API, fencing every authoritative adoption with server-authored monotonic revisions per logical key, canonical across aliases and preserved through restart, migration and causally safe tombstone GC.**

with **preserved through migration** and **monotonic adoption** added as annotated clauses.

For the settings surface, this adds two requirements to what rounds 2–3 documented: any document-level operation that merges or renames documents must carry causal history forward by maximum, and the single adoption gate must refuse to install anything older than what it already holds.

### Tests

Both new, both failing before their fix:

- `src/lib/board/keys.test.ts :: a project-key migration carries source causal history forward`
- `src/hooks/useBoardState.convergence.test.ts :: a late response carrying an older revision never rolls the board backward` — asserts both the rendered board and, via a remount, the session cache

### Verification

- Focused run by path: 237 pass, 0 fail
- Clean checkout of the new head, including dashboard and viewer mount points: 430 pass, 0 fail across 37 files
- `src/lib/accounts/` and `src/app/api/conversations/`: 2 failures, pre-existing and unchanged across all four rounds
- Isolated `HOME`, `XDG_CONFIG_HOME`, `TMPDIR`, `LLV_STATE_DIR` throughout; no runtime or registry tree swept
- `tsc --noEmit` clean, `eslint` clean on changed files, privacy publication gate PASS

Not merged, not deployed.

PR #772, opened 2026-07-29
fix(scheme): always-lasso background drag without native text selection, plus a hover-revealed selection check

### LINKED ISSUE #771: Board selection: background drag must always lasso without native text selection, plus a hover-revealed check that toggles canonical selection
## Current state (grounded in code, not a rebuild)

A marquee already exists. This issue is scoped to the two places it falls short of the intended behaviour, plus the affordance that is missing outright.

- `src/components/scheme/useLasso.ts` implements the marquee: it claims the background press through the camera's `onBackgroundDown`, tracks the drag on window listeners, commits the intersected node set on release, arms from `DRAG_MIN = 4`, cancels on Escape, and treats Shift as additive.
- `src/components/scheme/SchemeBoard.tsx` owns the one selection set (`multi`); `commitMarquee` merges additively, and `selectionBBox` / `pruneSelection` come from `src/components/scheme/lasso.ts`.
- `SchemeBoard.tsx` already reports `selectedPaths: orderedSelection(layout, multi)` into the view slice, so the `selected` scope in the Viewer snapshot already observes board selection. `src/lib/view/types.ts` carries `selectedPaths: string[]` and `ViewScopeKind` already includes `"selected"`.
- `src/components/scheme/nodes.tsx` renders an accent check badge plus a member tint when a node is `marked`.

## Gap 1 — a background drag leaves native text selection behind

`useLasso.onBackgroundDown` returns `true` to claim the press but never calls `preventDefault()`, and nothing applies `user-select: none` for the duration of an active marquee. `SchemeBoard.tsx` applies `select-none` only while `panning`. Dragging the background across cards therefore starts a native browser text selection underneath the marquee, and the gesture ends with stray highlighted text.

Required: a background drag always starts the lasso and suppresses native text selection for the whole gesture, including while the rect crosses cards. Deliberate selection inside conversation content must keep working — do not blanket-disable user selection on transcript panes to achieve this.

## Gap 2 — no hover-revealed check to toggle one card

The check in `nodes.tsx` is a non-interactive `<span>` that renders only once a node is already selected, so a single card cannot be added to or removed from the selection by pointing at it.

Required: hovering near a card edge reveals a subtle semi-transparent check; clicking it toggles that card inside the existing blue canonical selection; clicking again deselects. It must introduce no layout shift, must not obstruct card content, and must not conflict with scrolling or text selection. It needs a keyboard path (focusable, visible focus ring, Enter/Space toggles) and a touch path, since touch has no hover.

## One shared selection model

The lasso, the hover check and the existing selection gestures all read and write the same `multi` set. Preserve the existing `selectedPaths` contract exactly: same ordering through `orderedSelection`, same view-slice shape. A hover-check toggle must be visible in the snapshot's `selected` scope / `selectedPaths` immediately, so an orchestrator can inspect what the operator has selected.

## Tests

- Focused unit coverage for the text-selection suppression decision and for the toggle reducer.
- A DOM test proving a background drag that crosses a card leaves no native selection.
- A DOM test for hover reveal, click to toggle, repeat click to deselect, and the keyboard toggle.
- Assertions on `selectedPaths` after each gesture.
- Every test must be capable of going RED under a real mutation of the production path; an assertion nothing ever writes to does not count.

Rendered verification in the real app is required in addition to the tests.

## Delivery

Implementation lane only. No merge and no deploy without explicit operator authorization. A Fable visual critique of the rendered result is mandatory after implementation.

Latand:
Closes #771.

One canonical selection, owned at project scope, published by every view, with two
board gestures that write it.

### Gap 1 — a background drag no longer leaves native text selection behind

`useLasso.onBackgroundDown` claimed the background press but let its compatibility
mousedown through, so the browser anchored a text selection on the canvas and the
marquee crossing a card highlighted the transcript underneath.

The claim now `preventDefault()`s the press, drops whatever was already highlighted
(the clearing the suppressed mousedown owed), and takes the focus that press used to
take so the board keeps its arrow-key target and a composer inside a pane still
blurs. For the rest of the gesture a `selectstart` / `dragstart` guard refuses both.
The guard reads the pending press, so it goes quiet the instant the drag commits or
cancels — nothing is disabled on transcript panes. The viewport's `select-none` now
covers the live marquee as well as a pan, so the suppression is scoped to the
gesture rather than to the content. Touch keeps the default: the finger still pans
and taps.

### Gap 2 — a hover-revealed check toggles one card

The check was a non-interactive `<span>` rendered only once the card was already
selected. It is a button now, mounted on every card and revealed by opacity alone:

- coming near the card's top-right corner (or hovering the card) fades in a
  semi-transparent check; a member's check is filled accent;
- no layout shift and no obstruction — both the control and its hover zone sit
  outside the card's content box, and the zone's inner edge stops short of the
  pane's own header controls, so favorite / expand / delete / close keep every
  pixel of their hit area;
- keyboard path for free: a native `<button>` with a visible focus ring;
- touch path: `@media (hover: none)` keeps the checks visible and grows the tap
  area away from the card;
- a running session reveals every card's check (the panes are click-through then),
  and pointing at one brings its card fully back out of the session dim.

### Fable follow-up — a non-member's check no longer reads as disabled

The session dim was an `opacity`/`filter` on the card shell, and the check lives
inside it. Ancestor opacity and filter composite the whole subtree and no descendant
can opt out, so every non-member's toggle rendered as a faint grey circle — the
grammar of "disabled" — on precisely the control that invites extending the
selection. The dim now lands on the shell's **children** with the check's zone
excluded: the card recedes exactly as before (its content is one child, so it still
fades as a single layer) while the toggle keeps full contrast. A non-member's check
also carries more weight inside a running session than in the idle reveal. Idle
boards are untouched — at rest the check is still fully transparent.

## Part 2 — one selection that survives a mode switch

`multi`/`armed` were `useState` inside `SchemeBoard`, so leaving scheme mode
unmounted their owner and destroyed the selection; `ProjectDashboard` and
`MobileFocusView` both hardcoded `selectedPaths: []`.

**Lifted, not duplicated.** The selection now lives beside the durable board at
project scope in `useBoardState`, which already outlives every view switch.
Ephemeral on purpose: a project-keyed session map, never written to `prefs`, never
PATCHed, gone on reload — so it also works while the durable board is unavailable.
It is read *through* that map rather than copied into each store, and a write
notifies every holder of the project, so a second binding cannot publish a set the
operator has moved past. A store recreated behind a remount primes from the session
on its **first** snapshot, so a returning view paints the selection instead of
blinking empty. Three writers, each a thin wrapper over a pure reducer:
`toggleSelection`, `commitSelection`, `clearSelection` (plus the armed latch and
`pruneSelectionTo`). `toggleMember` remains the single membership writer for every
board gesture.

**The publication contract.** `selectedPaths` is the canonical selection projected
onto the current view's render order — same set everywhere, per-view order, the
stored set never reordered by publishing it. `orderedSelection` is byte-identical
to what it always published.

What happens to a member a view's order does not mention has two deliberate
answers, and the second one came out of the rendered run:

- the **scheme** omits it — its order *is* the board, so a path the board does not
  place has no board presence to report;
- the **list** and the **phone** append it. Their orders are not the authority on
  membership: the flat list only lists ROOT conversations, so a selected subagent
  leaf appears in no row at all. Dropping it there made the real snapshot report
  `[]` after a mode switch — indistinguishable, to the orchestrator, from the
  operator deselecting everything. The cap holds across both phases.

**Pruning — the trap.** The board's own prune effect is gone; it pruned against
`layout.nodes`, which is exactly wrong once the set outlives the view. The one
legitimate prune lives in `ProjectDashboard`, where the scan is: drop a member only
when its conversation is gone from the scanned files, never against an empty or
unloaded scan, so a failed poll cannot read as "every conversation disappeared".

**`visiblePaths`** falls back to the board's placed windows when the history list
has no rows of its own. Honest note: on the current `resolveProjectView` logic that
branch is only reachable when the project has no conversations at all, so this is
defensive rather than an observable fix — I did not manufacture a test for it.

**The snapshot side is unchanged.** `src/lib/view/snapshot.ts` is byte-identical to
`main`: `scopedPaths` already resolves `selected` / `focused-selected` from
`session.selectedPaths`, and `validateExplicitMembership` already admits them,
neither caring which mode reported them. New tests pin that down so `selected`
cannot quietly become a scheme-only scope.

## Tests

Unit: the background-press plan, the toggle / marquee-commit / prune reducers, and
the ordering contract (per-view order, omission vs. append, the cap in both phases).

`useBoardState.selection.test.ts`: survives a store recreation, primes on the first
snapshot, one set across two holders, per-project isolation, prune keeps what still
exists and drops what is gone, an unchanged prune notifies nobody, live while the
board is unavailable, and selecting never writes to the durable board.

`SchemeBoard.selection.dom.test.tsx`: the lasso gesture (press `defaultPrevented`,
`selectstart` refused mid-gesture, empty native selection, selection handed back on
release), cancel, hover reveal → toggle → repeat-click deselect, keyboard
activation, survival across unmount/remount, a relayout keeping membership, and the
check's dim exemption plus session weight.

`ProjectDashboard.selection.dom.test.tsx`: scheme → list → scheme round trip, a
member with no list row still published, multi-card selection in the list's order,
the phone's focus and map modes, a member the phone's board does not place, and
pruning asserted on the SET (a disappeared conversation does not resurrect as
selected when it returns; one the board merely stops PLACING stays a member across a
scan tick).

`selectionScope.test.ts`: `scope: selected` from a `list` / `mobile-focus` session
returns those conversations WITH their content, and an explicitly named selected
path is admitted as inside the current view.

**Every case verified RED against a mutation of the path it covers** — 36 in total:
dropping the press `preventDefault`, the `selectstart` guard, the selection clear,
the marquee `select-none`, the shared toggle reducer, the check's toggle handler,
the hover reveal, the session reveal, the shell's check-hover marker, four
variations of the press plan, both marquee branches, both prune halves, the un-lift,
each publisher back to `[]`, `visiblePaths` back to empty, pruning against a layout,
pruning without the empty-scan guard, no pruning at all, the per-store selection
copy, no priming on recreate, no cross-holder notify, order-blind projection,
`includeUnordered` not appending, the append-phase cap, `selected` resolving from
`visiblePaths`, `validateExplicitMembership` dropping `selectedPaths`, the check's
dim-exemption hook, and the session weight.

## Rendered verification

`scripts/capture-issue-771-selection.ts` boots the disposable demo fixture home,
serves it from the production build, and drives the real board in Chromium.
Observed:

- idle board completely clean, check computed opacity `0`; corner hover → `1`;
- a background drag from empty canvas across two cards: the marquee sits over four
  lines of real transcript text with **nothing highlighted**; on release both are
  members and the native selection is empty; a transcript drag straight after
  highlights normally;
- clicking a check selects exactly its own card, a second click drops the session;
- session dim, by computed style: shell `opacity 1 / filter none`, check zone
  `1 / none`, card content `0.35 / grayscale(0.7)`;
- mode switch, through the real `/api/agent/snapshot`: `mode=list`,
  `selectedPaths=1`, `visiblePaths=2`, `scope: selected` returned 1 conversation
  with 2 messages while the board was unmounted; back in scheme, still 1 member.

The rendered run is what caught the list-projection hole above — the DOM tests were
green at the time.

## Known limitations (recorded, deliberately out of scope)

- below ~50% zoom the check's glyph reads as a coloured dot, and touch targets
  shrink with the canvas — inherent to a zoomable board, every affordance scales
  that way;
- the pipeline stage chip's adjacency to the check has no demo fixture and needs one
  manual glance on a live pipeline board;
- the "N selected" label can overlap a single selected card's header title; this
  predates the PR.

Not merged and not deployed — no authorization for either.


Latand:
# Fable visual critique — the #771 selection UX, rendered

**What I rendered.** Production build served from the disposable demo fixture home, driven in headless Chromium at 1440×900. `bun scripts/capture-issue-771-selection.ts` produced the six scripted stills and every embedded assertion passed (check opacity 0 at rest → 1 on corner hover; native selection empty mid-drag and after; transcript drag after the lasso still highlights; toggle on/off round-trips). I then extended the harness locally (this review worktree only, nothing pushed) to visit the states the script skips: fit-all **40%** and **160%** corner hovers, the card-center discovery reveal, a true dim-lift (session running, a **non-member's** check hovered), the relay flow-strip board at 66%, mobile emulation with `(hover: none)` matching, and `prefers-reduced-motion: reduce`. Run note for deep checkouts: the fixture tmux socket overflows the unix path limit, so the capture needs `LLV_DEMO_TMUX_TMPDIR` pointed at a short directory — already documented in `scripts/demo-capture.ts`.

## What is genuinely good

- **The idle board is completely clean.** `771-idle.png` shows zero new pixels at rest. On a surface this dense that restraint is the single most important property of the feature, and it holds.
- **The three check weights are distinguishable at a glance.** Card hover shows a soft `accent/35` circle (`771x-card-hover.png` — present without shouting), pointer-on-check deepens to `accent/70` (`771-check-revealed-crop.png`), member is solid accent with a shadow (`771-check-selected-crop.png`). Membership additionally always co-occurs with the dashed bbox, the dim and the "N selected" bar, so misreading hover as membership is realistically impossible.
- **Placement clears the card chrome at every zoom I rendered** (40 / 50 / 66 / 160%). `771x-near-hover-crop.png` at 160% is the stress case: the circle floats off the rounded corner with comfortable air above the crown / expand / delete / close buttons. `771y-relay-crop.png` shows the same clearance on the busiest card available (flow-strip builder). Reveal is opacity-only — no layout shift anywhere in the series.
- **The lasso looks right.** `771-lasso-active.png` has the marquee veil crossing real transcript text with nothing highlighted underneath, and the crossed cards carry a thin solid candidate ring that reads clearly against the committed member language (dashed bbox + tint + filled checks) in `771-lasso-committed.png`. The committed session frame is coherent: bbox, "2 selected", receded non-members, composer bar.
- **The dim-lift reads as intentional.** `771y-dim-lift.png`: with one card selected, hovering a non-member's check pops that whole card back to full colour while the rest of the board stays receded — it reads precisely as "this one is about to join". Nice detail.
- **Touch is quiet.** Under `(hover: none)` (`771x-touch-board.png`) the always-visible checks are faint `accent/35` dots at each corner; on the four-card fixture the board reads calm, and the grown tap area points away from the header controls as promised.
- **Reduced motion has state parity** — same rendered result, `transitionDuration: 0s`.

Adjacent-card flicker I could only judge by geometry, since stills cannot show motion: each card reveals only its own check via `:hover`, fades are 0.16s, and fixture gutters dwarf the 20px hot-zone overhang — nothing suggests strobing on a pointer sweep.

## Findings, ranked

**Must fix before merge: nothing.** I looked for a blocking visual defect and did not find one.

**Would improve:**

1. **Non-member checks look disabled during a session.** The session reveals every card's check, and the session dim (35% opacity + `grayscale(0.7)`) then renders non-member checks as faint grey circles — visible on the receded card in `771-check-selected.png` and at the left edge of `771x-dim-lift.png`. Grey-on-grey is the visual grammar of "disabled", yet these are exactly the controls inviting the operator to extend the selection. The hover dim-lift rescues the affordance once the pointer arrives; at rest it undersells it. Worth exempting the check from the grayscale, or giving non-member checks a modestly higher weight while a session runs.
2. **Below ~50% zoom the glyph dissolves into a coloured dot,** and on touch the fitted board makes the always-visible checks (~11px on screen at 32%, `771x-touch-board.png`) too small to tap without zooming in. Both are inherent to a zoomable canvas and every other affordance on the board scales the same way — noting it as a known limitation is fine; a minimum on-screen size at far zoom (the way identity labels behave) would be the polish version.
3. **The pipeline stage chip adjacency is unrendered.** The chip mounts at the card's top-right (`-top-3 right-3`), directly beside the check zone; the demo home has no pipeline fixture, so I could only verify by geometry (~6px clearance, chip at z-7 under the check's z-8). One manual glance at a live pipeline board would close this.
4. **Pre-existing, for a follow-up if at all:** the "N selected" label can sit over the selected card's own header title when the bbox hugs a single card (`771-check-selected.png`). That behaviour predates this PR.

## Merge recommendation (visual / interaction standpoint only)

**Ready.** The control is quiet when it should be quiet, legible when it matters, and it sits inside the existing selection language rather than adding a second one. Finding 1 is the only change I would actually make, and it does not need to gate the merge.

Latand:
# Fable visual critique, second pass — final head (`0f58b59e`)

**What I rendered.** Same isolated setup as pass 1: production build of the final head served from the disposable demo fixture home, headless Chromium at 1440×900. I ran the PR's extended `scripts/capture-issue-771-selection.ts` (every assertion passed, including the new computed-style dim-shape checks and the list-mode snapshot read), then drove the two review areas myself with a local extension: a session at rest with the pointer parked away from every card, the scheme → list → scheme round trip, and the same page shrunk to 390×844 so the phone modes take over the identical in-page store with the selection still live.

Scope per the operator: only the two changed areas plus anything newly broken. Pass-1 verdicts stand and are not re-argued here.

## Area 1 — the dim exemption (my pass-1 finding 1)

**Taken well. This is the better board.**

- **The non-member check now reads as genuinely interactive at rest.** `771-session-nonmember-check.png` is the direct evidence: the receded card's content is washed to 35% and greyed, and its check sits at the corner in full-contrast accent (`accent/60` + shadow in a session — deliberately heavier than the idle `accent/35` reveal, which is the right hierarchy: in a session it is an invitation, at idle it is a hint). The grey "disabled" reading from pass 1 is gone.
- **The receded quality survived the exemption.** The computed-style checks confirm the intent precisely — shell `opacity 1 / filter none`, check zone `1 / none`, content `0.35 / grayscale(0.7)` — and `772-session-board.png` (pointer parked on empty canvas) shows the non-members still reading unambiguously as background. Each receded card now carries exactly one small saturated dot at its corner; on the fixture board that dot reads as "add this one" rather than as noise, and because it is the only saturated element on a grey card it does its job without a hover.
- **Idle cleanliness — the property I was asked to guard — is untouched.** `771-idle.png` at this head is still zero-new-pixels; the exemption only changes what a running session looks like, and a session is a deliberate, transient mode. The board does not read busier where it matters.

One honest caveat: I judged the session density on a four-card fixture. On a fifty-card board a session paints fifty accent dots over grey; I expect it to hold (uniform, small, semantically load-bearing) but I could not render that density.

## Area 2 — the selection surviving a mode switch

**The persistence itself is right, and the scheme round trip is seamless.** Select a card, switch to conversations, switch back: `772-back-to-scheme.png` is indistinguishable from `772-session-board.png` — dashed bbox, member colour, "1 selected", composer bar, all restored. The snapshot read from list mode returned the member. No complaints about the mechanism.

**The list gives the operator no legible signal that a selection is active.** Saying it plainly, as asked: `771-list-mode.png` shows the member's row pixel-identical to every other row — no tint, no check, no count anywhere on the screen. Meanwhile the selection is still live: "one message to every selected agent", "Clear from board" and the orchestrator's `selected` scope all remain keyed to it. That is an operator trap — a selection made minutes ago in scheme mode silently steers actions taken from a view that swears nothing is selected.

*Minimum affordance:* a compact **"N selected" pill next to the scheme/conversations toggle** (it is the one element both views share, and it is where the eye already goes on a mode switch), with tap-to-clear or tap-to-return-to-scheme. The fuller fix — member marking on the matching rows — is worth having, and note the pill covers the case rows cannot: a selected subagent leaf has no row in this root-only list, which is exactly what my rendered member was.

**Both phone modes have the same invisibility, plus one confusion hazard.** `772-phone-focus.png`: no trace of the live selection anywhere in the focus layout. `772-phone-map.png`: zero member marking on the map (`data-lasso-selected` count is 0 while the store holds a member), yet the map's *focused-node* highlight is a lavender fill with an accent ring — visually adjacent to the desktop's membership language, so an operator who selected cards on the desktop and opens the phone map can read the focus highlight as their selection and be wrong about both. The phone also offers no toggle affordance, so on the phone the selection is simultaneously invisible, uneditable, and still published to observers. The same "N selected" pill in the phone header (and eventually a member ring on map nodes, distinct from the focus highlight) is the minimum.

## Newly broken

Nothing found. The lasso states, check weights, placement, touch reveal and reduced-motion parity all re-rendered identical to pass 1 at this head, and the idle board is unchanged.

## Ranked findings

1. **List mode renders no signal of a live selection** (`771-list-mode.png`) — the one thing I would address before merge, since invisible persistence is the direct cost of this PR's own headline improvement. The pill described above is a small, contained fix.
2. **Phone modes: selection invisible and uneditable, and the phone map's focus highlight can masquerade as membership** (`772-phone-focus.png`, `772-phone-map.png`) — would improve; the publishing fix underneath is correct and observers now see the truth.
3. Session density on genuinely dense boards is extrapolated from a four-card fixture — a note for a later look, no action implied.

## Merge recommendation (visuals / interaction only)

**Area 1: merge-ready, and strictly better than what I reviewed in pass 1.** **Area 2: the model and the scheme-side rendering are ready; I would land the list-mode "N selected" pill before merge** — or have the invisible persistence explicitly accepted as a known gap with the phone treatment queued behind it. Nothing else stands between this head and merge on visual grounds.

PR #778, opened 2026-07-30
fix(mcp): publish the closed argument sets the servers enforce

### LINKED ISSUE #774: MCP argument schemas advertise open objects but validate against closed allowlists — 119 guessed-key rejections in 10 days
Several MCP tools advertise a permissive argument schema and then reject the call against a stricter server-side allowlist. The tool schema is the only thing a caller can see, so when it says "any object here", the caller guesses — and the guess is rejected by a validator whose allowed set was never published.

The failure is silent in the worst way: the schema is not wrong, and the validator is not wrong. They just disagree, and only the validator is authoritative.

## Evidence

Distinct rejection messages across agent transcripts, last 10 days:

| Count | Rejection | Surface |
|---|---|---|
| 131 | `unknown role` | `spawn_agent.role` |
| 24 | `unknown pipeline action` | `pipeline_action.action` |
| 18 | `unknown scope.mode` | `operator_snapshot.scope` |
| 18 | `invalid scope.kind` | `operator_snapshot.scope` |
| 17 | `unknown text.maxChars` | `operator_snapshot.text` |
| 16 | `unknown text.maxRecords` | `operator_snapshot.text` |
| 15 | `unknown view.includeFrame` | `operator_snapshot.view` |
| 11 | `invalid role parameter: lens` | `spawn_agent.roleParams` |
| 8 | `unknown text.mode` | `operator_snapshot.text` |
| 6 | `unknown view.includeSelection` | `operator_snapshot.view` |
| 6 | `unknown scope.project` | `operator_snapshot.scope` |
| 3+3+3+3+3 | `view.includeConversations`, `view.includeBoard`, `view.detail`, `text.maxRecordsPerConversation`, `scope.include` | `operator_snapshot` |
| 5+3+3+3+1+1 | `unknown role parameter:` `confirm`, `src`, `baseRef`, `access`, `task`, `displayName` | `spawn_agent.roleParams` |

`operator_snapshot` alone accounts for **119 rejected calls** whose only defect was a guessed key name. Every one of those is a wasted round trip that the caller then has to debug by reading server source.

## Root cause

`src/lib/mcp/server.ts` declares the nested argument objects as free-form records:

```ts
operator_snapshot: z.object({
  clientRequestId: clientRequestIdSchema,
  schemaVersion: z.literal(1).optional(),
  view: z.record(z.string(), z.unknown()).optional(),
  scope: z.record(z.string(), z.unknown()).optional(),
  text: z.record(z.string(), z.unknown()).optional(),
  caller: z.record(z.string(), z.unknown()).optional(),
}).passthrough(),
```

`src/lib/view/validation.ts` then enforces exact keys:

```ts
if (textOptions) exact(textOptions, ["include", "lastMessages", "maxCharsPerConversation"], "text");
```

`exact()` reports only the offending key:

```ts
if (unknown) throw new ViewValidationError("INVALID_REQUEST", `unknown ${field}.${unknown}`);
```

So `unknown text.mode` names what is wrong and never names what is right. A caller cannot self-correct from the error, and cannot pre-empt it from the schema.

The same shape repeats where a fixed set is published as an open type:

- `pipeline_action.action` is `z.string().min(1)`; the route checks membership in a 17-entry `ACTIONS` set.
- `spawn_agent.role` is `z.string().optional()`; `resolveRole` checks membership in the role registry and answers `unknown role`.
- `spawn_agent.roleParams` is a free record, correctly so — its valid keys depend on the chosen role — but the rejection still does not say which keys that role accepts.

Several guessed keys are real argument names borrowed from a neighbouring tool: `mode` is a genuine `lifecycle_events` argument, which is exactly what an open schema invites.

## Required behaviour

1. **A tool's argument schema publishes its accepted shape.** Where the accepted set is fixed and known at build time, it appears in the schema rather than only in a server-side validator.
2. **A rejection names the accepted alternatives.** `unknown text.mode` becomes `unknown text.mode (allowed: include, lastMessages, maxCharsPerConversation)`. This holds for closed-set rejections generally: unknown key, unknown role, unknown pipeline action, unknown role parameter.
3. **No behaviour change for valid calls.** Every call accepted today is accepted after the change, byte-identical in result. This is an error-path and schema-surface change only.
4. **The allowlists stay single-sourced.** The schema and the validator derive their accepted set from one constant, so a future key cannot be added to one and forgotten in the other. A duplicated literal list is the defect being fixed, reintroduced.
5. **`roleParams` stays open in the schema** — its shape is role-dependent — but its rejection enumerates the selected role's parameters.

## Scope

In scope, ordered by measured rejection volume:

- `operator_snapshot`: real object schemas for `view`, `scope`, `text`, `caller`
- `spawn_agent`: `role` and `roleParams` rejections name the available roles / that role's parameters
- `pipeline_action`: `action` becomes a closed enum in the schema, single-sourced with the route's `ACTIONS`
- `exact()` in `src/lib/view/validation.ts` and `src/lib/board/validation.ts`: rejections name the allowed keys

Out of scope, recorded here so the follow-up is not lost: `request_attention.target` (a discriminated union, 1 rejection observed), `create_pipeline.stages`, `flow_action.roles`. These share the pattern at a frequency that does not currently justify the churn.

## Tests

Each guarantee needs a test that fails without the change:

- a schema-level rejection for each newly typed `operator_snapshot` field, asserting the caller sees the constraint at the tool boundary
- an `exact()` rejection asserting the allowed keys appear in the message
- `unknown role` and `unknown role parameter` rejections asserting the enumeration
- `pipeline_action` asserting the enum and the route's `ACTIONS` set stay in agreement
- a regression asserting every argument shape accepted today is still accepted

Latand:
Several MCP tools advertised a permissive argument schema and then rejected the call against a stricter server-side allowlist. The tool schema is the only thing a caller can see, so when it says "any object here", the caller guesses — and the guess dies against a list that was never published anywhere.

Neither side was wrong on its own. They disagreed, and only one of them was authoritative.

## Evidence that motivated the scope

Distinct rejection messages across agent transcripts, last 10 days:

| Count | Rejection | Surface |
|---|---|---|
| 131 | `unknown role` | `spawn_agent.role` |
| 24 | `unknown pipeline action` | `pipeline_action.action` |
| 18 | `unknown scope.mode` | `operator_snapshot.scope` |
| 18 | `invalid scope.kind` | `operator_snapshot.scope` |
| 17 | `unknown text.maxChars` | `operator_snapshot.text` |
| 16 | `unknown text.maxRecords` | `operator_snapshot.text` |
| 15 | `unknown view.includeFrame` | `operator_snapshot.view` |
| 11 | `invalid role parameter: lens` | `spawn_agent.roleParams` |
| 8 | `unknown text.mode` | `operator_snapshot.text` |
| 6 each | `view.includeSelection`, `scope.project` | `operator_snapshot` |
| 3 each | `view.includeConversations`, `view.includeBoard`, `view.detail`, `text.maxRecordsPerConversation`, `scope.include` | `operator_snapshot` |
| 5+3+3+3+1+1 | `unknown role parameter:` `confirm`, `src`, `baseRef`, `access`, `task`, `displayName` | `spawn_agent.roleParams` |

`operator_snapshot` alone accounts for **119 rejected calls whose only defect was a guessed key name**. Several of those guesses are real argument names borrowed from a neighbouring tool — `mode` is a genuine `lifecycle_events` argument — which is exactly what an open schema invites.

## Schema mismatch classes this fixes

**1. Open object published against a closed key set.** `operator_snapshot`'s `view` / `scope` / `text` / `caller` were `z.record(z.string(), z.unknown())` while `validateSnapshotRequest` enforced an exact allowlist. They are now real object schemas.

They are `.strict()`, not plain `z.object`. A plain object schema *strips* unknown keys silently, which would have turned a wasted round trip into a wrong answer — the caller's `maxCharsPerConversation` typo would be dropped and the call would succeed with the wrong bound. A loud rejection is the correct trade here.

**2. Open string published against a closed enum.** `pipeline_action.action` was `z.string().min(1)` while the PATCH route admitted a fixed 18-entry set. It is now `z.enum(PIPELINE_ACTIONS)`, single-sourced with the route.

**3. Closed-set rejections that named no alternative.** `exact()` and `oneOf()` in the view and board validators, `unknown role`, and `unknown role parameter` all reported only what was wrong. They now name what is right:

```
unknown text.mode (allowed: include, lastMessages, maxCharsPerConversation)
invalid scope.kind (allowed: focused, selected, visible, focused-selected, paths)
unknown role: no-such-role (allowed: orchestrator, reviewer, verifier, builder, architect, cleaner, prod-auditor, deployer)
unknown role parameter: unexpected (reviewer accepts: diffSource, lens, mode, parallelN)
```

`roleParams` stays an open record in the schema — its valid keys depend on the chosen role, so it cannot be typed statically — but its rejection now enumerates that role's parameters.

## Single-sourcing

The allowlists live in one place each: `SNAPSHOT_*_KEYS`, `VIEW_SCOPE_KINDS`, `VIEW_RESOLUTIONS` and the snapshot bounds in `src/lib/view/types.ts`; `PIPELINE_ACTIONS` in `src/lib/pipelines/types.ts`. The tool schema and the server-side validator both read from those constants, so a key added to one and forgotten in the other cannot happen again. A duplicated literal list is the defect being fixed — reintroducing one would be the same bug wearing a different hat.

## Verification

Run locally; CI on this repo runs only the privacy gates, so green checks here prove very little on their own.

- 61 focused tests pass, 0 fail: `src/lib/view/validation.allowlists.test.ts`, `src/lib/view/view.test.ts`, `src/lib/board/validation.test.ts`, `src/lib/roles/registry.test.ts`, `src/lib/roles/parameters.test.ts`, `src/lib/mcp/server.test.ts`
- TypeScript exit 0
- ESLint exit 0 on all changed files
- privacy publication gate PASS
- `bun run build` exit 0
- `git diff --check` clean

New tests pin both halves: that a rejection names the accepted alternatives, and that the published tool schema and the validator admit the same key set. The second is the regression that actually matters.

One existing expectation changed deliberately — `registry.test.ts` pinned the old bare `unknown role parameter: unexpected` string. It now asserts the enumerated form, plus a new case for `unknown role`.

## Scope

10 files, +184/−46, confined to argument schemas, their validators, and the constants they now share. No runtime, bindings, scanner or component code is touched.

The broader Viewer HTTP ↔ MCP **method-surface** parity question — tools that are missing entirely rather than mis-typed — is deliberately NOT in this PR and remains a separate audit.

Closes #774

### DIFF
--- a/src/app/api/pipelines/[id]/route.ts
+++ b/src/app/api/pipelines/[id]/route.ts
@@ -2,18 +2,14 @@ import { NextRequest, NextResponse } from "next/server";
 
 import { requestPipelineTick } from "@/lib/pipelines/controllerSignal";
 import { patchPipeline, type PipelineCloseReport } from "@/lib/pipelines/engine";
-import type { PatchPipelineRequest, Pipeline, PipelineAction, PipelineRepoPreflightErrorCode } from "@/lib/pipelines/types";
+import { PIPELINE_ACTIONS, type PatchPipelineRequest, type Pipeline, type PipelineAction, type PipelineRepoPreflightErrorCode } from "@/lib/pipelines/types";
 import { rejectCrossOrigin } from "@/lib/sameOrigin";
 import type { ApiError } from "@/lib/types";
 
 export const runtime = "nodejs";
 export const dynamic = "force-dynamic";
 
-const ACTIONS = new Set<PipelineAction>([
-  "start", "update-draft", "set-position", "add-stage", "remove-stage", "reorder-stage", "set-edge",
-  "pause", "resume", "retry-stage", "skip-stage", "override-stage", "link-task", "unlink-task", "delete", "close",
-  "set-src",
-]);
+const ACTIONS = new Set<PipelineAction>(PIPELINE_ACTIONS);
 
 const CONTROLLER_ACTIONS = new Set<PipelineAction>(["start", "resume", "retry-stage", "skip-stage"]);
 
@@ -39,7 +35,9 @@ export async function PATCH(
   } catch {
     return NextResponse.json({ error: "invalid JSON" }, { status: 400 });
   }
-  if (!ACTIONS.has(body.action)) return NextResponse.json({ error: "unknown pipeline action" }, { status: 400 });
+  if (!ACTIONS.has(body.action)) {
+    return NextResponse.json({ error: `unknown pipeline action (allowed: ${PIPELINE_ACTIONS.join(", ")})` }, { status: 400 });
+  }
   const { id } = await ctx.params;
   try {
     const result = await patchPipeline(id, body);
--- a/src/lib/board/validation.ts
+++ b/src/lib/board/validation.ts
@@ -17,9 +17,10 @@ import type { BoardMutationV1 } from "@/lib/board/mutations";
 export const MAX_BOARD_BODY_BYTES = 48 * 1024 * 1024;
 export type BoardPatch = Partial<BoardProjectStateV1["prefs"]>;
 
+/* Names the accepted alternatives so a caller can self-correct (#774). */
 function exact(value: Record<string, unknown>, allowed: readonly string[], field: string): void {
   const unknown = Object.keys(value).find((key) => !allowed.includes(key));
-  if (unknown) throw new ViewValidationError("INVALID_REQUEST", `unknown ${field}.${unknown}`);
+  if (unknown) throw new ViewValidationError("INVALID_REQUEST", `unknown ${field}.${unknown} (allowed: ${allowed.join(", ")})`);
 }
 function record(value: unknown, field: string): Record<string, unknown> {
   if (!value || typeof value !== "object" || Array.isArray(value)) throw new ViewValidationError("INVALID_REQUEST", `invalid ${field}`);
--- a/src/lib/mcp/server.test.ts
+++ b/src/lib/mcp/server.test.ts
@@ -6,8 +6,12 @@ import path from "node:path";
 import { Client } from "@modelcontextprotocol/sdk/client/index.js";
 import { InMemoryTransport } from "@modelcontextprotocol/sdk/inMemory.js";
 
+import { PIPELINE_ACTIONS } from "@/lib/pipelines/types";
+import { SNAPSHOT_CALLER_KEYS, SNAPSHOT_SCOPE_KEYS, SNAPSHOT_TEXT_KEYS, SNAPSHOT_VIEW_KEYS } from "@/lib/view/types";
+
 import {
   MCP_TOOL_NAMES,
+  TOOL_INPUT_SCHEMAS,
   FileMcpReceiptStore,
   MemoryMcpReceiptStore,
   createViewerMcpServer,
@@ -1146,4 +1150,22 @@ describe("MCP tool service", () => {
       await server.close();
     }
   });
+});
+
+/* #774: `pipeline_action.action` was `z.string().min(1)` while the PATCH route
+   admitted a fixed set, and `operator_snapshot`'s nested objects were free-form
+   records against an exact-key validator. Both now read from one constant. */
+test("#774 tool schemas publish the closed sets their servers enforce", () => {
+  const pipelineAction = TOOL_INPUT_SCHEMAS.pipeline_action.shape.action;
+  expect(pipelineAction.options).toEqual([...PIPELINE_ACTIONS]);
+
+  const snapshot = TOOL_INPUT_SCHEMAS.operator_snapshot.shape;
+  expect(Object.keys(snapshot.text.unwrap().shape).sort()).toEqual([...SNAPSHOT_TEXT_KEYS].sort());
+  expect(Object.keys(snapshot.view.unwrap().shape).sort()).toEqual([...SNAPSHOT_VIEW_KEYS].sort());
+  expect(Object.keys(snapshot.scope.unwrap().shape).sort()).toEqual([...SNAPSHOT_SCOPE_KEYS].sort());
+  expect(Object.keys(snapshot.caller.unwrap().shape).sort()).toEqual([...SNAPSHOT_CALLER_KEYS].sort());
+
+  /* Strict, not stripping: an unknown nested key must stay a loud rejection. */
+  expect(snapshot.text.unwrap().safeParse({ mode: "digest" }).success).toBe(false);
+  expect(snapshot.view.unwrap().safeParse({ includeFrame: true }).success).toBe(false);
 });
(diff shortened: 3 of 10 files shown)


Latand:
## Review verdict: REQUEST_CHANGES

Fresh independent read-only review at exact head `a801e80b607b0930a29bece28e017e68c3f748bb`, as part of the MCP release scope (with #781 and release delta `efbb09e9..0d8b9643`).

Two of the divergences #774 exists to close are still open at this head.

### HIGH — `spawn_agent.role` is still published as open text

Evidence: `src/lib/mcp/server.ts:1001-1018`, `src/lib/roles/registry.ts:109-112`, `src/lib/roles/types.ts:1-9`.

The server enforces eight fixed role IDs, while the published MCP schema exposes `role` as an unrestricted string. Independently confirmed against the live surface: the advertised `spawn_agent` schema carries a bare `{"type": "string"}` for `role`, while `/api/roles` returns exactly eight IDs. This is the highest-volume #774 divergence and it remains unresolved by this PR — a guessed role still fails past the protocol boundary rather than at it.

- **Fix intent:** publish role IDs as a schema enum derived from one canonical role-ID source.
- **Constraints:** keep `roleParams` open, since its keys depend on the selected role; preserve overrideable role config and current valid-call results.
- **Acceptance:** `listTools` publishes every registry role exactly once; an unknown role fails at the protocol schema boundary; registry errors still enumerate alternatives; every valid role still resolves.

### HIGH — `operator_snapshot` schema still differs from its validation

Evidence: `src/lib/mcp/server.ts:1140-1161`, `src/lib/view/validation.ts:22-44`, `src/lib/view/validation.ts:74-91`.

The top-level object remains `.passthrough()` although validation enforces `SNAPSHOT_REQUEST_KEYS`. The schema also accepts invalid scope combinations, duplicate paths, and empty or oversized strings that validation then rejects. Independently confirmed: the advertised schema exposes an open top-level object plus open `scope`/`text`/`view`/`caller` objects.

- **Fix intent:** publish a strict top-level object, model `scope` as a conditional/discriminated shape, and share string/path constraints with validation.
- **Constraints:** keep `clientRequestId` as the MCP transport field; preserve valid response bytes.
- **Acceptance:** generated schema exposes `additionalProperties: false`; `kind=paths` requires paths and other kinds reject them; empty, oversized and duplicate paths fail at the MCP boundary; parity tests compare every schema-accepted shape against `validateSnapshotRequest`.

### Separately tracked (LOW, follow-up, not blocking this PR)

Concrete HTTP-vs-MCP parity gaps: flow creation, workflow listing/creation/actions and attention resolution exist over HTTP with no MCP equivalent, and `pipeline_action` advertises only `pipelineId`/`action` while `stageId`, `taskId`, `pos`, `stage`, `stageIds`, `edge` and `to` stay hidden behind `.passthrough()`. Worth its own issue with a machine-checked capability matrix rather than expanding this PR.

### Gates

Exact-head execution for this PR was **not** run: the review checkout sat at the #781 head and the review fence prohibited checkout/worktree mutation, so this diff and its tests were inspected statically. `git diff --check` passed on all three ranges. Treat the static-only gate coverage as a gap to close before merge.

### Release status

Not merged; production stays at `efbb09e94cc30e0f8606493baa843abad9eceea9`. No deploy prepared or requested.

Issue #777, opened 2026-07-30
MCP runtime reads still carry the socket env coupling #486 removed from writes — deployment_status reports a disabled plane while the plane is live


Latand:
`deployment_status` returns `runtime events are disabled` from an agent session while the runtime plane is fully up: the runtime host is running, `LLV_RUNTIME_EVENTS=1`, `LLV_RUNTIME_HOST_SOCKET` is set in the service env, and the socket file exists and is being served. The same deployment accepts `deploy_exact_sha` through MCP and completes it.

#486 removed the runtime-socket environment coupling from the **mutating** Viewer MCP tools and was completed. The **read/evidence** path was left on the old mechanism, so the coupling #486 set out to remove still governs how agents observe deployments.

## Evidence

`deploy_exact_sha` routes through the live HTTP surface, which is why it works:

```ts
// src/lib/mcp/bindings.ts
const receipt = await control.post("/api/runtime/deployments", { revision, idempotencyKey: requestId(args) });
```

`deployment_status` still resolves the runtime in-process from the MCP process's own environment:

```ts
async function deploymentStatus(args, dependencies) {
  if (!dependencies.runtimeEventsEnabled()) throw new Error("runtime events are disabled");
  const client = dependencies.runtimeHostClient();
  if (!client) throw new Error("runtime host socket is unavailable");
```

`runtimeEventsEnabled()` requires a socket in *that* process's env:

```ts
// src/lib/runtime/flags.ts
export function runtimeEventsEnabled(env = process.env) {
  return !runtimeEventsRolledBack(env) && Boolean(runtimeHostSocket(env));
}
```

The MCP server runs outside the runtime-host container and carries no `LLV_*` runtime variables, and its entry point loads no service env file. So the guard is not reporting a disabled runtime plane — it is reporting that this process cannot see one. The distinction is invisible to the caller, which is most of the cost: the message names a deployment state that is not true.

## The silent half

The same coupling sits on the projection path, where it fails quietly instead of loudly:

```ts
async function deploymentsForProjection(dependencies) {
  if (!dependencies.runtimeEventsEnabled()) return [];
```

Deployment evidence is then simply absent from board projections, with no error and nothing in the payload saying evidence was dropped. An agent reading a projection cannot distinguish "no deployments" from "deployments unreadable from here".

## Required behaviour

[...]

Issue #765, opened 2026-07-29
Skipped question cards stay pinned to the composer with no way to dismiss them


Latand:
## Problem

When an agent raises an interactive question card (the `AskUserQuestion` tool surface), the card stays pinned in the conversation **forever** once the operator declines or skips it. There is no close, dismiss, or skip affordance on the card itself.

The result is a dead card stuck at the bottom of a live conversation. It cannot be removed, it occupies vertical space in the composer area, and it visually competes with the real input box — so it actively gets in the way of continuing to work with that agent.

## Reproduce

1. Open any live agent conversation.
2. Have the agent raise a question card with two or more options.
3. Do not answer it — interrupt the agent, or send a different instruction instead.
4. The agent moves on and the question is functionally dead.

## Actual

The question card remains rendered and interactive-looking at the bottom of the conversation indefinitely. Nothing in the UI removes it. Scrolling does not retire it; it stays anchored near the composer.

## Expected

A question card the operator has skipped or superseded must be dismissible:

- an explicit close/skip control on the card,
- and the card should retire itself automatically once the question is provably moot — the agent has already continued past it, or the turn that raised it has ended without an answer.

A retired card should stay visible in the transcript as history if that is the existing convention, but it must stop occupying the live composer region and must stop presenting itself as awaiting input.

## Impact

Low severity, high friction. It degrades the primary surface the operator uses to steer agents, and it accumulates: every skipped question leaves another permanent obstruction in that conversation.

## Notes

Related lifecycle work already tracked separately: #757 (durable card retention — cards must not auto-disappear, completed workers fold instead) and #754 (closing a pipeline should hide internal stage windows). This issue is the mirror case: a card that *should* retire and currently cannot.

Latand:
## Operator report: the auto-retire half of this issue, with a pinned predicate

New operator sighting of the second bullet under *Expected* ("the card should retire itself automatically once the question is provably moot"): a yellow question block still renders as pending/actionable in the web transcript **after the operator has already moved on with later input**.

Scoping this issue's auto-retire half to a precise rule so it can be implemented and tested on its own. The explicit dismiss control (first bullet) stays open here.

### Where it is decided

The yellow block is rendered by `QuestionCard` from `file.pendingQuestion`, whose single producer is `pendingQuestionFor()` in `src/lib/scanner/questions.ts`. That function walks the tail records newest-first and:

- collects `tool_result` ids into an `answered` set,
[...]

PR #781, opened 2026-07-30
fix: route MCP deployment reads through Viewer control

Latand:


Latand:
## Review verdict: REQUEST_CHANGES

Fresh independent read-only review at exact head `9f0e4650199a49eac2fed9e219c70897fb621f4e`, as part of the MCP release scope (with #778 and release delta `efbb09e9..0d8b9643`).

### What this PR gets right

The no-selector `deployment_status` defect from #777 is genuinely repaired: the binding reads `/api/runtime/deployments?limit=25` through Viewer control and no longer depends on the MCP process's missing runtime environment (`src/lib/mcp/bindings.ts:1116-1144`).

A genuinely disabled plane also stays truthful — only an explicit `LLV_RUNTIME_EVENTS` rollback yields "runtime events are disabled", while missing-socket and configured-but-unreachable keep distinct responses (`src/lib/runtime/flags.ts:57-65`, `src/app/api/runtime/deployments/route.ts:24-47`). That was checked specifically, because a fix that merely masked a real disabled plane would be worse than the bug.

### HIGH — blocking: valid failed-deployment status objects become MCP refusals

Evidence: `src/lib/mcp/bindings.ts:90-97`, `src/app/api/runtime/deployments/[deploymentId]/route.ts:25-30`, `src/lib/runtime/contracts.ts:467-478`, `src/lib/mcp/deploymentReads.test.ts:141-170`.

`getViewerControl()` treats **any** truthy top-level `error` as a control-plane failure:

```ts
if (result.error || !response.ok) { throw new McpToolRefusal(...) }
```

But `ViewerDeploymentStatus.error` is a deliberate `string | null` domain field. A failed or rolled-back deployment is returned as **HTTP 200 with a non-null `error`**, so it is misclassified as a transport refusal. Net effect: the status of a failed deployment becomes unreadable over MCP — precisely the case an operator most needs to read. This is the same class of misreporting #777 exists to fix, inverted.

- **Fix intent:** discriminate control-plane failure by HTTP status or an explicit error-envelope discriminator; keep `error` as deployment-domain data on successful item reads.
- **Constraints:** retain the current distinctions among explicit rollback, missing socket, configured-plane transport failure, 404, and successful empty ledger.
- **Acceptance:** a production-adapter test serves HTTP 200 with `phase: "failed"` and a non-null `error`, and `deployment_status(deploymentId)` returns the structured deployment. Existing 404/503 behavior and all current read-path tests stay green.

### Gates run

`git diff --check` on all three ranges passed. Isolated `bunx tsc --noEmit` at this head (which includes the release delta) passed. Four focused read-path test files: 37 passed. Not run: build, lint, broad suites, live MCP calls, deployment checks.

### Release status

This PR is **not** merged and production stays at `efbb09e94cc30e0f8606493baa843abad9eceea9`. No deploy has been prepared or requested; the merge bar needs an APPROVE plus tsc/tests evidence, and CI here currently carries only the two privacy gates.

Note: this PR carries no `Closes #` reference, so merging as-is would leave #777 open — the same way #775 merged and left #765 open.

Issue #798, opened 2026-07-30
perf: /api/files projects the 22 MB registry twice per request and never warms, stalling the session catalog


Latand:
The session catalog stalls: the operator hit five failed load attempts, and `/api/files` still costs 4–8 s per request once the API is responsive again. The cost is concentrated in three stages, and two of them are the same work done twice.

## Measured on the deployed revision

`Server-Timing` from a live request (7.66 s total):

| stage | ms |
|---|---|
| `files-registry` | 2576.8 |
| `files-flows` | 2261.1 |
| `files-session-titles` | 2178.5 |
| `files-json` | 129.3 |
| `files-role-titles` | 66.1 |
| everything else combined | < 30 |

Three stages account for ~7.0 s of 7.66 s. An earlier request in the same window measured 2322.2 / 2135.1 / 2052.5 for the same three, so this is stable, not a spike.

Response headers on every request:

```
x-llv-files-cache: stale
x-llv-files-generation: 0
x-llv-files-target-generation: 1
x-llv-files-cache-requests: 120
```

The generation never reaches its target, so no request is ever served warm — 120 requests in, still generation 0.

## State size behind it

- `state/agent-registry.json` — 22,576,436 bytes: 4,631 conversations, 2,182 entries, 2,738 receipts, 936 membership keys, 2,590 lineage edges, 3,728 held deliveries
- `state/flows.json` — 1,525,795 bytes
- persisted snapshot: schema 9, 433 files, 29 projects, 583,660 bytes
- roots: ~6,290 Codex and ~968 Claude JSONL files

A staging instance serving the same endpoint stays at ~0.38 s, so this is state-volume dependent rather than a pure code regression.

## Root cause, measured

In `buildFilesResponse`:

1. `registry.readOnlySnapshot()` plus `conversationLookupFromSnapshot` / `conversationForPath` **deep-clone per entry** on a read-only lookup. That is `files-registry` (~2.3 s).
2. `overlaySessionTitles(files)` then builds `sessionTitleProjector()`, which calls `registryProjection(agentRegistry())` **again** — a second full registry read and projection (~2.1 s) over the same 22 MB, while the snapshot and projection from step 1 are already in hand and unused.
3. `files-flows` (~2.2 s) spans from before titles through pipeline reconciliation and role-title projection.

So the dominant cost is one registry parsed and projected twice per request, plus per-entry cloning on a path that never mutates.

## Contention during cold start

[...]

PR #799, opened 2026-07-30
fix(files): prevent production catalog starvation

Latand:


Latand:
Superseded: issue #798 was closed as completed on 2026-07-31 by #815 (serve cache while projection is active) and #816 (bound projection waits behind cache); this branch conflicts with main and re-implements the same single-projection idea against a base from before those merges. Closing without merge.

PR #940, opened 2026-08-06
test(runtime): assert the Claude structured-host rollback contract (#939)

### LINKED ISSUE #939: fix(test): stale opt-in expectation in claudeStreamBrokerHost.test.ts fails on main
## Failure

`bun test src/lib/runtime/claudeStreamBrokerHost.test.ts` fails on main (verified at 78073ad2 via a pristine `git archive` extract):

```
(fail) ClaudeStreamBrokerHost > requires the exact structured-host opt-in before start
Received promise that resolved
  at src/lib/runtime/claudeStreamBrokerHost.test.ts:1395
```

## Mechanism

Commit 656e60de (`fix(runtime): make the rollback switches actually roll back`) changed `structuredHostsEnabled` in `src/lib/runtime/flags.ts` to opt-out semantics: hosting is on unless `LLV_STRUCTURED_HOSTS` is an explicit rollback value (`0`/`false`/`off`/`no`). The Codex twin test was updated in the same commit ("keeps structured hosting enabled unless the rollback switch is explicit", `codexAppServerHost.test.ts`), but the Claude test "requires the exact structured-host opt-in before start" still asserts the old opt-in contract — it expects `LLV_STRUCTURED_HOSTS: "true"` to be rejected with "structured hosts are disabled". Under the new semantics `"true"` enables hosting, `startClaudeStructuredHost` resolves, and the `.rejects` assertion fails deterministically.

## Fix

Rewrite the Claude test to mirror the updated Codex one: assert rollback values reject and non-rollback values (including `"true"`) start the host. Rename it accordingly.

Found during the fresh review of PR #938 (whose diff does not touch this test or `flags.ts`; the failure reproduces at the merge base).

Latand:
Closes #939.

`bun test src/lib/runtime/claudeStreamBrokerHost.test.ts` failed deterministically on main. The case "requires the exact structured-host opt-in before start" still asserted the retired opt-in gate — `LLV_STRUCTURED_HOSTS=true` refused with "structured hosts are disabled", only a literal `"1"` allowed to start. `structuredHostsEnabled` in `src/lib/runtime/flags.ts` is opt-out: hosting is on unless the value is an explicit rollback spelling, so `"true"` starts the host and the `.rejects` assertion failed. The Codex twin was updated when the contract moved; the Claude one was not.

## What changed

Only `src/lib/runtime/claudeStreamBrokerHost.test.ts`. `flags.ts` and every production source file are untouched — the runtime contract is right, the expectation was stale.

The case is renamed to "keeps structured hosting enabled unless the rollback switch is explicit", mirroring `codexAppServerHost.test.ts`, and asserts the real contract in both directions:

- every rollback spelling the shared reader accepts (`0`, `false`, `off`, `no`) makes `structuredHostsEnabled` false, refuses the start with "structured hosts are disabled", and leaves the child unspawned;
- unset, `"true"` and `"1"` all report enabled and start a host bound to the spawned session, then release.

Each start builds a fresh fake child so the runs cannot leak stream state into each other.

## Verification

Run with `HOME`, `XDG_CONFIG_HOME`, `LLV_STATE_DIR` and `TMPDIR` pointed at a throwaway temp directory — no whole-directory sweeps, no process sweeps.

| check | result |
| --- | --- |
| `bun test src/lib/runtime/claudeStreamBrokerHost.test.ts` before | 44 pass, **1 fail** — `requires the exact structured-host opt-in before start` |
| same command after | **45 pass, 0 fail**, 215 expect() calls |
| `tsc --noEmit` | clean |
| `eslint src/lib/runtime/claudeStreamBrokerHost.test.ts` | clean |
| privacy publication gate (with `--check-commits`) | PASS |

The RED reading came from running the pre-fix file as it stood in the checkout and from `git show` for the historical contract; nothing was stashed.

The rewritten case is not vacuous against the old contract either: the pre-`656e60de` predicate was `env.LLV_STRUCTURED_HOSTS !== "0"`, under which `false`/`off`/`no` would have started a host and the new rollback assertions would fail.


Latand:
Orchestrator merge note: fresh reviewer (pipeline e74354f9) found no code-level defect — diff confined to the one test file, expectations match current main's opt-out contract (flags.ts / registry.ts), no merge conflicts vs main, privacy gate pass. Its two gate failures were environmental (read-only reviewer sandbox, see #1052). Orchestrator re-ran gates in a writable isolated worktree: `bun test src/lib/runtime/claudeStreamBrokerHost.test.ts` → 45 pass / 0 fail; `bunx tsc --noEmit` → only pre-existing PdfPane/pdfjs-dist resolution errors unrelated to this test-only diff. Merging.

Issue #981, opened 2026-08-13
refactor(orchestrator): seats-only resolution — drop the legacy record entirely


Latand:
Part of PRD #976 (slice E, follow-up to slice D). Reviews validate against the PRD's operator requirements.

## Requirement anchor

PRD #976 decision 6: the legacy mode is removed entirely. After slice D the legacy record is write-only compatibility; this slice deletes it.

## Mechanism

Migrate the three remaining readers of the legacy single-instance record to the per-project seat layer, then delete the record:

- `src/lib/orchestrator/authority.ts:111` (`legacyManagerConversationId` fallback in `authorizedManagerSeats`) — resolve manager authority from active seats + revocations only. This is a pure source-removal: seats already supersede the legacy pointer because `syncLegacyRecord` mirrors the newest seat. No behavioural restriction on the orchestrator is added anywhere.
- `src/lib/runtime/realtimeInjection.ts:164` — resolve the manager conversation from the seat of the relevant project instead of the global record.
- `src/lib/mcp/bindings.ts:1367` (`get_orchestrator` engine/model boot-window fallback) — resolve from the registry/seat data only.
- Then delete `src/lib/orchestrator/store.ts`, `syncLegacyRecord` in `seatCommand.ts`, and stop writing `state/orchestrator.json` (leave existing files on disk untouched — just never read or write them again).

## Acceptance criteria

- Bridge channel scope (`src/lib/bridge/routing.ts`), deploy/manager authority, and voice realtime injection resolve correctly from seats for: an active seat, a rotated seat (predecessor revoked), and no seat.
- A designated-before-seats legacy-only manager no longer resolves (acceptable: the operator re-designates through the new UI; call this out in the PR body).
- `store.ts` is gone; nothing in `src/` references `orchestrator.json`.
- Authority-path tests updated and green; `bun tsc` green.

## Out of scope

Any new authority semantics or restrictions — this is strictly a source-of-truth consolidation.

Latand:
**Scope addendum from slice D (#980): there is a fourth reader, in the browser.**

This issue's mechanism lists three readers of the legacy record — `authority.ts:111`, `realtimeInjection.ts:164`, `bindings.ts:1367`. The #980 lane found a fourth:

[...]

Issue #976, opened 2026-08-13
PRD: per-project Orchestrator chat in the left sidebar — draft create, rotate, mobile pin


Latand:
# PRD: per-project Orchestrator chat in the left sidebar

Design interview completed 2026-08-12 → 2026-08-13 (Architect grill session, orchestrator chat surface). This PRD is the review anchor: **every review round of the child issues validates against the requirements below**, not against any intermediate revision.

## Originating requirement (operator, 2026-08-12, translated from Ukrainian)

> "I'd like to look at how the orchestrator works when we designate one, and then be able to open and create an orchestrator through a simple UI — the same draft conversation with model / provider / account parameter pickers, but! All of it happens in the sidebar — on the left. We have document previews in the sidebar now — I want to put the orchestrator chat there too, and the conversation menu can be made larger, but it should still look pleasant when stretched by height, as it will be on desktop. On mobile it maybe needs to be an almost normal window with a menu transition, or maybe a pin of the orchestrator conversation in the conversation list… and of course a separate one per project."

## Operator decisions (interview, translated from voice answers; unrelated asides elided)

1. **Desktop geometry:** a left dock panel **pushed into the layout** between the project rail and the board — not an overlay. Project rail stays visible; the panel follows the selected project; the document preview sheet stays on the right; both can be open at once.
2. **Panel content:** the real conversation feed — `LogFeed` + `TmuxComposer` mounted in the panel column. No bespoke chat renderer.
3. **Draft create flow:** engine/model/effort/account pickers (extracted from `DraftAgentPane` into a shared module) plus a **textarea prefilled with the default orchestrator prompt** — "when the orchestrator is created, the prompt simply lands in a text field; no separate 'mandate' concept — the person edits it however they want". Confirm calls `POST /api/orchestrator/seat`. **Both engines must be supported** — "the whole capability must be supported", any engine/model/account the draft offers. Cwd = project root, no field.
4. **Rotate / delete:** rotate from the panel via `POST /api/orchestrator/rotate` (same draft, prefilled). No separate delete in v1 — closing the conversation card leaves the seat `exists: false` and the panel returns to draft.
5. **Mobile:** a pinned orchestrator row always first in the project conversation list; tap opens the standard `MobileFocusView`; with no orchestrator the row reads "Create orchestrator" and opens the draft as a fullscreen sheet.
[...]

PR #985, opened 2026-08-13
chore(orchestrator): retire the legacy global entry — Overview button + the designation POST (#980)

### LINKED ISSUE #980: chore(orchestrator): retire the legacy global entry — Overview button + /api/orchestrator routes
Part of PRD #976 (slice D). Reviews validate against the PRD's operator requirements.

## Requirement anchor

PRD #976 decision 6, operator verbatim (translated): *"That legacy mode should be removed entirely, or modified."* The legacy global-orchestrator entry competes with the per-project seat layer: hardcoded config, no model/account choice, a separate record, adoption semantics of its own.

## Mechanism

- Remove `OrchestratorChatButton` from the Overview header (`src/components/OverviewBoard.tsx:118`, `src/components/OrchestratorChatButton.tsx`) and the client resolve-or-spawn flow (`src/components/orchestratorChat.ts`).
- Remove the `GET/POST /api/orchestrator` routes (`src/app/api/orchestrator/route.ts`) — their only caller is the removed button.
- KEEP `src/lib/orchestrator/store.ts` and the `orchestrator.json` record for now, **write-only** via `syncLegacyRecord` (`src/lib/orchestrator/seatCommand.ts:187`): bridge authority (`src/lib/orchestrator/authority.ts:111`), voice realtime injection (`src/lib/runtime/realtimeInjection.ts:164`) and the `get_orchestrator` boot-window fallback (`src/lib/mcp/bindings.ts:1367`) still read it. Full removal is slice E.
- Delete tests that exist only for the removed flow; keep seat-layer tests untouched.

## Acceptance criteria

- No UI path spawns a global orchestrator; the only creation paths are the seat draft (slices A/C) and MCP `create_orchestrator`.
- `/api/orchestrator` and `/api/orchestrator` POST return 404 (routes gone); `/api/orchestrator/seat` and `/api/orchestrator/rotate` untouched.
- Bridge scope resolution and voice injection behave exactly as before (the record keeps being written by seat activation).
- `bun tsc` and the touched test files green.

## Out of scope

Removing `store.ts`/`orchestrator.json` and migrating its readers — slice E.

Latand:
Slice D of PRD #976. Closes #980.

## What the operator asked for

PRD #976 decision 6, verbatim (translated): *"That legacy mode should be removed entirely, or modified."* The legacy global-orchestrator entry competed with the per-project seat layer — hardcoded config, no model/account choice, a record of its own, its own adoption semantics.

## Removed

- `OrchestratorChatButton` and its mount in the Overview header (`OverviewBoard.tsx`).
- The client resolve-or-spawn flow `src/components/orchestratorChat.ts`.
- Both of their test files (`OrchestratorChatButton.dom.test.tsx`, `orchestratorChat.test.ts`) — they existed only for this flow.
- The four `orch.*` strings that existed only for the button, from `en.ts` and `uk.ts`. No other key was touched.
- `POST /api/orchestrator` — the designation/adoption half of that flow.

No UI path spawns a global orchestrator any more. The creation paths left are the seat draft (`/api/orchestrator/seat`, slices A/C) and MCP `create_orchestrator`.

## One deliberate deviation: `GET /api/orchestrator` stays

The issue records the removed button as the routes' only caller. A repo-wide grep says otherwise:

`src/components/voice/managerIdentity.ts:51` fetches `GET /api/orchestrator` to answer "which conversation is the designated manager", and `TmuxComposer.tsx:1741` uses that answer to scope the viewer-context prelude to the manager alone. That module was written precisely to stop the prelude leaking into unrelated workers' turns, and it **fails closed** — so deleting the GET would raise no error anywhere. It would silently stop prepending the operator's view for every conversation, with nothing to notice it by.

Retiring it belongs with that reader's migration, which is slice E (#981) — whose inventory also lists only the three server-side readers and likewise misses this browser one. Worth folding into #981.

Consequence: `POST /api/orchestrator` answers **405**, not the 404 the acceptance criteria name. 404 requires the route file to be gone entirely; Next 16 returns `405 Method Not Allowed` for an unexported method on a route that still exists (`node_modules/next/dist/docs/01-app/01-getting-started/15-route-handlers.md:43`). The entry is gone either way — nothing can designate a global orchestrator over HTTP.

Verified against a real production build:

| request | before | after |
| --- | --- | --- |
| `POST /api/orchestrator` | 200, designates | **405** |
| `GET /api/orchestrator` | 200 | 200 (unchanged) |
| `POST /api/orchestrator/seat` | 400 (validation) | 400 (untouched) |
| `POST /api/orchestrator/rotate` | 400 (validation) | 400 (untouched) |

## Kept, as the issue requires

`src/lib/orchestrator/store.ts` and the `orchestrator.json` record stay write-only via `syncLegacyRecord`. Its three readers — bridge/manager authority, voice realtime injection, and the `get_orchestrator` boot-window fallback — are untouched, and their tests were neither modified nor skipped. Because seat activation keeps writing the record, bridge scope resolution and voice injection behave exactly as before.

## Incidental fix

`src/app/servedPayloadSecrets.test.ts` started its server with `bunx next start`, which runs it under node, where the instrumentation hook dies on "SQLite state stores require the Bun runtime" — every request 500s and six assertions read an error page instead of a served payload. CI does not run this file, so it failed silently. It now spawns the command `package.json:31` starts the viewer with, and the file is green. Its one-click designation test became an assertion that the retired entry is gone.

## Verification

All runs used an isolated `HOME` / `XDG_CONFIG_HOME` / `LLV_STATE_DIR` / `TMPDIR`; nothing touched the shared state directory.

- 120 focused tests green across 15 files: orchestrator route, seat route, `seatCommand`, `store`, orchestrator authority, `OverviewBoard` (render, catalog, targets), `TmuxComposer` viewer-prelude and selected-context, `VoiceComposerHost`, realtime injection, i18n, bridge recovery, and `servedPayloadSecrets` against the real build.
- `bunx tsc --noEmit` clean.
- `eslint` on every touched file clean.
- Production build succeeds.
- Privacy publication gate: PASS.

Not deployed, per the PRD's delivery policy for this batch.


Latand:
Orchestrator decision on the parked question in this lane.

**Retain `GET /api/orchestrator` for now; `POST` stays deleted. Merge as-is.**

The builder verified the issue's claim that the removed Overview button was the route's only caller, and it does not hold: `src/components/voice/managerIdentity.ts:51` calls the GET to answer "which conversation is the designated manager", feeding the manager-scoped viewer-context prelude at `TmuxComposer.tsx:1741`. That module fails closed, so deleting the GET would raise nothing anywhere — it would silently stop prepending the operator's view for every conversation until slice E lands.

The operator's requirement (PRD #976 decision 6) is that the legacy *mode* is "removed entirely, or modified". POST is the creation entry, and it is gone: no HTTP path designates a global orchestrator any more. The remaining GET is a read that slice E migrates to the seat layer along with the other readers. Trading a silent regression for a few hours of route-file symmetry is a bad trade.

Two consequences recorded deliberately:

- `POST /api/orchestrator` answers **405**, not the 404 named in this issue's acceptance criteria. 404 requires deleting the route file, which is the decision above. Next 16 returns 405 for an unexported method on a surviving route (verified against a production build). The entry is closed either way, so the acceptance criterion is met in substance and amended in letter.
- The `managerIdentity` browser reader is folded into #981, which previously listed only the three server-side readers.

The out-of-scope `servedPayloadSecrets.test.ts` fix (it spawned the server under node, where the instrumentation hook dies on the Bun-runtime requirement, so six assertions were quietly reading a 500 page) is accepted and disclosed rather than reverted — it was failing silently and CI does not run that file.

Stale docs (`docs/design/orchestrator-agent.md:32-37`, `docs/design/conversation-monitor.md:27`) stay for slice E, which rewrites that area.

PR #916, opened 2026-08-05
feat(identity): enforce semantic spawn titles and migrate legacy identity

Latand:
## Summary

- require semantic titles at Viewer-controlled spawn writers and central registry admission
- preserve one durable semantic title across replay, settlement, resume successors, generations, entries, JSON, and SQLite
- persist browser draft titles through reload recovery, including image-only launches and stable pre-title attempt upgrades
- require and publish `title` in the MCP `spawn_agent` schema and bindings
- run one guarded Identity Wave A+D migration at startup with dry-run support, retryable evidence reads, durable counters, and JSON/SQLite parity
- retitle placeholder generations, entries, and receipts from clean receipt/transcript evidence; rekey stale shared-store paths with legacy continuity; stamp cycle-safe orchestrator lineage

Closes #913

## Mechanical cleanup

Rebased onto `origin/main` at `d099f050c692be963497a3a9fcf8e43c0901a02f` and preserved `main`'s packaged-Bun WakaTime worker executable selection from #1157 in `src/lib/viewerInstrumentation.ts`.

Dropped four unrelated curation/files commits from this branch:

- `66a15a49` `fix(viewer): stabilize dashboard files without render refs`
- `f09bbfe7` `fix(projects): serialize crown mutations`
- `238da9a6` `fix(projects): fence crown poll acknowledgements`
- `8cba8b55` `fix(projects): acknowledge crown poll generations`

`d7fa73b5` remains because its file-route changes repair semantic-title fixtures required by this PR. The curation polling and crown-mutation acceptance item belongs to a separate change and has been removed from this PR's scope.

No merge or deploy was performed.

## Acceptance criteria

- [x] Every specified Viewer-controlled spawn writer passes a semantic, non-placeholder title: HTTP/UI/MCP proxy, pipeline stage, flow review round, task spawn, structured recovery, account-migration successor, transcript-host resume, and legacy workflow stage.
- [x] Legacy workflow stages stamp `<workflow task> · <stage>`; long workflow and pipeline task titles retain the stage suffix.
- [x] Browser drafts persist and submit titles through reload recovery, including empty/image-only validation and stable pre-title attempt upgrade.
- [x] The per-project built-in orchestrator request carries a semantic title through real spawn-route admission.
- [x] MCP `spawn_agent` schema fixtures and bindings require and publish semantic titles.
- [x] Fresh central registry admission refuses missing, generic, punctuation-only, and Markdown-separated placeholder titles before receipt creation.
[...]

Issue #473, opened 2026-07-20
feat: add WakaTime activity integration


Latand:
## Goal

Add a clean WakaTime integration for Agent Log Viewer so agent work observed by the viewer is represented accurately in WakaTime.

## Pipeline

1. Research official WakaTime ingestion and plugin conventions plus the viewer lifecycle and activity data available on current origin/main.
2. Produce an implementation-ready architecture with explicit ownership, privacy, idempotency, throttling, failure, and configuration boundaries.
3. Implement the selected design from a clean worktree pinned to current origin/main.
4. Run focused tests, type checks, and at most two independent review rounds.
5. Publish a non-draft PR with evidence and any remaining limitations.

## Acceptance criteria

- Activity maps to stable WakaTime project, entity, language/category, and time semantics.
- The integration is opt-in and keeps API credentials out of logs, transcripts, repository state, and browser payloads.
- Duplicate events, restarts, idle periods, delivery failures, and rate limits have defined behavior.
- Core viewer scanning and agent execution stay responsive when WakaTime is unavailable.
- Unit or integration tests cover the behavioral contract.
- Operator documentation explains setup, disablement, and verification.
- Relevant checks pass on the PR head.

## Delivery contract

The result is a reviewable non-draft PR. Review is capped at two rounds for this task.

PR #1017, opened 2026-08-15
fix(wakatime): preserve agent time and count direct operator activity (#763)

### LINKED ISSUE #763: feat(worktime): produce deterministic operator-supervised #zvit drafts
## Problem

The current WakaTime integration from #473 records every visible Claude/Codex turn as an `ai coding` entity. Active sampling can continue during silent tool execution. A downstream daily report runner that reads WakaTime plus manual logs therefore has two symmetric failure modes:

- supervised operator work can be missed when no editor heartbeat exists;
- unattended agent execution can be counted as operator work.

The scheduled runner also depends on model availability, has no deterministic catch-up pass, and can mark a report as sent before delivery is acknowledged.

## Ownership and boundary

Viewer should own a privacy-minimized operator-engagement ledger because it already owns account-root discovery, session parsing, fan-out/resume lineage, project context, and WakaTime export. A server-side scheduler can consume daily rollups without receiving transcript text.

The existing `agent-log-viewer/...` WakaTime entities remain useful for Viewer activity analytics. Worktime calculation must classify them as synthetic agent activity and exclude them from editor evidence.

### 1. Canonical operator events

- Tag direct operator input at ingestion with a stable `operatorEventId` and origin such as composer, realtime input, or API human input.
- Preserve the source ID through resume, fork, fan-out, forwarding, bridge transport, and account-root mirrors.
- Prefer native prompt/turn/message IDs. For historical events without a stable ID, use a documented fallback derived from normalized content, lineage, and a bounded timestamp window.
- Collapse copied/resumed/fan-out forms to the earliest canonical operator event.
- Exclude system messages, scheduler prompts, injected spawn instructions, bridge envelopes, tool results, agent-authored follow-ups, and background execution.

### 2. Evidence and attribution

- Scan every configured Claude/Codex account root and Viewer project generation.
- Ingest real editor intervals from WakaTime while excluding Viewer-generated `agent-log-viewer/...` entities and boundary projects.
- Attribute an event only when repository, cwd, explicit task context, or durable project metadata supports the project.
- Preserve ambiguous events in a separate review bucket. Never apply a default paid project.
- Store evidence digests, event IDs, source class, project, and episode boundaries. Do not store transcript text in rollups.

### 3. Canonical daily calculation

For each project and Europe/Kyiv day:

1. Sort deduplicated operator events.
2. Keep events in one supervised episode when adjacent gaps are at most 30 minutes.
3. Break the episode after a longer unattended gap.
4. Use `max(lastEvent - firstEvent, 10 minutes)` for each episode.
5. Union supervised episodes with real editor intervals.
6. Count cross-project overlap once, using the configured project-priority rule.
7. Round the daily project total to the nearest 0.5 hour; keep a 0.5-hour minimum for a valid deliverable episode.

Expose raw minutes, rounded hours, evidence counts, excluded synthetic duration, deduplication counts, and ambiguous duration for audit.

### 4. Deterministic delivery contract

- Export a model-free daily rollup through a CLI or authenticated local endpoint.
- Run a daily idempotent catch-up scan for every unprocessed complete day, including days after quota or host failures.
- Separate lifecycle fields: `calculated_at`, `drafted_at`, `delivery_attempted_at`, `delivered_at`, destination, receipt ID, and last error.
- Treat delivery as complete only after the transport returns a receipt and read-back confirms the destination/message.
- Default to a private draft. Group publication requires a separate explicit approval path.

## Acceptance tests

- [ ] One operator prompt copied into a parent, worker, resume, and mirrored account produces one canonical event.
- [ ] A forwarded worker prompt inherits the original operator event and adds zero minutes.
- [ ] A 45-minute silent tool call after the final operator event adds zero supervised time.
- [ ] Operator events 29 minutes apart form one episode; events 31 minutes apart form two.
- [ ] A singleton deliverable episode contributes 10 raw minutes and rounds to 0.5 hour.
- [ ] Viewer-generated WakaTime entities contribute zero editor minutes.
- [ ] Real editor intervals union with chat episodes without double-counting.
- [ ] Cross-project overlap is assigned exactly once according to configured priority.
- [ ] Ambiguous evidence appears in the review bucket and contributes zero reportable time.
- [ ] Failure on one scheduled day is backfilled by the next healthy run.
- [ ] A weekly schedule boundary still processes the previous complete day.
- [ ] A failed delivery leaves `delivered_at` empty and remains retryable.
- [ ] Re-running a completed day is idempotent and does not create another message.
- [ ] Rollup storage and logs contain no prompt text, credentials, account identifiers, or chat identifiers.

## Implementation sequence

1. Define the operator-event and rollup schemas, privacy boundary, and historical fallback rules.
2. Add provenance at every human-input ingress and propagate it through lineage.
3. Build the deterministic episode calculator with synthetic WakaTime filtering.
4. Add an auditable dry-run CLI/exporter and historical fixtures.
5. Update the server-side consumer to use catch-up and receipt-backed delivery state.
6. Run a private-draft backfill and compare it with an operator-approved incident fixture.
7. Update `docs/wakatime.md` and the design documentation with the two distinct activity semantics.

## Related

- #473
- #629
- PR #477

Production delivery, source changes, deployment, and service changes are outside this issue-creation step.

Latand:
Closes #763

Implements the superseding owner clarification in https://github.com/Latand/live-log-viewer-next/issues/763#issuecomment-5120580507.

## Summary

- preserve every existing #473 agent and orchestrator WakaTime heartbeat, including unattended execution and silent tool work
- add direct-operator activity at validated legacy, structured runtime, pending-answer, task dispatch, task spawn, direct spawn, and realtime ingress boundaries
- mark Viewer monitor, MCP, orchestrator, bridge, and self-named agent traffic with server-verifiable non-operator provenance while preserving the zero-ceremony local-browser flow
- require the exact `LLV_WAKATIME_ENABLED=1` activation predicate; disabled recording performs no sidecar or registry access and does not affect Viewer delivery
- converge Claude Viewer delivery-ledger message ids with their original server action key while retaining stable terminal-only legacy Claude prompts and excluding SDK, peer, and coordinator traffic
- retain realtime operator identities in a bounded browser outbox until acknowledgement, across transient failures and call teardown, with no speech text stored
- materialize each direct gesture as one fixed 10-minute interval, drain stale sidecars in bounded batches with diagnostics, and keep the existing endpoint, payload, cadence, credentials, retry behavior, and project attribution

## Correctness and privacy

Direct browser gestures receive stable server identities. Retries, resume, fan-out, transcript mirrors, and Claude delivery acknowledgements reuse the same identity. Task fan-out records one gesture before delivery; direct and task spawns record before actuation, so launch failure keeps the activity event. Agent capabilities and authenticated internal-service lanes produce zero direct-operator events.

The durable Viewer sidecar stores only an opaque hash, engine, project, and server timestamp. The realtime browser outbox stores only random 64-character hexadecimal identities. Prompt text, speech text, answers, labels, transcript paths, account data, and credentials are absent from these stores.

## RED evidence

- exact activation and zero-state recording: disabled cases reached registry/state before the shared predicate existed
- internal-producer provenance: monitor/MCP calls without a conversation capability were classified as operator traffic before signed service provenance
- task dispatch and spawn capture: focused route tests failed before stable gesture recording was wired
- Claude convergence and stale sidecars: focused run had 1 pass and 2 failures before transcript identity mapping and expired interval materialization
- realtime durability: focused browser/server run had 1 pass and 5 failures before the persistent hash-only outbox and dedicated live-peer acknowledgement action
- bounded stale draining: focused run had 0 passes and 1 failure before the 256-action per-tick limit and backlog diagnostic

## GREEN evidence

Exact head: `f60b2e4b31f2a0c0e78163bfc169abed35aeaa57`

- isolated focused/integration suite: 364 passed, 0 failed, 1,318 assertions across 20 files
- explicit TypeScript check: `bun --bun node_modules/.bin/tsc --noEmit` passed
- changed-file ESLint check: passed with zero warnings
- production build: passed, including Next static generation and MCP bundle generation
- generic publication privacy gate against `origin/main`: passed
- commit-aware publication privacy gate against `origin/main`: passed
- commit whitespace check: passed

No merge, deployment, service restart, production backfill, port access, or user-data mutation was performed.

Ready for a fresh exact-head full-diff review.

## Final round — one operator event id per bulk submit (#763)

Exact head: `25034f5aa7b6da8faef948076b27f2e6de4056b6`

`origin/main` was merged into this branch first (merge commit `a26fa48e`, no rebase, no force-push). Both sides kept: this lane's direct-operator activity gates alongside main's semantic-spawn identity wave (hoisted spawn attempt lookup, `title` now required on every new spawn) and the bridge-directive answer trailer. The one semantic follow-up: the operator-gesture spawn fixture gains the now-required explicit title, since that test is about the gesture ledger and not about title derivation.

### The P1

A bulk submit is one operator input, but the bar fans it out over one structured task-send request per selected target. Each request minted its own operator gesture, so the recorder saw two unrelated gestures for a two-target submit and queued **two** operator heartbeats for one input.

The bar now mints a single gesture id per submit and carries it on every per-target request. Nothing downstream needed new logic: the send route already derives the recorder's idempotency key from the request's `clientRequestId`, and the heartbeat queue already folds a repeated key into the stream it opened. The failed-slot retry replays the same runner, so it stays on the same gesture rather than counting the operator again; a separate submit is a separate input and gets its own heartbeat.

**Legacy tmux left untouched, deliberately.** With the «create task» toggle off, the bar still broadcasts through `POST /api/tmux`. That path remains reachable and is unchanged — the tmux transport is being removed (#1161) and must not gain new logic.

### Scope

- `src/components/scheme/BulkActionBar.tsx` — mints the gesture id per submit
- `src/components/tasks/taskApi.ts` — `sendTask` carries it on the wire
- `src/components/scheme/BulkActionBar.operatorGesture.dom.test.tsx` — new regression test

No other product change. The send route, the recorder, and the heartbeat queue are untouched.

### RED evidence

The new test mounts the real bar with two structured targets and drives the operator's own send gesture; each per-target request runs through the real send route, the real operator recorder, and the real heartbeat queue writing to an isolated state file. Against the pre-fix sources (restored from the merge commit, both files reverted in place):

```
expect(pending).toHaveLength(1)
  Expected length: 1
  Received length: 2

expect(pendingHeartbeats()).toHaveLength(2)   # two submits
  Expected length: 2
  Received length: 4

0 pass, 2 fail
```

With the fix in place both are 1 and 2 respectively, and every request carries the same gesture id whose digest keys the single queued heartbeat.

### GREEN evidence

- `bunx tsc --noEmit --incremental false` — clean
- touched tests by path under an isolated `HOME`/`XDG_CONFIG_HOME`/`LLV_STATE_DIR`/`TMPDIR`: **383 passed, 0 failed, 1,658 assertions across 20 files** (new bulk-bar test, `bulkActions`, task send/spawn routes, spawn route, answer, tmux, operator authority, MCP bindings and bridge directive, monitor viewer API, orchestrator seat and seats, realtime selected context, runtime http, turn duration, all three WakaTime suites)
- `bun scripts/privacy-publication-gate.ts --base $(git merge-base HEAD origin/main) --check-commits` — PASS
- `eslint` — 17 problems (13 errors, 4 warnings), byte-identical to the count on a clean `origin/main` checkout; none in the files touched here

### Pre-existing failure carried in from main

`src/lib/telegram/reportSpawn.test.ts` → "the report class decides the whole capability surface admission reserves" fails with `no launch reserved: 400 {"error":"title is required for every new spawn"}`. It fails identically on a clean `origin/main` checkout, so it arrived with the #916 identity wave and is out of this lane's scope. Filed as #1198.


Latand:
Orchestrator note (2026-08-25): parked after a rebase and five fresh review rounds today. The ledger/outbox over-build, the per-action heartbeat fan-out, the id-less spawn gap, the project_unresolved fallback and the stale docs were all fixed and re-verified; the last round (at the current head) still holds one P1:

- P1 — `src/components/scheme/BulkActionBar.tsx:205` — one bulk operator submit is split into per-target task-send or legacy tmux requests without a shared gesture identity, so a two-target submit records two operator heartbeats (violates "exactly one operator heartbeat per operator input").

Given the operator's decision that the tmux path is being removed, the right fix is a single operator event id minted by the bulk bar and carried by every per-target request, applied only on the structured path. Leaving this for a short follow-up lane rather than a sixth round here.

Issue #611, opened 2026-07-23
Same-account continuation stalls queued when a builder is parked at a quota-warning/prompt during review relay (forces manual preserve-commit + pipeline restart)


Latand:
## Pattern (recurring, observed 3× on 2026-07-22/23 during #609/#610 review cycles)

When a structured builder is mid-review-relay and its Codex account hits a **quota-warning prompt** (or a tmux fixture leaves the pane at a prompt), the **same-account continuation stays `queued` and never delivers**. The auto-relay of reviewer findings into the builder conversation stalls. To avoid reproducing #565, the orchestrator has had to repeatedly: preserve the builder's work as a recovery commit, then start a **fresh structured pipeline from that exact base** and re-attach the reviewer.

Concrete instances tonight:
- #610 repair relay stuck on a "Codex quota-warning prompt detector" + same-account continuation queue → worked around via separate Sol pipeline `0a936174`.
- #610 builder stuck again after a tmux fixture; same-account continuation stayed queued → preserved as commit `d80c6674`, restarted as Sol pipeline `59a7e1ab`.

## Why it matters

Each occurrence costs a preserve-commit + pipeline restart + reviewer re-attach — pure orchestration overhead, and it is a close cousin of the delivery-stall classes #344/#565 that the operator was most upset about. The recovery is manual/heuristic; it should be automatic or prevented.

## Ask

- Detect the quota-warning / prompt-parked builder state and either (a) auto-advance/dismiss the benign quota-warning prompt so the continuation delivers, or (b) auto-reseat the continuation to a healthy account instead of leaving it `queued` indefinitely.
- A same-account continuation that cannot deliver within a bound must surface as a clear blocked state with a one-click recover, not a silent `queued`.
- Regression: a builder parked at a quota-warning prompt receives a queued continuation → it is delivered (or reseated) automatically without a manual preserve-commit + pipeline restart.

Related: #344 (continuation ignores selected account), #565 (spawn to non-active account deadlock), #574 (kill queued forever).

Latand:
Fresh reproduction (2026-08-15, ~14:19–15:05), Claude flavor of this class:

- A Claude Opus builder hit its session limit mid-lane ("resets 2:50pm"), with its review-flow in `fixing` and the fix commit already pushed.
- Two `send_message` deliveries queued against it (14:22 and 14:56, after the reset) and **never drained** — the transcript never moved again.
[...]

PR #1029, opened 2026-08-18
fix(runtime): execute pending account switches at turn end, force on next send (#1028)

### LINKED ISSUE #1028: fix(runtime): a pending account switch survives turn completion — banner stuck, queued sends stranded
## Symptom (operator screenshot, 2026-08-18 ~17:30)

A structured Claude builder shows the banner "Account switch pending — after current turn". The turn then COMPLETES ("Worked for 26s", card idle) — and the switch still does not execute: the banner stays, the composer shows "switch pending", and the operator's next message sits "queued" indefinitely. The operator expects the switch to be effectively instant once the turn ends.

## Expected semantics (operator-stated)

1. A pending account switch executes AT TURN END — the moment the current turn finishes, not "sometime later".
2. At the latest, the NEXT SEND forces it: tear down the old session/host and start the successor via the structured JSON-RPC path on the target account, delivering the queued message as its first input. In the operator's words: «гасимо стару сесію і піднімаємо нову через jsonrpc з новим промптом коли я вже відправлю».
3. A pending switch must never strand queued deliveries — this is the #611 delivery-stuck class wearing an account-switch hat.

## Suspected mechanism

The "after current turn" trigger listens for turn completion but misses it (race between the completion event and the switch scheduler, or the executor only runs on some poll that never fires for an idle host), leaving pendingAction unexecuted; send admission then defers to the pending switch instead of forcing it. Needs tracing in the runtime account-rebind path (structured host lifecycle + send admission + the switch scheduling written when the operator picks a new account mid-turn).

## Acceptance

- Switch requested mid-turn: executes within seconds of turn completion (old host down, successor up on the target account, banner gone).
- Switch pending + operator sends: the send forces the switch — successor host starts on the target account via the structured spawn/resume path and the queued message is its first delivery. No manual resume, no indefinite "queued".
- A switch that cannot execute (target account dead/limit) surfaces an actionable error instead of silent pending.
- Repro test: request switch during a turn, complete the turn without further events → assert rebind executes; variant where the completion event is "missed" → next send still forces it.

Refs: #611 (queued deliveries to parked hosts), #926 (pinned-account spawn refusals).

Latand:
Fixes #1028.

## Diagnosis

A pending switch on a structured conversation had **no path at all** to execute. Two gates, both fed by the same broken evidence:

1. **`waiting-turn` never releases.** `advanceConversationMigration` leaves `waiting-turn` only when `successorCreationReady` sees `conversation.turn` as `terminal`/`idle`. That field is written by the inventory pass from the *authoritative* transcript projection (`turnStateFromRecords(records, codex, authoritative = true)`), which for Claude closes a turn **only** on a `result` record or a terminal API error. The Claude CLI never writes a `result` record into its session JSONL — verified against a real transcript: the record types are `user`/`assistant`/`attachment`/…, and a completed turn ends on an assistant record carrying `stop_reason: "end_turn"`. The forward projection reads every assistant record as busy, so the durable turn is `busy` forever.
2. **The escape hatches are pane-shaped.** `projectedInventoryTurn` releases a busy projection on `pane_at_composer`, a live usage wall, or a dead stalled turn. A structured (pane-less) host produces none of them, so nothing ever releases it. `completeProviderTurnObservation` re-derives from the same bytes at successor-creation time, with the same result.

That is why the card can read idle while the switch stays pending: the card's activity comes from the **non**-authoritative parse (`jsonl_turn_completed`, which does honour `stop_reason: end_turn`), while the coordinator uses the authoritative one. Two projections of the same file, disagreeing.

Compounding it: in the Viewer process the migration reconciler has no timer — it runs only on `requestAccountMigrationTick()`. The inventory-worker process does tick every 60s, but a structured successor spawn there throws `StructuredDeliveryControllerUnavailableError`, which the coordinator swallows as "return unchanged". So even a released turn had no reliable executor. And on the send side `deliveryFence` returned `held`, deferring to a switch that could never run — the #611 stranding class, wearing an account-switch hat.

Reproduced before touching anything: switch requested mid-turn → `waiting-turn`; turn completes with no further events; `advanceConversationMigration` returns `waiting-turn` unchanged. Writing the durable release alone moves it to `requested` and then stalls again on `completeProviderTurnObservation` — both gates confirmed independently.

## Fix

**`src/lib/runtime/structuredTurnRelease.ts` (new)** — `releaseStructuredTurnForPendingSwitch(conversationId)` records the structured host's own terminal event as a durable `terminal`/`lifecycle` turn observation. It writes only while a switch is in `waiting-turn`/`requested`, and only on top of the observation it was derived from (`expectedTurnObservedAt`), so a turn that started again keeps its busy projection.

**`src/lib/accounts/migration/coordinator.ts`** — `structuredLifecycleRelease()`: a lifecycle release recorded at or after the transcript's newest write outranks the transcript derivation, in both the inventory projection and `completeProviderTurnObservation`. Any later write moves the mtime past the release and the transcript takes the projection back, so the release cannot outlive its evidence.

**`src/lib/runtime/structuredDeliveryController.ts`** — turn-end trigger. `publishHostState` already runs on every host lifecycle change and already computes `turn = activeTurnRef ? "running" : "idle"`; an idle host now records the release and wakes the migration controller. Result: the switch executes within a tick of the turn ending, banner gone.

**`src/lib/runtime/structuredMessageDelivery.ts`** — belt and braces. A send arriving while a switch is pending against an **idle** host now runs the switch to its end inside the request, then delivers against the successor session. The message is therefore admitted after the commit — assigned to the successor generation and delivered as its first input — instead of being queued against a predecessor about to be retired. Two deliberate limits: a **running** turn still keeps the banner's promise, and an **unknown** turn axis (recovering, degraded, gone) is not evidence that anything finished, so those sends keep waiting rather than tear down a session that may still be working. If the successor's host is not published yet, the message is admitted durably and drained onto the successor by the coordinator.

**`src/lib/accounts/migration/{provider,safeHistoryCopy}.ts`** — actionable failure. Target-account unavailability now throws a named `MigrationTargetUnavailableError` instead of a bare `Error`, so `sanitizeProviderError` can surface `"target account is not signed in; sign it in or switch to another account"` (code `target-account-unavailable`) rather than collapsing it into the generic preflight message. It lands on the failed ribbon next to Retry / Keep. Internal messages stay redacted as before.

## Verification

`src/lib/runtime/structuredTurnRelease.test.ts` (new, isolated temp registry + temp board file), per the issue's repro list:

- switch requested mid-turn, turn completes with no further events → advance stalls on the transcript alone, executes after the release (successor committed on the target account);
- source turn started again after a release → busy projection returns, switch stays put;
- **send forces it** → switch commits in the request, the command goes to the successor generation, and the reservation settles `delivered` against `successor-native` — nothing left queued, nothing to send again;
- send with an unknown turn axis → conservative hold, no switch attempted;
- target account unavailable → `failed-recoverable` with `errorCode: target-account-unavailable` and the actionable message;
- release refused when no switch is waiting.

```
bun test src/lib/runtime/structuredTurnRelease.test.ts            7 pass 0 fail
bun test src/lib/runtime/structuredMessageDelivery.test.ts
         …accountReseat.test.ts …sqlite.test.ts                  81 pass 0 fail (4 files)
bun test src/lib/accounts/migration/coordinator.test.ts         105 pass 0 fail
bun test src/lib/accounts/migration/provider.test.ts
         safeHistoryCopy.test.ts                                 39 pass 0 fail
bun test src/lib/accounts/migration/controller.test.ts
         controllerSignal/intentLiveness/turnState               37 pass 0 fail
bun test src/lib/accounts/migration/controllerScan.test.ts
         controller.performance/quotaController/quotaPolicy      21 pass 0 fail
bun test src/lib/runtime/structuredDeliveryController.test.ts     1 pass 0 fail
bunx tsc --noEmit                                                 clean
bun scripts/privacy-publication-gate.ts --base origin/main        PASS
```

Suites were run by path only; nothing swept the shared runtime/registry state.

**Pre-existing red, untouched by this branch:** `src/lib/runtime/structuredDelivery.integration.test.ts` fails 2 of 33 — "a migration-held delivery switches from the source host to the published Codex/Claude successor". Verified identical on the unmodified base (stashed the branch, same two failures). They encode the opposite of `commitSuccessor`'s current policy, which terminalizes held deliveries at commit ("send again to authorize a fresh delivery action") since `96007ed5`. This PR deliberately does not relitigate that policy: it removes the need for a held delivery to survive a commit at all by admitting the operator's message *after* the switch lands. Worth its own issue.

Lint reports 12 pre-existing errors in unrelated files (`src/hooks/useRuntime.ts`, `src/lib/reaperRuntime.performance.test.ts`); none in the files this branch touches.


Latand:
## Fix round for the REQUEST_CHANGES verdict — `7443ce4a`

Took the reduction in finding 6: the second executor is gone, and the release is the host's own live state.

**1 (HIGH) — two writers for one transition.** Forcing now *wakes the owner* instead of running a second executor. A send against an idle host with a pending switch kicks the structured drain and waits for it; the queue's reconfigure — the only code that commits AND then releases + terminates the predecessor session key — performs the transition. Only a switch nobody owns (an engine drain, the active-account reseat one line above) is advanced from the send. Belt and braces on the other side: `advanceConversationMigration` now refuses outright when an in-flight reconfigure owns the migration (`reconfigureOwnsMigration`), so a tick, a route retry, or a send cannot commit behind it; the owner names itself with the new `reconfigureOperationId` option, which `structuredReconfigure` passes. The turn-end trigger no longer commits anything: the queue's own drain already re-fires the owning reconfigure when `activeTurnRef` clears, so all that remains there is a coordinator wake for the ownerless switches that otherwise wait on the 60s poll. Covered by `structuredAccountSwitch.test.ts`: the turn-end path asserts `releasedKeys === [sourceKey]` and the source row left `{ status: "dead", structuredHost: null, claimOwner: null }` with `reconfigure.status === "applied"`; the forced-send path asserts the same predecessor teardown plus delivery against the successor; and one test asserts a tick-shaped advance leaves the switch at `waiting-turn` on the source account.

**2 (MEDIUM) — the churned timestamp fence.** Deleted, along with `structuredTurnRelease.ts`, the synthetic `terminal`/`lifecycle` observation, the `expectedTurnObservedAt` CAS, and the `projectedInventoryTurn` special case — so the inventory projection is byte-identical to `main` for every conversation again. Readiness now reads what you pointed at: a registered structured host for the source generation's key whose `activeTurnRef` is null. One strengthening over the review's sketch, forced by six accountReseat tests that went red: a persisted row also carries `activeTurnRef: null` when nothing ever recorded a turn on it, so the row only counts while its engine process is *verifiably* alive (pid plus start identity, the `verifiedProcessAlive` idiom). A restart-orphaned row is not turn-end evidence, and there is a test for exactly that, next to the running-turn and dead-host cases.

**3 (MEDIUM) — the image 409.** `holdDuringRuntimeSynchronization` is out of this path. When the switch commits but the successor has not published a host, the send now falls through to the ordinary durable admission — the reservation lands against the successor generation as `runtime-images`, refs and all — and only the in-request command is skipped, with a migration tick requested so the coordinator drains it. The capability judgement follows: a send the switch owns, or one awaiting the successor's host, is judged optimistically, because the projection in hand describes a session that is not the target. Test: pending switch + idle host + image payload + no successor session → `ok: true, outcome: "held"` with a `runtime-images` reservation on `successor-native`. The predecessor's stale session is also no longer republished or recovery-spawned in that window.

**4 (LOW) — the phase list.** Gone with the module. The send forces on `deliveryFence === "held"`, which covers every in-flight phase, and `advanceConversationMigration` re-checks its own creation fence. Test: a migration left at `successor-starting` with `providerReceipt === null` still advances to committed.

**5 (LOW) — the 60s wait.** A forced switch that reaches a settled fence requests a migration tick before delivering its own message. One caveat I will not paper over: deliveries held *earlier* in the same pending window are terminalized by `commitSuccessor` ("send again to authorize a fresh delivery action") under the policy `96007ed5` established, so the tick drains what is drainable rather than resurrecting them. Changing that policy is the same question as the two pre-existing integration reds below, and still deserves its own issue.

### Verification (per-file, isolated temp state; no directory sweeps)

```
bun test src/lib/runtime/structuredAccountSwitch.test.ts                    12 pass 0 fail
bun test src/lib/runtime/structuredMessageDelivery{,.accountReseat,.sqlite}.test.ts
                                                                            74 pass 0 fail
bun test src/lib/runtime/structuredDeliveryController|Reconfigure|Queue|Recovery.test.ts
                                                                            71 pass 0 fail
bun test src/lib/accounts/migration/coordinator.test.ts                    105 pass 0 fail
bun test src/lib/accounts/migration/{provider,safeHistoryCopy,controller*,intentLiveness,turnState,quota*}.test.ts
                                                                            90 pass 0 fail
bunx tsc --noEmit                                                           clean
bun scripts/privacy-publication-gate.ts --base origin/main                  PASS
lint                                                                        no findings in touched files
```

Two coordinator tests now name the operation they stand in for (`reconfigureOperationId: "rebind-old"` / `"same-target-old"` and their successors) — they simulate the reconfigure executor, and under the new rule an executor has to say so.

`structuredDelivery.integration.test.ts` still fails the same 2 of 33 ("a migration-held delivery switches from the source host to the published Codex/Claude successor"), unchanged from the untouched base.

Latand:
Round-2 verification (orchestrator, after two reviewer-spawn infrastructure failures — #1031): all five round-1 findings closed in the demanded shapes. (1) single-owner: structuredTurnRelease executor deleted entirely; send path kicks the owning drain and advances directly only when no reconfigure owns the switch. (2) the timestamp fence died with that file; the turn-end trigger anchors on the migration phase set. (3) image sends under a forced switch become durable reservations against the successor generation, coordinator-drained — no 409. (4) the stuck-phase refusal is gone. (5) migration tick fires on the force path. Gates: coordinator.test.ts 105/105, structuredAccountSwitch.test.ts 12/12 (isolated LLV_STATE_DIR), tsc clean.

PR #1064, opened 2026-08-20
fix(privacy): authenticate trusted Telegram vendor root

Latand:
Trusted privacy-gate prerequisite for #1063.

- authenticates relative paths, portable Git executable state, byte lengths, and bytes for the complete patched chigwell/telegram-mcp v3.2.22 vendor tree;
- binds protected digest `344c9f0b9ef56c1ac4935a2245b6d33206e03bb22df86c5d9e7638869915ebb2` to the vendor tree carried by #1063 head `51300397103133a0e3499e2f3b1f2ff212b622f2`;
- exempts exactly the third-party `credential` and `home_path` idiom classes; `known_value` remains enforced;
- keeps all authority in the trusted scanner; candidate manifests and checksums cannot update the protected digest.

Coverage:
- fixed independently derived regular and executable digest vectors;
- content, byte-length, executable-bit in both directions, root-path, addition, deletion, rename, simultaneous file-plus-manifest, symlink, FIFO, and unreadable-tree failures;
- exact protected vendor match with every other finding class still armed.

Verification:
- privacy publication gate: 116 tests / 541 assertions passed;
- TypeScript typecheck passed;
- ESLint passed;
- candidate privacy self-scan passed;
- trusted-main scanner passed against this candidate;
- hosted privacy-publication and privacy-tracker-audit passed on exact head `33bdc116667c94ffc768055363f99dddb8738b5e`.

### DIFF
--- a/scripts/privacy-publication-gate.test.ts
+++ b/scripts/privacy-publication-gate.test.ts
@@ -20,7 +20,6 @@ import { auditGithubPublication, shouldFailGithubAudit } from "./privacy-github-
 import {
   commitMessageFindings,
   formatPrivacyReport,
-  sensitiveClasses,
   TRUSTED_TELEGRAM_VENDOR_EXEMPT_FINDING_CLASSES,
   TRUSTED_TELEGRAM_VENDOR_ROOT_DIGEST,
   trustedVendorRootDigest,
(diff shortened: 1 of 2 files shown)


Latand:
REJECTED — Parser machinery is removed because the product requirement needs one 2FA password field.

FIXED — #1063 head `51300397103133a0e3499e2f3b1f2ff212b622f2` now uses an uncontrolled ref-backed password input. Submission reads the live DOM value, clears it before the async request, and phase changes re-enter with an empty field. Fable’s layout and actionable failure state are preserved.

Evidence: 2 focused DOM tests, 9 render tests, 13 i18n tests, TypeScript, ESLint, and both original/new trusted scanner passes for the complete first-party UI file set. #1064 exact head `33bdc116667c94ffc768055363f99dddb8738b5e` contains only the mode-aware trusted vendor-root rule and has green hosted privacy checks.

Issue #1086, opened 2026-08-21
feat(telegram): Daily Reports — scheduled chat report with history, Run now, and the telegram grant for report runs


Latand:
Parent: #714 · Builds on: #1059 (Telegram Connect, shipped) · Deferred there as "Daily Reports and report history".

## Outcome

The Viewer runs a **scheduled Telegram Daily Report** for the connected account and keeps its history. Once a day (operator-set time, Europe/Kyiv) — or on **Run now** — a Viewer-launched, board-visible Codex conversation reads the operator's Telegram through the packaged read-only `telegram` MCP, analyses the window since the last successful report, and writes one short, action-oriented report. The operator reads it in the Viewer (Telegram delivery is a follow-up — the connector has no write tools yet).

This is the operator's stated primary use of the Telegram connector: "walk through my private messages and the groups I name once a day, report on the previous day, pick up tasks I may have missed". The report must **include every important item, briefly but with enough context; no arbitrary item cap** — the operator explicitly rejected the terse, slop-heavy daily digest a separate VPS agent already sends.

## Product contract

**Settings (Telegram panel, reuse existing flyout/sheet primitives — no new navigation):**
- Enable / disable; schedule time (default 10:00 Europe/Kyiv; weekdays/daily).
- Sources: *private dialogs* (all dialogs with human activity in the window; bots and service broadcasts excluded unless urgent) plus an operator-picked list of **groups**, each flagged *full* or *light pass* (light = only mentions of the operator, replies to them, 1–2 notable threads).
- Run now.
- Report history: one row per run (date, window, status: `ok | quiet | failed | account-mismatch`, open report). A failed run shows a human-readable error, never silence.

**Run semantics:**
- Window = since last successful report, capped at 72 h; first run = 24 h. A run that fails does not advance the window; the next run catches up.
- Verify `get_me` against the identity recorded at Connect; on mismatch the run ends `account-mismatch` with no report.
- Private-dialog discovery: `get_chats` / `list_chats` order is **not** recency (pinned/folder order — confirmed on the live connector: a dialog with a voice message 16 h ago was absent from the first page while a dialog last active 6 weeks ago ranked second). Pick active dialogs by `last_message.date`, or enable the connector's incoming feed (`incoming_feed_status` → `~/.local/state/telegram-mcp/incoming_feed.jsonl`) and read the window from it. Do not rely on list order.
[...]

PR #1088, opened 2026-08-21
fix(telegram): connector survives concurrent reads; crashes are recorded and visible (#1087)

### LINKED ISSUE #1087: fix(telegram): connector dies under a few concurrent large reads and the respawn is invisible to the operator
## Symptom

The packaged read-only Telegram connector (`telegram` MCP, loopback streamable-HTTP from #1059) stopped accepting connections (`curl: (7) Failed to connect … Could not connect to server`) while serving three **concurrent** `get_messages` calls with `page_size: 120` against one supergroup. `connector.json` then showed a new pid with ~18 s uptime: the supervisor respawned it, and `connection.json` / `GET /api/telegram` kept reporting `connected` throughout, so nothing surfaced to the operator.

A second, lighter burst (3 concurrent `get_messages` × 30 + 2 `get_chat`) succeeded; the same three 120-message pages fetched **sequentially** succeeded. `vendor-src/mcp_errors.log` is empty and the viewer container log carries no connector line for the window, so the crash reason is unrecorded.

## Why it matters

Any agent that fans out reads (the upcoming Daily Reports run, or an operator asking "summarise these three groups") can take the shared connector down for every session for ~20 s, and the status surface never shows it. The respawn also drops in-flight calls without an error the caller can distinguish from a network blip.

## Expected

- The connector survives a handful of concurrent bounded reads, or the supervisor serializes/queues them (a small per-process concurrency cap in front of the vendored server is enough).
- A crash is recorded: supervisor writes exit code/signal + last stderr lines to a Viewer-owned log, and `connection.json` / the Telegram panel show a transient `restarting` state with the last restart time.
- A health-check row in the panel counts restarts in the last 24 h.

## Repro

1. Connected Telegram, connector healthy.
2. Fire three `tools/call get_messages {chat_id:<large supergroup>, page:N, page_size:120}` concurrently (N = 2,3,4).
3. Observe connection refused on the loopback port, then a new pid in `state/telegram/connector.json`; status API still `connected`.

Latand:
Closes #1087.

## Proven cause: the connector was not killed by concurrency — the Viewer killed it

Three findings, in the order they were established.

**1. Every fresh status read destroyed the connector.** `runHealthCheck` in
`src/lib/telegram/service.ts` called `stopConnector()` unconditionally before the
enrollment-bridge health check (the bridge needs the same `StringSession`, which
the vendored per-session lock hands to exactly one process), then re-spawned via
`ensureConnector`. `GET /api/telegram?fresh=1` is what the Telegram panel issues
every time it is opened. So opening the panel = SIGTERM (2 s grace, then SIGKILL)
→ bridge round trip → respawn → readiness probes: a ~10–25 s window in which the
loopback port is refused and every in-flight `tools/call` is dropped, ending with
`connection.json` written back to `connected`.

Confirmed live on the host, not just read from the source: the running connector's
kernel start time (`/proc/<pid>/stat` field 22 + `btime`) was `13:19:57.07`, and
`connection.json.lastHealthCheckAt` for the same connection was `13:19:57.913`.
The process serving the account had been spawned by the health check less than a
second earlier — exactly the report's "new pid with ~18 s uptime while the status
API still says connected". It happened again later in the same session (a second
new pid), with no reads in flight at all: the trigger is the status refresh, not load.

**2. The vendored server survives the burst from the repro.** Reproduced against
`vendor/telegram-mcp` itself — the real tree, real `mcp` 1.29.0 SDK, real
`bin/telegram-mcp-server.py` — with Telethon replaced by an offline stub so no
real account is touched: 3 concurrent `get_messages` with `page_size: 120`
succeed (2.5 s wall, all three answered), and so do 12 concurrent calls at
`page_size: 500` (244 KB per answer). The process stays up and the port stays
open in both. The HTTP/session layer is not what dies; the MCP session manager
already catches per-session exceptions (`streamable_http_manager.py`), and
`get_messages` catches everything itself, which is also why
`vendor-src/mcp_errors.log` was empty in the report.

**3. Nothing could have recorded a crash even if one had happened.** The
supervisor spawned with `stdio: ["ignore", "ignore", "ignore"]`, so the
connector's stderr went to `/dev/null`, and `mcp_errors.log` only receives
ERROR-level tool exceptions. A dropped in-flight call, reproduced against the
old entrypoint, surfaced to the caller as
`RemoteProtocolError: peer closed connection without sending complete message
body` — indistinguishable from a network blip, as the issue says.

Honest bound: I did not fan out three 120-message reads against the operator's
live account (it would use their real account and could take their connector
down), so a Telethon-level failure under real concurrency is not formally
excluded. Finding 1 alone reproduces every symptom in the report, and the crash
log added here is exactly the evidence that would settle it next time.

## The fix

**Health stops being destructive** (`service.ts`, `connector.ts`). When the
shared connector is up and ours, the health check asks *it*: `get_me` over the
already-verified read-only surface, which only a live, authorized Telethon client
can answer. That proves what the bridge proved, without a teardown. The
destructive bridge path stays as the fallback for exactly the case where it costs
nothing — a connector that cannot answer is not serving anyone either — and it
still classifies `expired`, which `get_me` cannot.

**A connector that is merely busy is never torn down.** The health call has three
outcomes, not two. A handshake that was *refused* (closed socket, bad proof, a
surface that is no longer read-only) is `null` and falls through to the bridge. A
handshake that *succeeded* while the account question ran out of its 15 s budget —
or a handshake that itself ran out of budget, which means the listener was there
and the event loop was saturated — is `"busy"`: the process is alive, ours, and
still serving, and a missed health deadline is no evidence that anything finished.
Nothing is stopped and no durable verdict is rewritten; the last completed check
stands and the next one asks again. Conflating that with "dead" is what turned a
fan-out of large reads into an outage.

**Crashes are recorded, and nothing raw is ever persisted** (`connector.ts`,
`bin/telegram_connector_monitor.py`). The child's stderr reaches
`state/telegram/connector-stderr.log` (0600, Viewer-owned, truncated per spawn so
a tail always belongs to the process that died) **only through the redactor** —
the monitor owns the server child's stderr and filters every line on the way to
the sink, so the identifiers never touch disk in the first place. An exit this
Viewer did not cause appends `{at, pid, exitCode, signal, observed, stderr[]}` to
`state/telegram/connector-crashes.log` (0600 ndjson, last 50 kept, stderr bounded
to 20 lines / 8 KB / 400 chars per line), and a verified restart is counted in
`state/telegram/connector-restarts.json` (timestamps trimmed to 24 h). The tail
copy is redacted again on the way into the crash log, which is idempotent.

That output is not innocent: the
vendored error helper logs the failing call's own arguments
(`Error in get_messages (chat_id=…, query=…)`, `telegram_mcp/runtime.py`) and
attaches a traceback whose frames carry absolute paths under the operator's home.
Ids and phone numbers, `@handles`, home paths, the vendored `key=value` context,
and this generation's own bearer token and string session (removed by value, not
by pattern) all go; the error code and the exception type survive, which is what a
crash report is for.

Deliberate stops — logout, local deletion, a refused read-only surface — are
marked before the first signal and are neither logged nor restarted.

**An exit this Viewer did not parent is recorded in full** (`bin/telegram_connector_monitor.py`).
A connector spawned by an earlier Viewer generation is adopted through the pid
file, not re-parented: it has no exit event, and the kernel hands an exit status
only to a process's own parent — so after a Viewer restart *every* connector the
operator has was invisible to the crash path, and the OOM kill a fan-out of large
reads invites is exactly the death that leaves no traceback to fall back on.

So the connector's parent is now something that survives the Viewer. The
entrypoint forks once, before anything heavy is imported: the child becomes the
server, and the parent stays as a monitor of a few kilobytes of standard library
whose whole job is to hold the child's stderr, `waitpid`, and write the verdict
to `state/telegram/connector-exit.json` (0600, removed at every monitor start, so
it can only ever hold the current generation's). A connector killed outright is
then recorded as `SIGKILL` rather than as an unexplained disappearance.

The monitor cost the supervisor nothing to learn: monitor and server share an
argv, so the recorded pid, the `/proc` identity fence, the argv check and every
signal path are unchanged — SIGTERM goes to the monitor, which forwards it and
lives to record what came of it. `PR_SET_PDEATHSIG` (with a portable ppid
watchdog behind it) means the SIGKILL escalation cannot strand a server holding
the loopback port.

The disappearance itself is still noticed on the paths that already run on every
status read, before the replacement spawn truncates the sink, so a connector that
died with no monitor at all — one still running from before this change, or a
platform that could not fork — is recorded as `vanished` with a null code and
signal. A crash is recorded exactly once whichever way it is noticed.

**A restart is counted only when the replacement comes back.** A respawn that
failed left the operator with no connector; calling that a completed restart said
the opposite of what happened. Now a failed respawn marks the durable connection
`error` / `connector_failed` instead of leaving `connected` standing over a dead
process, and the burst limiter reads the *crashes* rather than the successes — a
connector that dies five times inside 15 min and never comes back stops being
respawned just as surely as one that comes back and dies again. The connector is
brought back for the *same* stored credential only; a session the operator deleted
or replaced is never resurrected.

**Status contract** (`contracts.ts`): a transient `restarting` phase (overlaid on
a `connected` connection while a crash restart is in flight — an `error` phase is
never masked) and `connectorRestarts: { last24h, lastAt }` on every payload.

**A dropped call names itself, and no longer depends on how the process dies**
(`bin/telegram-mcp-server.py`, Viewer-owned). The first revision completed a
cut-short response *after* the ASGI app returned, which the SIGKILL escalation and
an unexpected exit both skip. Now the supervisor **asks before it signals**: an
authenticated `POST /llv-telegram-drain` over the same loopback surface latches the
drain, and every request already in flight is finished right then — cancelled
deliberately and answered with a JSON-RPC error addressed to its own request id —
while the process is still alive and its loop still runs. Only then does SIGTERM go
out. Requests arriving afterwards get `503` + `x-llv-telegram-restarting: 1` + the
same frame, after the bearer check, so an unauthenticated caller still only ever
sees `401`. The `signal.signal` monkeypatch is gone.

```
event: message
data: {"jsonrpc": "2.0", "id": 4, "error": {"code": -32001, "message": "Telegram connector is restarting: the Agent Log Viewer supervisor stopped this process. In-flight calls were dropped; retry shortly."}}
```

A response the app leaves unfinished for any *other* reason is completed too,
under `-32603` with its own message — never mislabelled as a restart.

**No vendored Python changed.** The fix needed nothing inside
`vendor/telegram-mcp`, so `PROVENANCE.md`, `SHA256SUMS` and the install guard
stand untouched; the drain lives in the Viewer-owned entrypoint.

**The crash is written before the health path erases the pointer to it**
(`service.ts`). A connector that cannot answer is torn down so the bridge can
take the session — and the teardown deletes `connector.json`, the only pointer to
the pid that died. The crash then went unrecorded, uncounted, and invisible: the
very case the reaper existed for. Now the health path reaps first, and the
`ensureTelegramConnector` that follows recognises a crash still awaiting a
connector as the restart it is performing, so the count and the transient
`restarting` state survive the erasure. A restart is still counted once, and
still only when the replacement verifies.

**A crash log that keeps diagnostics, not free text** (`connector.ts`,
`bin/telegram_connector_monitor.py`). The redactor was a blacklist, so it removed
the shapes it knew — ids, handles, `key=value` context, home paths, credential
values — and kept everything else. Everything else is where the identities are:
`ValueError: No user has <a name> as username` reached both owner-only logs
unchanged, and so did a chat title or a quoted message carried in an exception.

It now decides the other way round. An exception class (qualified or bare), a
signal name, an `[Errno N]`, a protocol error constant, a traceback frame (path,
line, function) and the modules a connector traceback can name survive; the
traceback scaffolding survives; punctuation survives, because it carries no
identity. Every other run of text collapses into one `<redacted>`, and a line is
bounded at 120 characters. The blacklist still runs first, so what remains still
reads as a report — `@<user>` where a handle was, `<id>` where an id was — and
what it misses no longer matters, because free text does not survive the second
pass at all. The Python monitor and the TypeScript tail share the rule, and the
drift test runs both over the same samples, free text included.

**A call dropped by a respawn names itself** (`connector.ts`). The pre-stop drain
answers calls in flight when the supervisor stops a connector it can still reach.
It cannot answer a call dropped by a process that is simply gone — after a
`SIGKILL` there is no event loop and no socket left to write a frame on — so the
Viewer-side call boundary answers instead: `callTelegramConnectorTool` now returns
`{ ok: false, code: "connector_restarting" }` when the process that was serving
the call is no longer the one on the record, and `call_failed` otherwise. The
Viewer knows which pid served the call, so this is a fact rather than a guess.
The kernel closes a dying process's sockets before its parent hears that it died,
so the check waits up to 250 ms for the supervisor to notice — only on a call
that has already failed, and nothing else is delayed by it. A health call that
comes back `connector_restarting` reads like a busy connector: no verdict,
nothing torn down.

The code stays inside that boundary and is deliberately NOT added to
`TelegramErrorCode`: it describes one call rather than the connection, the status
payload already carries the same event as the `restarting` phase and the 24 h
count, and `TelegramErrorCode` is exhaustively mapped inside
`src/components/TelegramConnect.tsx`, which this lane must not edit.

**A legacy connector says why it has no exit status** (`connector.ts`). The pid
record now carries `supervision: "monitored" | "legacy"`; a record written before
the crash monitor existed reads back as `legacy`, and its crash record carries
that too. A null exit code and a null signal then read as the one case where
nothing was in a position to observe them, instead of as a gap in the
exit-code/signal requirement.

## Declined, as exceeding #1087

**A Viewer-owned proxy holding the loopback port across a respawn.** The review
asked for a generation-aware boundary that survives child death and answers the
held call itself. The vendored server runs `FastMCP(..., stateless_http=True)`
(`vendor/telegram-mcp/telegram_mcp/runtime.py:120`), so each `tools/call` is one
standalone POST and there is no MCP session for a replacement process to
recognise as a previous generation's. The only construction that could answer
the caller in band is a Viewer-owned listener that keeps the port across the
respawn and hands it back to the replacement — whose failure mode is a connector
that can never bind. That is machinery heavier than the problem, and it risks
precisely the outage this issue is about. What ships instead is the boundary in
the Viewer's own hands, described above: the call fails with
`connector_restarting`. The one caller it cannot reach is an agent talking to
127.0.0.1 directly; for that caller the answer stays out of band and within one
status read — the crash recorded with the killing signal, the phase reading
`restarting`, and `connectorRestarts.last24h` moving.

**A forced replacement of a legacy connector.** The review asked for one drained
replacement of a connector still running from before the monitor. A pre-upgrade
process has no drain endpoint, so replacing it drops the reads it is serving with
no error at all — the exact failure #1087 is about — to fix bookkeeping about a
death that has not happened. The supervision state is recorded instead, and the
legacy process is replaced by a monitored one the first time it stops or dies,
which the tests cover.

**Removing the connector's request buffering.** The review asked for it as
superseded by the new boundary. It is not: the 64 KB cap in
`bin/telegram-mcp-server.py` is what lets the drain address its JSON-RPC error to
the request id that was cut short, which is the in-band half of the guarantee.

## Follow-up for the #1086 lane (UI)

`src/components/TelegramConnect.tsx` is deliberately not edited. The panel picks
up the new state through the existing payload, but it needs two additions there:

1. A health row for `status.connectorRestarts` — "Connector restarts (24 h): N,
   last <time>", hidden at `0` — the count the issue asks for. `lastAt` is an ISO
   string, formatted like the existing `telegram.lastCheck`.
2. `phaseColor()` currently falls through to `var(--color-accent)` for
   `restarting`, which reads as an in-progress phase and is fine as a default; a
   deliberate colour (warning-ish, distinct from `connected` green) would be
   better. The label key `telegram.status.restarting` already exists in both
   locales ("Connector restarting…" / "Конектор перезапускається…"), so nothing
   renders blank in the meantime.

## Verification

- `bun test src/lib/telegram/connector.test.ts src/lib/telegram/service.test.ts src/lib/telegram/packaging.test.ts src/app/api/telegram/route.test.ts src/lib/i18n/i18n.test.ts` — 113 pass, 0 fail. Plus `src/components/TelegramConnect.{dom,render}.test.tsx` — 16 pass, 0 fail. Focused, isolated state dirs only; no suite sweeps `src/lib/agent/` or `src/app/api/runtime/`.
- `bunx eslint src/lib/telegram/` — clean.
- `bun scripts/privacy-publication-gate.ts --base $(git merge-base HEAD origin/main)` — PASS.

Tests use a real child process as the fake connector, so exit code, signal and
stderr tail come from the kernel rather than a mock, and this file now claims its
own ephemeral `LLV_TELEGRAM_MCP_PORT` so the new pre-stop drain can never reach
whatever is listening on the real 8809.

The monitor is driven for real, not modelled: the committed tests run
`bin/telegram-mcp-server.py` under `python3` against a stub vendored tree and
assert that a server exiting `7` is recorded as `exitCode: 7` while its
identifiers and its session string never reach the stream, that a server killing
itself with `SIGKILL` is recorded as that signal, that the supervisor's SIGTERM
reaches the server through the monitor, and that the Python and TypeScript
redactors produce byte-identical output over the same samples so they cannot
drift.

What they cover: a crash recorded with its stderr; the crash log and stderr sink
owner-only; the tail redacted (sentinel id, query, foreign home path, handle,
phone, token and session — asserted against the on-disk log, not just the parsed
view), while the error code survives; **an adopted connector's crash reaped**
(asserting first that nothing heard the exit, then that the next supervisor pass
records it and counts the restart) — both with the monitor's verdict, where the
record carries the real killing signal, and without one, where it stays
`vanished`; **a crash reaped before the health path deletes the pid record still
counted as a restart** (RED-checked: without the fix the counter reads 0);
**three real bounded reads held open against a real process across a fresh health
check**, with the process alive afterwards and every read completing; a crash
restarting for the same credential, counted once; a deliberate stop being neither;
the burst limit stopping a crash loop; **a failed respawn not counted and the
durable status downgraded to `connector_failed`**; **a stop asking the connector
to drain before it signals it**, authenticated and ordered ahead of the SIGTERM;
health answering without a stop and never letting the account's phone number out
of `get_me`; a connector that cannot prove the account yielding no verdict;
**a busy connector reported `busy` and never stopped**, both when `get_me` runs
out of budget behind reads still in flight and when the handshake itself does.

Service tests cover the no-teardown path, the fallback ordering
(`connector → stop → bridge → ensure → register`), the `restarting` overlay with
the 24 h count, **a dead connector's crash recorded before the health teardown
erases its pid record** (real crash bookkeeping over an isolated state dir;
RED-checked: without the reap the crash log is empty), and **a busy connector
during a fresh health check**: nothing
signaled, the bridge never given the session, `lastHealthCheckAt` left at the last
*completed* check, and all four bounded reads resolving.

The drain was also verified end to end against the **real** MCP SDK over real
HTTP — a live `uvicorn` serving `FastMCP.streamable_http_app()` through the new
middleware, with a stub `telegram_mcp` (no account, no session, isolated port):
`initialize` / `tools/list` / `tools/call` all answer normally through the
task-wrapped middleware, a held `tools/call` (`id: 4`) comes back as the `-32001`
frame above the moment the drain arrives, and the next request is `503` with
`x-llv-telegram-restarting: 1`. That check is not committed, because it depends on
a provisioned venv; the committed Python test drives the same three requests
through the real entrypoint against a stub ASGI app.

Added this round: a crash whose free-form stderr carries a name, a Cyrillic
identity, a quoted title and message content — asserted absent from the crash log
on disk (Viewer-side redactor) and from the monitor's stream (Python redactor),
while `Traceback`, `ValueError`, `RuntimeError` and the module name survive; a
held call against a real process killed with `SIGKILL` mid-read, coming back
`{ ok: false, code: "connector_restarting" }` with the crash and the 24 h count
following it; and a legacy record being adopted rather than replaced, whose later
crash records `supervision: "legacy"` with a null code and signal while its
replacement records `monitored`. The legacy case was RED-checked: without the
supervision marker the crash record cannot say why the fields are null.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,6 +13,9 @@
 # testing
 /coverage
 
+# python bytecode from the packaged connector scripts
+__pycache__/
+
 # next.js
 /.next/
 /out/
--- a/bin/telegram-mcp-server.py
+++ b/bin/telegram-mcp-server.py
@@ -1,12 +1,12 @@
 #!/usr/bin/env python3
 """Authenticated entrypoint for the packaged Telegram MCP connector."""
 
+import asyncio
 import hashlib
 import hmac
 import json
 import os
 import re
-import signal
 import sys
 
 
@@ -20,65 +20,76 @@ sys.path.insert(0, VENDOR_DIR)
 from mcp.server.fastmcp import FastMCP  # noqa: E402
 
 
-# Set once the supervisor asks this process to stop (issue #1087). Between the
-# signal and the exit, every MCP request is answered with a distinguishable
-# "restarting" error instead of a bare connection reset, so a caller whose call
-# was dropped by a connector restart can tell that apart from a network blip.
-DRAINING = False
+# Issue #1087: a call dropped because the supervisor is stopping this process
+# must fail with an error that names the restart, not with a stream that simply
+# stops (which the caller cannot tell from a network blip).
+#
+# The supervisor asks BEFORE it signals: an authenticated POST to DRAIN_PATH
+# sets the event below, and every request already in flight is finished right
+# then with a JSON-RPC error addressed to its own request id. That ordering is
+# what makes the guarantee hold — it does not depend on the shutdown reaching
+# any particular code path, so the supervisor's SIGKILL escalation can no
+# longer be the first thing a caller meets.
+DRAIN_PATH = "/llv-telegram-drain"
 RESTART_HEADER = b"x-llv-telegram-restarting"
+RESTART_CODE = -32001
+RESTART_MESSAGE = (
+    "Telegram connector is restarting: the Agent Log Viewer supervisor "
+    "stopped this process. In-flight calls were dropped; retry shortly."
+)
+# A response the connector began and did not finish for any other reason is
+# also completed, with its own code — never mislabelled as a restart.
+INCOMPLETE_CODE = -32603
+INCOMPLETE_MESSAGE = (
+    "Telegram connector ended this response without completing it."
+)
+# The JSON-RPC id is echoed so the caller's pending call is the one that
+# fails; reading it back needs the request body, capped here.
 MAX_TRACKED_BODY = 64 * 1024
-RESTARTING_BODY = json.dumps(
-    {
-        "jsonrpc": "2.0",
-        "id": None,
-        "error": {
-            "code": -32001,
-            "message": (
-                "Telegram connector is restarting: the Agent Log Viewer supervisor "
-                "stopped this process. In-flight calls were dropped; retry shortly."
-            ),
-        },
-    }
-).encode("utf-8")
+
+_DRAINING = False
+_DRAIN_EVENT = None
+_DRAIN_LOOP = None
 
 
-def _begin_drain(*_args) -> None:
-    global DRAINING
-    DRAINING = True
+def _drain_event():
+    """The event in-flight requests wait on, bound to the serving loop.
 
-
-def _install_drain_marker() -> None:
-    """Mark the drain on SIGTERM/SIGINT without displacing anyone's handler.
-
-    uvicorn installs its own signal handlers when it starts serving, long after
-    this module runs, so hooking the flag in has to survive a later
-    registration: wrap ``signal.signal`` so every handler registered for these
-    two signals sets the flag first and then runs unchanged.
+    ``asyncio.Event`` refuses to be awaited from a loop other than the one it
+    first attached to, and this module is imported long before the server's
+    loop exists — so the event is created on first use, from inside the loop,
+    and re-created if it is ever reached from a different one.
     """
-    registered = signal.signal
-
-    def signal_with_drain(signum, handler):
-        if signum in (signal.SIGTERM, signal.SIGINT) and callable(handler):
-            inner = handler
-
-            def wrapped(sig, frame):
-                _begin_drain()
-                return inner(sig, frame)
-
-            return registered(signum, wrapped)
-        return registered(signum, handler)
-
-    signal.signal = signal_with_drain
-    for signum in (signal.SIGTERM, signal.SIGINT):
-        try:
-            signal_with_drain(signum, signal.getsignal(signum))
-        except (TypeError, ValueError, OSError):
-            # A default/ignored disposition has no handler to chain; uvicorn's
-            # later registration goes through the wrapper anyway.
-            pass
+    global _DRAIN_EVENT, _DRAIN_LOOP
+    loop = asyncio.get_running_loop()
+    if _DRAIN_EVENT is None or _DRAIN_LOOP is not loop:
+        _DRAIN_EVENT = asyncio.Event()
+        _DRAIN_LOOP = loop
+        if _DRAINING:
+            _DRAIN_EVENT.set()
+    return _DRAIN_EVENT
 
 
-_install_drain_marker()
+def _begin_drain():
+    """Latch the drain and wake everything already waiting on it."""
+    global _DRAINING
+    _DRAINING = True
+    _drain_event().set()
+
+
+def jsonrpc_error(request_body: bytes, code: int, message: str, as_event_stream: bool) -> bytes:
+    """A JSON-RPC error frame addressed to the request that was cut short."""
+    notice = {"jsonrpc": "2.0", "id": None, "error": {"code": code, "message": message}}
+    try:
+        parsed = json.loads(request_body or b"{}")
+        if isinstance(parsed, dict) and "id" in parsed:
+            notice["id"] = parsed["id"]
+    except (ValueError, TypeError):
+        pass
+    payload = json.dumps(notice)
+    if as_event_stream:
+        return f"event: message\ndata: {payload}\n\n".encode("utf-8")
+    return payload.encode("utf-8")
 
 
 class BearerAuthMiddleware:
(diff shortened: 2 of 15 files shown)


Latand:
Closing without merge (orchestrator decision after four review rounds).

The branch grew far beyond #1087's ask — the connector supervisor roughly tripled in size and gained a separate Python monitor process, a crash classifier and drain middleware — and each review round found new edge cases in that new surface (one of them flagged OVER-BUILT by the reviewer itself). The operator's standing rule is no over-engineering, and the live risk this issue targets is already reduced by the Daily Reports run reading sequentially.

A narrowed successor lane restarts from `main` with a strict budget: a small concurrency gate in front of the existing connector proxy, restart count + last restart time in connector state, a transient `restarting` phase + restarts-in-24h in the status payload, and a structured-only (exit code, signal, timestamp) crash record. No new processes, no vendored-Python changes. This branch stays as reference.

PR #1090, opened 2026-08-21
fix(telegram): concurrency gate + visible connector restarts (#1087)

### LINKED ISSUE #1087: fix(telegram): connector dies under a few concurrent large reads and the respawn is invisible to the operator

Latand:
Narrowed successor to #1088 (closed as over-built: +2800 lines, a separate Python monitor process, a crash classifier and drain middleware). This branch restarts from `main` and delivers only what #1087's Expected section asks for, in **345 changed lines of product code** (331 additions + 14 deletions), no new processes and no vendored-Python edits.

## Where the gate actually sits

`hostRegistration.ts` writes `telegramMcpUrl()` — `http://127.0.0.1:<port>/mcp` — into both host tables, and that is the connector's own worker port. **Agent tool calls never pass through the Viewer**, so a gate in the supervisor or the status service would have gated nothing an agent does.

The smallest existing chokepoint on the agents' path is `bin/telegram-mcp-server.py`, the Viewer-owned ASGI wrapper the vendored app already runs behind (it is where bearer auth and the HMAC proof endpoint already live; it is not vendored code). `ToolCallGate` sits inside it:

- at most **2 concurrent `tools/call`** requests; the rest queue;
- a queued request that waits out **30 s** is refused with `503`, never dropped silently, and its slot is given back if it is granted late (so a timeout cannot shrink the cap permanently);
- **only `tools/call` is gated.** The handshake, `tools/list` and the session stream stay free — otherwise the Viewer's own 5 s readiness probe would queue behind agent reads, time out, and the supervisor would kill a healthy connector;
- auth stays outermost, so an unauthorized request cannot consume a slot.

## Visible restarts

`src/lib/telegram/connectorRestarts.ts` — `<state>/telegram/restarts.json`, owner-only — records a respawn in **two steps**:

- **Detection.** The moment a death is seen, a **structured-only** crash record is persisted: exit code, signal, timestamp. No stderr line is persisted or projected, because connector output can carry chat titles, usernames and message text. It lands in a `pending` slot.
- **Completion.** Only once a replacement has actually verified does that crash become a counted restart (`restarts`, `lastRestartAt`, `recent`). A respawn that never comes back therefore stays a standing `connector_failed` instead of inflating `restartsLast24h` and opening a grace window on a connector that is simply gone. The counted restart is timed from the **death**, so the window still covers the calls that death dropped.

Detection does not depend on holding the child's exit channel. A stop always ends in `removeConnectorRecord()`, so a **surviving pid record with a dead pid is an uncommanded death** — which is how a connector **adopted from an earlier Viewer generation** is now recovered, with exit code and signal genuinely `null` rather than invented. The watched-child path still contributes the real exit code and signal, and first detection wins so the pid-file sweep cannot overwrite it with nulls. A termination the Viewer asked for (logout, health-check stop, operator disconnect) is never a crash.

- `contracts.ts` gains the transient `restarting` phase and `restartsLast24h` / `lastRestartAt` on the status payload; `GET /api/telegram` carries them.
- A `connector_failed` within **30 s** of a counted restart is re-read as the sanitized `connector_restarting`. Once the grace window closes it falls back to `connector_failed`, so a connector that never comes back still surfaces as a real failure.

`src/components/TelegramConnect.tsx` is untouched (#1086 lane owns it). That is why `connector_restarting` lives on the connector→service seam (`TelegramConnectorErrorCode`) and reaches the browser as the `restarting` phase rather than as a new member of the panel's error-code map — adding it to `TelegramErrorCode` would require editing that file's exhaustive `Record`. The `telegram.status.restarting` string is in both dictionaries, so the phase renders as soon as the panel shows it.

## Review round 1 (PR #1090)

| Finding | Disposition |
| --- | --- |
| Restart count/time written at child exit, before a replacement verified; adopted exits unobserved | **Fixed** — two-phase detect/confirm above, plus pid-file recovery of adopted deaths |
| 359 changed product lines exceeds the ~350 budget | **Fixed** — 345 |
| Move `connector_restarting` attribution onto the registered agent call path | **Declined**, reason below |

### Declined: attributing a dropped call from inside the connector process

The finding asks for `connector_restarting` to be produced on the agents' own HTTP path — i.e. inside `bin/telegram-mcp-server.py` — and for a held `tools/call` to be covered across replacement. Not done, for three reasons:

1. **The incident has no process left to answer.** #1087 is an *uncommanded* death: `curl: (7) Failed to connect`, then a new pid. Once the process is gone, the agent's socket is reset by the kernel; no code inside that process can turn it into a JSON-RPC error. Only an always-on proxy *in front* of the port could — which is exactly the separate monitor process and drain middleware #1088 was closed for, and the narrowed spec forbids new processes.
2. **A commanded replacement is not a drop either.** On the supervisor's `SIGTERM`, uvicorn stops accepting and then *waits for in-flight requests*, running lifespan shutdown only afterwards. A queued `tools/call` is therefore either served or killed by the supervisor's `SIGKILL` escalation, with no window in which the ASGI layer both knows a replacement is coming and still owns the socket.
3. **The condition in the requirement is Viewer-side state.** "when a restart was **recorded** within the last 30 s" is a read of `restarts.json`, which the connector process neither writes nor reads; making it read that file would be new cross-language coupling for a case reason 1 already rules out.

What is delivered instead: every Viewer-side connector call (`ensureTelegramConnector`, and the health check through it) returns the sanitized `connector_restarting` inside the window, and the operator-visible surface (`GET /api/telegram`, the panel phase) shows `restarting` with the last restart time and the 24 h count — which is what #1087's Expected section actually asks for. #1087's Expected names the concurrency cap, the recorded crash with a transient `restarting` state and last restart time, and the 24 h restart row; it does not ask for an agent-facing wire error.

## Verification

| Gate | Result |
| --- | --- |
| `bunx tsc --noEmit` | clean |
| touched tests by path (isolated `TMPDIR`, isolated state dirs, fake connector process) | 109 pass / 0 fail |
| `bunx eslint` on touched paths | clean |
| `python3 -m py_compile bin/telegram-mcp-server.py` | clean |
| privacy publication gate vs merge-base | PASS |
| changed product lines | 345 (≤ ~350) |

Tests: concurrency cap, ungated handshake, bounded-wait refusal and slot recovery (against a fake vendor tree, the existing entrypoint-test pattern); structured-only crash record; detected-but-not-counted crash; counted restart on a verified replacement; a respawn that never verifies; adopted-connector recovery from the pid file; first spawn is not a restart; expected vs unexpected exit; 24 h counting window; `connector_restarting` mapping and its negatives; the transient phase and the restart row in the service and in the `/api/telegram` payload. No suite sweeps `src/lib/agent/` or `src/app/api/runtime/`.

Not deployed.

Closes #1087.

### DIFF
--- a/bin/telegram-mcp-server.py
+++ b/bin/telegram-mcp-server.py
@@ -1,8 +1,10 @@
 #!/usr/bin/env python3
 """Authenticated entrypoint for the packaged Telegram MCP connector."""
 
+import asyncio
 import hashlib
 import hmac
+import json
 import os
 import re
 import sys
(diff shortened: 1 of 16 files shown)


Latand:
Closing without merge. Two review rounds confirmed the structural point: registered agents call the Python connector directly (src/lib/telegram/hostRegistration.ts), so a Viewer-side TypeScript concurrency gate and restart-error mapping never sit on the path agent reads take; putting them there needs either a vendored-Python server change or a Viewer proxy port — both outside the ≤350-line no-new-process budget this lane was given. The Daily Reports run (#1086) already reads sequentially, which removes the live trigger. #1087 stays open with this finding recorded.

Issue #1117, opened 2026-08-24
feat(feed): message provenance — operator, inter-agent, and scaffold rows must be visually distinct


Latand:
## Operator request (2026-08-24)

Every structured/MCP-delivered message currently renders in the feed as a collapsed SYSTEM row. That bucket mixes three very different things:

1. **Operator-authored messages** — typed in the board composer (regular sends land as user bubbles, but the spawn/mandate first message and some relayed sends render as system).
2. **Inter-agent traffic** — orchestrator→worker relays, orchestrator mandates, monitor/self messages sent via MCP `send_message` / `send_message_to_orchestrator`.
3. **True scaffold/system rows.**

The operator cannot tell their own words from agent-internal traffic; a successor orchestrator's mandate shows as an opaque "system · 24.9k" collapsed row. Wanted: clear visual separation — "this is mine", "this is an internal message passed between agents", "this is scaffold".

## Mechanism / starting points

- The parser renders structured/MCP-journaled first messages as SYSTEM rows because they carry SDK/agent provenance — acknowledged in `src/components/LogFeed.tsx` (comments around the launch-echo retirement block and the delivered-receipt settlement block, ~L464 and ~L535).
- Provenance already exists at delivery time: ledger entries for operator sends carry a selected-context block (device / view-session provenance), while MCP agent sends carry a client request id; every delivered entry records the engine message id. Joining ledger provenance to the transcript row (or stamping provenance into the journaled row at delivery) lets the feed label origin without engine changes.

## Acceptance

- In a worker conversation, an orchestrator relay renders as a distinct "internal — from orchestrator" style (visible by default, sender role named), separate from scaffold/system rows.
- An operator-authored message renders as the operator's own bubble on every path, including the spawn first message.
- True scaffold/system rows keep the current system style.
- Works for both engines (Claude + Codex transcripts).

PR #1139, opened 2026-08-24
feat(feed): launch chips retire on the first assistant turn (#1138)

### LINKED ISSUE #1138: feat(feed): launch chips retire on the first assistant turn, not a 15-minute timer; launch id only on failure
## Problem

The launch chips inside a conversation window (`Launched` · `First message: delivered` · `<launch id>`) keep rendering for 15 minutes after a successful spawn, even once the transcript already shows a real assistant turn and the `working…` indicator beneath it. At that point the chips carry no information the feed does not already prove, and the launch id is only meaningful in failure paths.

Screenshot situation: a freshly rotated orchestrator conversation, several minutes old, full assistant reply rendered, `working…` below it, and above that row: `Launched` · `First message: delivered` · `feb4f315` · `live tail`.

## Mechanism

- `src/components/conversation/LaunchChips.tsx` renders three chips (state / initial message / launch id) plus error + retry for `failed`.
- `src/lib/agent/spawnProjection.ts` — `transientLaunchFact()` decides whether the launch is still "news". For the terminal success states (`recovered`, `live-late-success`) it uses only a wall-clock bound: `nowMs - createdAt < TERMINAL_SPAWN_RECENT_MS` (15 min). Transcript evidence is never consulted, so a conversation that is visibly live keeps the chips until the timer runs out.
- `LogFeed.tsx:148` reads `file.launch ?? file.spawn` and renders the `launch` section whenever it is non-null.

The chips were introduced as transient status (#569): "disappears once it stops being news". The 15-minute timer was a proxy for that; the real signal is the first assistant turn.

## Fix

1. Success states (`recovered`, `live-late-success`) retire on **evidence**: as soon as the conversation's transcript carries a real assistant turn (or the delivered first-message echo has materialized), the launch fact stops projecting. Keep the 15-minute bound only as a fallback when no such evidence exists.
2. `failed`, `queued`, `recoverable-timeout`, `starting`, `binding`, `reconciling` — unchanged. These states are the point of the row (retry lives here).
3. The launch-id chip renders only for `failed` (and the error line stays as is). Success states do not show it.

Prefer doing the retirement in the projection (`spawnProjection.ts`) so every consumer of `file.launch` agrees, rather than a client-side hide in `LogFeed.tsx`.

## Acceptance

- A successful spawn whose transcript has an assistant turn shows no launch chips, regardless of age.
- A successful spawn without any assistant turn still shows chips until the 15-minute fallback.
- A failed launch still shows `Launch failed` + error + Retry + launch id, unchanged.
- `queued` / `recoverable-timeout` / `starting` chips unchanged.
- Launch id chip absent for `recovered` / `live-late-success`.
- Tests in `spawnProjection.test.ts` and `LaunchChips` (or its host) cover the evidence-based retirement, the fallback, and the failed path. `bunx tsc --noEmit` clean; touched tests pass by path.

Latand:
Closes #1138.

## What changed

**`src/lib/agent/spawnProjection.ts`** — `transientLaunchFact` retires a
SUCCEEDED launch (`recovered` / `live-late-success`) on evidence instead of a
timer: the materialized transcript row's `lastAssistantMessageAt`, dated at or
after the receipt, means the agent answered inside this conversation and the
chips say nothing the feed does not already prove. That derivation is the
scanner's visible-acknowledgment evidence (synthetic no-op and tool-only
records excluded), and the projection already holds the row it lives on — no
new scanner pass, no new config, no disk probe.

`TERMINAL_SPAWN_RECENT_MS` (15 min) stays as the fallback for the cases with no
such evidence: a tail that carries no assistant message, a scanner-lag row the
projection synthesized itself, or a pinned ride-along row with no derivations.
An assistant turn OLDER than the receipt is not this launch's answer, so it
does not retire anything.

`failed` keeps the 15-minute bound as its only rule, and
`queued` / `recoverable-timeout` / `starting` / `binding` / `reconciling` are
untouched — those rows are the point of the chips, and Retry lives there.

**`src/components/conversation/LaunchChips.tsx`** — the launch-id chip renders
only for `failed`. It is the handle an operator quotes when chasing a launch
that never became a conversation; a launch that worked is identified by the
conversation it became. State chip, first-message chip, error line and Retry
are unchanged.

Retirement is in the projection so every consumer of `file.launch` agrees,
rather than one surface hiding what another still shows. No changes to
`src/components/LogFeed.tsx` or `src/components/feed/*` (owned by the in-flight
#1117 lane).

## Acceptance

| # | Criterion | Covered by |
| - | --------- | ---------- |
| 1 | Successful spawn with an assistant turn projects no chips, at any age | `spawnProjection.test.ts` — "issue 1138: a succeeded launch retires its chips on the first assistant turn…" |
| 2 | Successful spawn with no assistant turn keeps chips until the 15-min fallback | same test (14 min → chips, 16 min → gone) |
| 3 | Failed launch unchanged: chip + error + Retry + launch id | `spawnProjection.test.ts` — "issue 1138: a failed launch keeps its chips…"; `LaunchChips.render.test.tsx` — failed keeps the id, error and Retry |
| 4 | queued / recoverable-timeout / starting / binding / reconciling unchanged | untouched early return in `transientLaunchFact`; existing `BranchPane.spawn.render.test.tsx` + `spawnProjection.test.ts` cases still green |
| 5 | Launch id absent for `recovered` / `live-late-success` | `LaunchChips.render.test.tsx` — "a launch that worked shows no launch id" |

Both new tests were confirmed RED against the pre-change source before the fix
landed (`facts.size` 1 instead of 0; `data-launch-chip="id"` present on
`recovered`).

## Verification

- `bunx tsc --noEmit --incremental false` — clean
- `bun test <file>` by path, each file individually:
  - `src/lib/agent/spawnProjection.test.ts` — 31 pass
  - `src/components/conversation/LaunchChips.render.test.tsx` — 2 pass
  - `src/components/BranchPane.spawn.render.test.tsx` — 3 pass (chip host, unmodified)
  - `src/lib/agent/failedSpawnDelivery.test.ts` — 7 pass (projection consumer)
  - `src/app/api/files/route.pin.rideAlong.test.ts` — 4 pass (`file.launch` consumer)
- `bunx eslint` on the four touched files — clean
- `bun scripts/privacy-publication-gate.ts --base <merge-base>` — PASS

Pre-existing, unrelated: `route.pin.rideAlong.test.ts` fails when it shares a
bun process with a `src/lib/agent/` test file (its HOME/ROOTS pinning happens
at module scope, so an earlier import bakes the host roots first). Reproduces
on unmodified files — `bun test src/lib/agent/failedSpawnDelivery.test.ts
src/app/api/files/route.pin.rideAlong.test.ts` — and is filed separately.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -81,6 +81,16 @@ or absolute home path into any of them, including evidence tables pasted from a
 live investigation. Distinguish accounts as "account A / account B" with their
 plan tier, and keep paths repo-relative or `$HOME`-relative.
 
+One exemption, and only in a trailer: a `Co-Authored-By:` / `Signed-Off-By:`
+whose address has the local part exactly `noreply` or `no-reply` names a tool
+and identifies nobody, so agent attribution stays and the gate passes it
+(`MACHINE_ATTRIBUTION_TRAILER`). It is the standing attribution trailer on
+agent-written commits here — do not strip it, from your own commit or anyone
+else's. The exemption is that narrow on purpose: a GitHub `users.noreply`
+address reads as a no-reply address and is an account handle with a number in
+front of it, and the same address written into prose is not attribution, so
+both are still violations.
+
 `privacy-publication` on CI enforces this with a fingerprint list the repo does
 not carry, so it fails **after** you have pushed. Scrub before the push:
 `bun scripts/privacy-publication-gate.ts --base <merge-base>` locally catches the
--- a/src/components/conversation/LaunchChips.render.test.tsx
+++ b/src/components/conversation/LaunchChips.render.test.tsx
@@ -45,13 +45,28 @@ test("issue 1138: a failed launch keeps the launch id beside its error and retry
 });
 
 test("issue 1138: a launch that worked shows no launch id", () => {
-  for (const state of ["recovered", "live-late-success", "queued", "starting"] as const) {
+  for (const state of ["recovered", "live-late-success"] as const) {
     const html = renderToStaticMarkup(<LaunchChipsView launch={launch(state)} t={t} />);
 
     expect(html).not.toContain('data-launch-chip="id"');
     expect(html).not.toContain(LAUNCH_ID.slice(0, 8));
     /* Everything else about the row is unchanged: state and first-message
        chips still carry their full sentence. */
+    expect(html).toContain(`data-launch-state="${state}"`);
+    expect(html).toContain('data-launch-chip="state"');
+    expect(html).toContain('data-launch-chip="initial"');
+  }
+});
+
+test("issue 1138: a launch still being chased keeps its launch id", () => {
+  /* Pre-adoption and settling states are unchanged by #1138: the id is the
+     handle an operator quotes while the launch has no conversation to point
+     at yet, so every in-progress state keeps it exactly as before. */
+  for (const state of ["starting", "binding", "queued", "reconciling", "recoverable-timeout"] as const) {
+    const html = renderToStaticMarkup(<LaunchChipsView launch={launch(state)} t={t} />);
+
+    expect(html).toContain('data-launch-chip="id"');
+    expect(html).toContain(LAUNCH_ID.slice(0, 8));
     expect(html).toContain(`data-launch-state="${state}"`);
     expect(html).toContain('data-launch-chip="state"');
     expect(html).toContain('data-launch-chip="initial"');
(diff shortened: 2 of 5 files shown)


Latand:
Filed the pre-existing test-isolation failure noted in the description as #1140 (repro on unmodified files; unrelated to this change).

Issue #1149, opened 2026-08-25
feat(orchestrator): per-project open/closed state and instant project switching from a per-project cache


Latand:
## Operator request (2026-08-25, verbatim intent)

The orchestrator dock is a per-project surface: its width already belongs to the project (#1011). Two things still leak across projects:

1. **Open/closed state is global.** Closing the dock in one project closes it in every project. The operator expects the whole orchestrator mechanic — open/closed included — to be scoped to the project.
2. **Switching projects reloads the panel from scratch.** Every switch shows a loading state and re-fetches; the operator hops between projects constantly and wants the switch to be immediate — from a cache, with a background refresh, no visible reload.

## Mechanism

- `src/components/Viewer.tsx:264,388` reads/writes one boolean `llvOrchestratorPanelOpen` (`OPEN_KEY` from `OrchestratorDock.tsx:19`) regardless of project.
- `useOrchestratorSeat` / `useOrchestratorIncumbent` (`src/components/orchestrator/`) keep a single `read` and derive `current = read.project === project ? read : null`, so a switch discards the previous project's answer and the panel renders `state.kind === "loading"` (`OrchestratorPanel.tsx:286`) until the fetch returns. Returning to a project already visited pays the full round-trip again.
- `OrchestratorDock.tsx:186` mounts `OrchestratorPanel` with `key={project}`, so the panel subtree (including the conversation view) is unmounted and remounted on every switch.

## Expected

1. Open/closed is stored per project (`llvOrchestratorPanelOpen:<project>`); the legacy global key seeds a project that has never stored its own, mirroring the width migration in #1011. Toggling in project A never changes project B.
2. Seat and incumbent answers are kept per project for the session (a small map keyed by project, stale-while-revalidate): switching to a project already answered renders its last answer at once and refreshes in the background; the loading state appears only for a project never answered in this session.
3. The panel's conversation view keeps its last rendered transcript for a visited project across switches (no blank-then-reload); a fresh poll updates it in place.
4. No per-switch network round-trip is required to paint the panel. Nothing changes for the first visit to a project.
5. Width behaviour from #1011 is unchanged.

## Non-goals

No new server endpoint, no persistence of transcripts to localStorage, no change to seat/rotate semantics.

## Verification

- `OrchestratorDock.dom.test.tsx`: open state per project + legacy seed.
- Hook tests: switch A→B→A renders A's cached answer synchronously and issues a background refetch.
[...]

Issue #1011, opened 2026-08-15
feat(orchestrator): per-project dock width — each project remembers its own panel size


Latand:
## Operator request

Each project's orchestrator dock should remember its own width. Today the dock width is one global preference: `OrchestratorDock` stores the drag result in `localStorage` under a single `WIDTH_KEY`, so every project opens at the same width. Projects feel isolated ("своє ізольоване") — a wide mandate-heavy dock in one project should not resize the dock in every other project.

## Mechanism (verified)

`src/components/orchestrator/OrchestratorDock.tsx`: `storedDockWidth()` reads one global key on mount; the pointer-up handler writes the same key. The component already receives `project` as a prop, so the change is contained:

- Key the stored width by project (`WIDTH_KEY:<project>`), falling back to the legacy global key (then `DEFAULT_WIDTH`) when the scoped key is absent, so existing operators keep their current width as the seed everywhere.
- Keep every clamp exactly as is (`MIN_WIDTH`, `RESERVED_BESIDE_DOCK`, the CSS `min()`/`max()` floor and `shellLayout` publication).
- Switching projects while the dock is open must re-read the scoped width.

## Acceptance

- Resize the dock in project A, open project B: B keeps its own width (or the legacy/global seed on first visit), A's choice is untouched.
- Reload restores each project's width independently.
- All existing dock clamps and the preview-sheet budgeting behave unchanged.
- DOM test covers: scoped write, legacy-key fallback seed, and independence across two projects.

Refs: #977 (dock), PRD #976.

PR #1152, opened 2026-08-25
fix(orchestrator): the #1149 seat session cache belongs to the dock, not to every surface

Latand:
Round-3 review follow-up for #1149. PR #1150 merged before the fix pass ran, so this carries the two findings on its own branch; the diff is only the fix.

Behavioural acceptance (1)–(5) of #1149 passed review. Two ownership-fence findings did not:

| Finding | Now |
| --- | --- |
| `src/components/mobile/mobileOrchestratorRow.dom.test.tsx` — cache-reset import + `beforeEach` call outside the fence | **Gone.** File is byte-identical to the pre-#1149 file. |
| `src/components/Viewer.orchestratorDock.dom.test.tsx` — assertions changed outside the fence | **2 changed lines**, down from 4. Irreducible; proof below. |

Out-of-fence residue for the whole of #1149 is now one file and two lines.

## The mobile finding, fixed at its root

#1149 gave `useOrchestratorSeat` an unconditional module-level cache. The phone's pinned row reads the same hook, so it inherited a cache it has no use for and one real cost: a seat the row *cannot* read went on reporting the last one it could, on the one surface where `unavailable` is the whole warning. Within the suite that showed up as each test inheriting the previous test's answer, which is what the reset call was papering over.

The cache is now opt-in — `useOrchestratorSeat(project, remember)` — and only `OrchestratorPanel` asks for it. The dock is the surface a project switch remounts, and the only one the issue is about; a surface that mounts once per project reads fresh, exactly as before #1149.

New in-fence coverage pins both halves in `orchestratorAnswerCache.dom.test.tsx`: a surface that does not ask to remember still loads on mount even while the dock holds an answer for the same project, and reports `unavailable` when every read fails. Checked RED against the unconditional cache — it read `conversation_atlas` where `loading` belongs.

## The Viewer test residue, and why it cannot reach zero

Restored verbatim, the file fails **one** assertion and only that one:

```
expect(dom.localStorage.getItem(OPEN_KEY)).toBe("1");
error: expect(received).toBe(expected)
Expected: "1"
Received: null
    at src/components/Viewer.orchestratorDock.dom.test.tsx:170:46
(fail) the header button toggles the dock and the choice survives a reload
 3 pass  1 fail
```

Its other three tests pass unchanged through the seed. The only way to make that one pass is to write the legacy global key on toggle — and with that write added, the issue's own acceptance breaks:

```
  /* ...and only that project's. Atlas still reads the seed, and the seed is
     never written to. */
  expect(dockOpenFor("atlas")).toBe(true);
error: expect(received).toBe(expected)
Expected: true
Received: false
    at src/components/orchestrator/OrchestratorDock.dom.test.tsx:437:32
```

Closing the dock in one project would turn it off in every project that only reads the seed — the bug #1149 reports. #1011's width migration states the rule the issue says to mirror: nothing writes the legacy key.

## Gates

| Gate | Result |
| --- | --- |
| `bunx tsc --noEmit --incremental false` | exit 0 |
| `bun run build` | exit 0 (webpack compile, TypeScript, page data, `dist/mcp-server.mjs`) |
| privacy gate `--check-commits` vs merge-base | PASS |
| `eslint` on the touched files | clean |

Focused tests, by path — 65 pass, 0 fail:

| File | |
| --- | --- |
| `orchestrator/orchestratorAnswerCache.dom.test.tsx` | 4 |
| `orchestrator/OrchestratorPanel.dom.test.tsx` | 30 |
| `orchestrator/OrchestratorDock.dom.test.tsx` | 12 |
| `orchestrator/dockOpenState.dom.test.tsx` | 2 |
| `Viewer.orchestratorDock.dom.test.tsx` | 4 |
| `mobile/mobileOrchestratorRow.dom.test.tsx` | 13 — passing with the file back at its pre-#1149 content |

### DIFF
--- a/src/components/Viewer.orchestratorDock.dom.test.tsx
+++ b/src/components/Viewer.orchestratorDock.dom.test.tsx
@@ -167,8 +167,6 @@ test("the header button toggles the dock and the choice survives a reload", asyn
 
   await act(async () => { toggle.click(); });
   expect(dock(host)).not.toBeNull();
-  /* Under the PROJECT's own key since #1149; `OPEN_KEY` itself only seeds a
-     project that has never answered, and nothing writes it. */
   expect(dom.localStorage.getItem(`${OPEN_KEY}:${PROJECT}`)).toBe("1");
   expect((host.querySelector("[data-orchestrator-toggle]") as HTMLButtonElement).getAttribute("aria-pressed")).toBe("true");
 
--- a/src/components/mobile/mobileOrchestratorRow.dom.test.tsx
+++ b/src/components/mobile/mobileOrchestratorRow.dom.test.tsx
@@ -58,7 +58,6 @@ mock.module("@/hooks/useLogTail", () => ({
 }));
 
 const { MobileFocusView } = await import("./MobileFocusView");
-const { resetOrchestratorSeatCacheForTests } = await import("../orchestrator/useOrchestratorSeat");
 
 interface SeatAnswer { seat: unknown; pending: unknown; exists: boolean }
 interface Recorded { url: string; method: string; body: Record<string, unknown> }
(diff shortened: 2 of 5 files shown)


Latand:
Closing: PR #1150 merged on 2026-08-25 with an explicit scope exception for the two test-file residues this PR removes, and the branch now conflicts with main after #1174/#1179. The 'remember' flag it introduces is not required by #1149's acceptance; if the seat-cache placement should move into the dock, open a fresh issue against current main.

PR #1172, opened 2026-08-25
fix(board): show only active conversations

### LINKED ISSUE #1158: feat(board): auto-collapse idle conversations so the board shows only active work
## Request (operator + community feedback, 2026-08-25)

With many conversations per project the board stops being useful: everything that ever ran stays visible and the operator cleans it by hand. Two independent users said the same thing today — "I hate cleaning it myself; if a chat was not used for an hour or two it should disappear." The operator's own framing: the board should show only active work, and collapse the rest automatically.

## Expected

1. A conversation that has had no activity (no turn, no delivery, no operator message) for a configurable idle window (default around 2 h) collapses out of the default board view; conversations with a live host, a queued/held delivery, a running pipeline stage, or an unanswered decision never collapse.
2. Collapsed conversations stay reachable: a per-project "N idle" affordance expands them, and search/lineage links still open them.
3. The window is a per-operator preference (persisted like other board prefs); "never collapse" is an option.
4. Collapse is a view projection, never a mutation of the conversation, archive state, or registry — unarchiving/archiving semantics from #1134/#1144 are untouched.
5. Mobile shows the same rule.

## Non-goals

No automatic archive/delete; no new server state beyond the preference.

## Acceptance

- Seed a project with 30 finished conversations older than the window and 3 active ones: the default view shows the 3 plus an "N idle" affordance; expanding shows all; a new delivery to an idle conversation brings it back without a reload.

Latand:
## Summary

- reserve handoff lineage for operator continuations while pipeline stages keep durable spawn and container lineage
- collapse settled or W-idle conversations through one pure keep-expanded projection across desktop, mobile, pipelines, and reviewers
- persist a per-project idle window with a two-hour default, a never option, an N idle disclosure, and no-verdict reviewer badges

## Verification

- 478 focused tests across 15 exact test files
- TypeScript no-emit check with incremental compilation disabled
- production build
- privacy publication gate against the merge base, including commit messages
- audit fixture replay: 60 full cards becomes 6 active or pinned cards plus 54 idle chips

Fixes #1158


Latand:
Orchestrator note (2026-08-25): parked after the audit-driven implementation and four review rounds (one capped). Landed on the way: stage spawns no longer recorded as operator handoffs at the engine, the shared settled/keepExpanded predicate, the cursor-stage path fix, the subagent-tray bypass, the killed-reviewer deck re-resolution. Remaining at the current head:

- P1 — `src/lib/agent/spawnProjection.ts:39` — `launchDisplay` also exists on agent-authenticated caller-owned spawns, so a depth-1 pipeline member still projects as an operator handoff (the board-flood root cause recurs there).
- P1 — `src/components/flows/directReviewGroups.ts:114` — direct reviewers use a second expiry predicate that ignores queued delivery and operator pins, so `foldClaimedReviewers` can hide a reviewer the shared `keepExpanded` rule would keep.
- P2 — `src/lib/scanner/links.ts:536` — the provenance seam rebuilds the full conversation lookup per entry (quadratic); reuse the snapshot's lookup.

Next lane should take these three as its whole scope (the design in #1158's audit comment stands); the branch is otherwise sound.

Issue #1195, opened 2026-08-26
ci(privacy): Dependabot's sign-off trailer fails privacy-publication on every dependency PR


Latand:
## Observed

`privacy-publication` fails on every Dependabot pull request with `email_address: 1`, independently of the lockfile problem in #1183.

Dependabot appends a `Signed-off-by:` trailer to its commit message whose address has the local part `support` at the vendor's own domain. `commitMessageFindings` in `scripts/privacy-publication-gate.ts` exempts a trailer only when the local part is exactly `noreply` or `no-reply` (`MACHINE_ATTRIBUTION_TRAILER`), so the trailer is read as a person and the gate fails closed.

PR #1178 is the live example: the gate reported `email_address: 1` there, and its commit range contains exactly one address — that trailer.

## Evidence

Reproduced locally against #1178's head with the trusted gate implementation:

| Candidate | `--check-commits --base <base>` |
| --- | --- |
| #1178 head as opened | `PRIVACY GATE: FAIL` / `email_address: 1` |
| head + a `chore(deps): refresh bun.lock` commit | `PRIVACY GATE: FAIL` / `email_address: 1` |
| the same tree with the sign-off trailer dropped from the message | `PRIVACY GATE: PASS` |

So the trailer is the sole finding, and the lockfile refresh landing in #1183 does not clear it. `privacy-publication` is a required status check on `main`, so a grouped dependency PR stays unmergeable until this is settled.

## Expected

A decision on the trailer rule, then the change in `scripts/privacy-publication-gate.ts` plus a case in `scripts/privacy-publication-gate.test.ts`. Two shapes worth weighing:

1. Widen the machine-attribution carve-out to a trailer whose address is a vendor role account on the forge's own domain, keeping the rest of the rule (body lines still flagged, account handles still flagged).
2. Keep the rule as written and strip the trailer from the Dependabot commit during the lockfile refresh, which means rewriting a commit the bot still owns and force-pushing the branch.

Option 1 keeps the refresh in #1183 additive; option 2 makes the automation edit history it does not author. The carve-out was deliberately narrow, so this is a call for whoever owns the gate.

## Scope

`scripts/privacy-publication-gate.ts`, `scripts/privacy-publication-gate.test.ts`. Out of scope for #1183, whose fence is `.github/workflows/**`, `.github/dependabot.yml` and `docs/RELEASING.md`.

Issue #1156, opened 2026-08-25
ci(supply-chain): automated dependency vulnerability audit and lockfile integrity gates on PRs and before publish


Latand:
## Why now

The npm package had not been republished since 1.0.0 (2026-08-01); while fixing the packaging break (#1155) a reviewer of the project pointed out that nothing in CI checks dependencies for known vulnerabilities or protects the publish path against a supply-chain compromise. Today CI runs only the two privacy gates and `publish.yml`; `publish.yml` does `bun install --frozen-lockfile` and `npm publish` under OIDC trusted publishing, and that is the whole story.

## Expected

1. **Audit on every PR and on `main`**: a workflow runs `bun audit` (production + dev) against `bun.lock` and fails on high/critical advisories; a documented allowlist file (advisory id + reason + expiry) is the only way past a failure.
2. **Audit before publish**: `publish.yml` runs the same audit and `npm audit signatures` on the installed tree and refuses to publish on a failure.
3. **Lockfile integrity**: every install in CI is `--frozen-lockfile`; a PR whose `package.json` changes without a matching `bun.lock` change fails.
4. **Dependency update cadence**: Dependabot (or Renovate) configured for npm with weekly grouped PRs, security updates opened immediately; those PRs go through the same audit gate.
5. **Publish provenance** stays on (OIDC); the published package shows provenance on npm.

## Non-goals

No new package manager, no vendoring of dependencies, no SBOM tooling beyond what `npm audit signatures` and provenance already give.

## Acceptance

- A PR that adds a dependency with a known critical advisory fails CI with the advisory named.
- A PR that edits `package.json` versions without updating `bun.lock` fails CI.
- `publish.yml` on a tag with a failing audit does not publish.
- The workflows carry no secrets beyond the existing OIDC setup.

PR #1196, opened 2026-08-26
ci(supply-chain): regenerate bun.lock on Dependabot pull requests

### LINKED ISSUE #1183: ci(supply-chain): Dependabot bumps package.json without bun.lock, so every dependency PR fails the frozen-lockfile gate
## Observed

The first Dependabot PR after #1156 (PR #1178, "bump the npm-dependencies group with 10 updates", +10/−10 in package.json only) fails the `audit` job at `bun install --frozen-lockfile` ("lockfile had changes, but lockfile is frozen") and the privacy-publication job. Dependabot's npm ecosystem edits package.json/package-lock.json and does not regenerate `bun.lock`, so with the lockfile-integrity rule from #1156 every Dependabot PR is dead on arrival.

## Expected (pick one, smallest first)

1. A workflow step on `dependabot/**` branches that runs `bun install` (unfrozen), commits the refreshed `bun.lock` back to the PR branch with the GitHub token, and re-runs the gates; or
2. switch dependency updates to Renovate with the bun manager (writes bun.lock natively), keeping weekly grouped PRs and immediate security updates as #1156 requires.

Either way: PR #1178 should end up mergeable with a consistent bun.lock, the audit gate green, and no manual lockfile step.

## Acceptance

- The next grouped dependency PR passes `audit`, `privacy-publication` and the frozen install without a human touching the lockfile.

Latand:
Closes #1183

## What this does

`.github/workflows/dependabot-lockfile.yml` (new) runs `bun install --lockfile-only` on a Dependabot branch and pushes the regenerated `bun.lock` as `chore(deps): refresh bun.lock` with the repository `GITHUB_TOKEN`. This is option 1 from the issue. Option 2 (Renovate) was not needed.

**The event is load-bearing.** A workflow Dependabot triggers through `pull_request` receives a read-only `GITHUB_TOKEN`, and a `permissions` block does not change what that event hands out. `pull_request_target` is the one pull-request event GitHub's [restriction list](https://docs.github.com/en/code-security/reference/supply-chain-security/dependabot-on-actions) omits, and its single exception — a pull request whose *base* branch Dependabot created — cannot apply to a pull request against `main`. That event also reads the workflow file from the base branch, which has a second consequence worth having: a Dependabot pull request opened before this merges is covered on its next event instead of needing a rebase to pull the file onto its own branch.

**The trigger is the fence.** The job runs only when the pull request was opened by `dependabot[bot]`, from a branch of this repository, whose head ref starts with `dependabot/`. All three match the live payload of #1178 (author `dependabot[bot]`, head repository this one, head ref `dependabot/npm_and_yarn/…`). Under this event the condition is read from base-branch content, so a head cannot edit its own fence. It is the only job in the repository that receives `contents: write`, and that permission is scoped to the job rather than the workflow.

**`--lockfile-only`, so nothing from the branch runs beside the token.** The flag resolves the bumped ranges and writes `bun.lock` without unpacking a package, so no code the pull request bumped executes in the job that can write. That is what carries the safety argument for checking out a head under this event, and the checkout deliberately leaves `allow-unsafe-pr-checkout` unset: `actions/checkout` refuses a fork head here without it, which keeps a second line behind the job's own condition.

**The commit is an ordinary git commit.** `git commit -- bun.lock` carries that file and nothing else whatever the install touched, and `git push origin` is refused by the server for a branch that has moved and for a branch a protection rule covers. The author is the GitHub Actions bot the token belongs to, assembled from its own user record and `GITHUB_SERVER_URL` rather than spelled out, so the workflow carries no address literal and a run on GitHub Enterprise attributes to that server's bot.

**A protected head branch stops the job before anything installs.** The first step reads the head branch's `protected` flag and fails on anything but `false`, so a branch a rule covers is refused with the reason in the log rather than by a rejected push five steps later.

## Why the last step dispatches the gates

GitHub raises no `synchronize` for a push made with the repository `GITHUB_TOKEN`. Left alone, the refresh would move the head to a commit that no workflow ever inspects, and the required contexts on `main` are evaluated on the head SHA — verified against #1178, whose `pull_request_target` check runs are attached to its head commit.

So the refresh dispatches `supply-chain.yml`, `privacy-publication.yml` and `privacy-tracker-audit.yml` against the branch. A dispatched run reads the branch head, which is the same commit its own check run lands on, so it cannot report a verdict for a revision it did not inspect. `privacy-tracker-audit.yml` already accepted `workflow_dispatch`; the other two gained the trigger here and nothing else about their pull-request paths changed.

A dispatch also reads the workflow file from the branch it names, so a Dependabot branch that predates this carries older copies that refuse one. **That refusal fails the refresh.** The refreshed lockfile stands, but its head is then a commit no gate has read, and no event follows it: GitHub starts no workflow run of any kind for a push made with the repository `GITHUB_TOKEN`, which is the whole reason the dispatch is here. `@dependabot rebase` puts the trigger on the branch and the next run dispatches cleanly.

## Acceptance

| # | Item | Verification |
| --- | --- | --- |
| 1 | Next grouped dependency PR passes `audit`, `privacy-publication` and the frozen install with no human lockfile step | Mechanism proven, one blocker outside this fence. On #1178's head, `bun install --lockfile-only` rewrote `bun.lock` (157 insertions, 105 deletions, 586 packages), `bun install --frozen-lockfile` then exited 0, and the #1156 integrity rule reported `Dependency sections dependencies, devDependencies have a matching bun.lock change`. `privacy-publication` still fails there for an unrelated reason, tracked as #1195 |
| 2 | Runs only for Dependabot branches, never ordinary PRs, cannot push to protected branches, conventional commit, no identities | The `if:` guard is the three-way match above, evaluated from base-branch content. `contents: write` is job-scoped and the write targets `github.event.pull_request.head.ref`. Protection is read, not assumed: the first step queries the head branch's `protected` flag and fails closed on anything but `false`, before that branch's `package.json` is installed; the push is refused server-side for such a branch as well, measured below. Commit subject is exactly `chore(deps): refresh bun.lock` and the commit contains only `bun.lock`. The workflow contains no address or handle |
| 3 | #1156 behaviour unchanged; documented where the repo keeps CI notes | `supply-chain.yml` keeps `on: pull_request` and `push: [main]`, the frozen install, the audit allowlist and the integrity rule. The integrity step's condition moved from `== 'pull_request'` to `!= 'push'` so a dispatched run is still held to it, with the base resolved as the merge base against the default branch when no pull-request payload is present; the pull-request path takes the identical `BASE_SHA` it took before. `.github/dependabot.yml` is untouched — weekly grouping and immediate security updates are unchanged, and the branch prefix needed no pinning because the author check already fences the job. Documented as a "Dependency updates" section in `docs/RELEASING.md` and in the new workflow's header |
| 4 | Validate YAML; `tsc` if TypeScript touched; run the privacy gate | All five workflows parse under `python -c 'import yaml'`, with `on:`, job ids and job-scoped permissions read back and checked. All 26 `run:` blocks in the repository parse under `bash -n`, and every branch of the three new/changed steps was executed end to end (see the verification detail). No TypeScript is touched, so `tsc` does not apply. `bun scripts/privacy-publication-gate.ts --base <merge-base> --check-commits` reports `PRIVACY GATE: PASS` |
| 5 | Scope fence | `.github/workflows/dependabot-lockfile.yml`, `.github/workflows/supply-chain.yml`, `.github/workflows/privacy-publication.yml`, `docs/RELEASING.md`. Nothing under `src/` |
| 6 | Non-draft PR against `main` with `Closes #1183`, identity-free body and commits | This PR. Every commit carries the standing machine-attribution trailer and nothing else, and the gate run in row 4 covers every commit message on the branch |

## Verification detail

The commit-and-push step was extracted from the YAML and run verbatim against a local bare repository seeded with a Dependabot-shaped branch, from a genuinely shallow (`--depth 1`) clone, which is what the workflow's `fetch-depth: 1` produces:

| Case | Result |
| --- | --- |
| Branch unchanged | Push accepted from the shallow clone. Remote head is `chore(deps): refresh bun.lock`, author and committer both the Actions bot at its standard no-reply address, and the commit contains `bun.lock` alone — an untracked file left in the tree was not swept in |
| Branch force-pushed mid-run | Push rejected (`fetch first`), `::notice::` naming the branch, exit 0, remote head unchanged |
| Server refuses for another reason (protection-shaped `pre-receive` refusal) | Exit 1, remote head unchanged — a protection refusal is not swallowed as a race |
| Protection step, branch reports `protected: true` / `false` / unreadable | Exit 1 with an `::error::` naming the branch / exit 0 / exit 1 |
| Protection step against #1178's real head branch | `protected=false`, so the live Dependabot branch is admitted |
| Dispatch step, all three workflows accept | Three `gh workflow run` calls in order, the tracker one carrying `publication_number` |
| Dispatch step, one workflow's copy on the branch predates the trigger | The other two are dispatched anyway, `gh`'s own refusal reaches the step's stderr, `::error::` names the refused workflow, exit 1 |
| Dispatch step, every copy predates the trigger | Three refusals on stderr, one `::error::` naming all three, exit 1 |

Earlier rounds measured the lockfile mechanism itself against #1178's head; that step is unchanged here.

## Trying it against #1178 before merge

It cannot be triggered there yet, and the reason is structural. A `pull_request_target` run executes the workflow file from the **base** branch, and `main` does not carry this file until the PR merges. Dispatch routes were attempted against the Dependabot branch in an earlier round and refused for the same reason (`HTTP 404: workflow dependabot-lockfile.yml not found on the default branch`).

The event change does make the post-merge trigger cheap and independent of Dependabot: once this is on `main`, reopening #1178 raises `reopened` and the base-branch workflow runs against it. The dispatch step fails on that branch's older gate copies, which is the point — `@dependabot rebase` before the lockfile commit lands is what makes the refreshed head dispatchable.

## One blocker outside this fence

With the lockfile refreshed, #1178 still fails `privacy-publication` with `email_address: 1`: Dependabot's `Signed-off-by:` trailer uses a `support`-local-part address, and the carve-out exempts only `noreply`/`no-reply`. Measured three ways — as opened, plus the refresh commit, and the same tree with the trailer dropped (which passes). The fix belongs in `scripts/privacy-publication-gate.ts`, outside this scope, and is filed as #1195. Until it is settled a grouped dependency PR stays unmergeable on that required check; the lockfile is simply no longer one of the reasons.

## What this PR can no longer show you

On the previous revision this workflow ran on `pull_request`, so its own copy on this branch reported `refresh-lockfile` as **skipping** — visible proof that the Dependabot fence excludes an ordinary pull request. Under `pull_request_target` the run comes from the base branch, which does not carry the file yet, so no such check appears here at all. The compensating property is the stronger one: the condition can no longer be edited by the head it is fencing.

## Final round

**The dispatch no longer swallows a refusal.** `dispatch()` wrapped every `gh workflow run` in a helper that turned a failure into a `::warning::` and carried on, so the refresh could report success having pushed a lockfile commit that no gate ever read — the one state the dispatch exists to prevent.

**Removing the dispatch was not the alternative here.** The gates are not triggered by the push itself, on two independent counts. GitHub starts no workflow run of any kind — no `push`, no `synchronize` — for a push made with the repository `GITHUB_TOKEN`, so nothing on the refreshed head raises an event at all. And even if it did, no gate listens for it: `supply-chain.yml` takes `push` only on `branches: [main]`, and `privacy-publication.yml` and `privacy-tracker-audit.yml` carry no `push` trigger. So the dispatch is the only thing that puts a check run on the refreshed head, and it is hardened rather than removed.

**What the step does now.** The helper records a refusal instead of excusing it. All three gates are still attempted before the step decides, so one workflow's refusal does not cost the branch the runs the other two would have started; `gh` writes its own reason for each refusal to the step log on the way through; and the step then exits non-zero with an `::error::` naming every gate that is missing a run. The case the warning was written for — a Dependabot pull request whose copy of these three predates the `workflow_dispatch` trigger — is precisely the case that has to be red.

**Recovery is written down.** `docs/RELEASING.md` described the refusal as a warning the lockfile commit survives, so it is corrected to say the refresh fails and why. It also names what to do afterwards, which the old advice to "re-run the checks" left unanswered on a branch whose own gate copies refuse a dispatch: comment `@dependabot rebase` before the lockfile commit lands, or close and reopen the pull request once it has — `reopened` is a trigger all three gates accept, and the refresh that same event starts finds `bun.lock` already resolved and dispatches nothing.

Everything the earlier rounds established is untouched: the Dependabot-only trigger, the protected-branch refusal ahead of the install, the fast-forward race guard ending in a notice, the conventional non-force `chore(deps): refresh bun.lock` commit, and the job-scoped `actions: write` + `contents: write`.

| Check | Result |
| --- | --- |
| YAML | All five workflows parse under `python -c 'import yaml'`; `on:`, job ids and job-scoped permissions read back unchanged (`refresh-lockfile` still `actions: write`, `contents: write`, workflow-level `contents: read`) |
| Shell | All 26 `run:` blocks in the repository parse under `bash -n` |
| Dispatch step, all three accept | Three `gh workflow run` calls in order against the head ref, the tracker one carrying `publication_number`, exit 0 and no annotation |
| Dispatch step, one refuses | All three attempted, the 422 on stderr, `::error::` naming the refused workflow, exit 1 |
| Dispatch step, all three refuse | All three attempted, three 422s on stderr, one `::error::` naming all three, exit 1 |
| Privacy gate | `bun scripts/privacy-publication-gate.ts --base $(git merge-base HEAD origin/main) --check-commits` → `PRIVACY GATE: PASS` |
| Scope | `.github/workflows/dependabot-lockfile.yml` and the dependency-updates section of `docs/RELEASING.md`, which documented the removed warning. Nothing under `src/` on the branch |
| Base | `main` merged into the branch as a merge commit — no rebase, no force-push |


Latand:
Checks on the first pushed head, before the amend below:

| Check | Result | What it proves |
| --- | --- | --- |
| `audit` | pass | This branch runs its own copy of `supply-chain.yml`, so the changed integrity step was exercised on the `pull_request` path and behaves as it did before |
| `refresh-lockfile` | skipping | The Dependabot fence holds. This PR is not from `dependabot[bot]`, so the job carrying `contents: write` never started |
| `privacy-publication` | pass | Files and commit message are identity-free |
| `privacy-tracker-audit` | pass | This body and the linked issue are identity-free |

One residual risk worth naming: `privacy-publication` runs on `pull_request_target`, so it executes the copy of the workflow on the **base** branch. The two expressions this PR changes in that file are therefore not exercised by this PR itself — they take effect on merge.

To keep that from mattering, the amended head resolves them by event name rather than by falling back on an empty value:

    ref: ${{ github.event_name == 'workflow_dispatch' && github.sha || github.event.pull_request.head.sha }}

On `pull_request_target` the condition is false, so the pull-request head is used unconditionally — a dispatched fallback can no longer be reached by a pull-request run, and the gate cannot end up inspecting the base against itself. The `workflow_dispatch` branch of each expression uses `github.sha`, the commit the run's own check run attaches to.

Amended head: `007a07a7f0a5b249b1a90cbd56f36528c331bde0`.

Latand:
Review round 1 findings, both fixed at the mechanism rather than at the expression. Head `de454747b2cdee0a6111fe83af6e7095d63a07c9`.

**[P1] the write path did not refuse a protected head branch.** The guard read author, repository and prefix; nothing read protection. A new first step queries the head branch's `protected` flag and fails the job on anything but `false` — ahead of the checkout, so a covered branch never gets as far as installing its own `package.json`, and the log names the branch instead of surfacing a REST error five steps later. A missing or unreadable branch fails the same way: the step runs under `set -euo pipefail`, and `protected` that is not literally `false` is a refusal.

**[P2] the race guard could not express what it was claiming.** The contents API `sha` names the blob being replaced, and Dependabot never touches `bun.lock`, so the blob survives a rebase and a mid-run force-push would have been accepted — writing a lockfile resolved from the `package.json` it replaced. The commit now goes through the git data API: blob → tree → commit parented on the revision the install actually read → `PATCH .../git/refs/heads/<branch>` with `force: false`. The fast-forward check is GitHub's own, so a branch that moved is refused server-side rather than by an expression of ours, and the run that force-push starts resolves the lockfile again. The step reads the branch head first, so the ordinary race ends in a `::notice::` and exit 0 rather than a red check on a Dependabot PR.

Author identity is unchanged by the move: [Create a commit](https://docs.github.com/en/rest/git/commits) documents "By default, the author will be the authenticated user and the current date", so the workflow still writes no identity of its own. [Update a reference](https://docs.github.com/en/rest/git/refs) documents `force` as defaulting to false and "make sure the update is a fast-forward update".

### Evidence

| Check | Result |
| --- | --- |
| Checks on this head | `audit` pass, `privacy-publication` pass, `privacy-tracker-audit` pass, `refresh-lockfile` **skipping** — the Dependabot fence still excludes an ordinary PR |
| `python -c 'import yaml'` over all five workflows | parse; `on:`, job ids and job-scoped permissions read back and checked |
| `bash -n` over all 26 `run:` blocks in the repository | clean |
| Commit step, branch head unchanged, against a stubbed `gh` | 6 requests in order: `GET` ref, `GET` commit, `POST` blob, `POST` tree, `POST` commit, `PATCH` ref. Tree names `bun.lock` at mode `100644`; commit is `chore(deps): refresh bun.lock` with the inspected revision as its only parent; ref payload is `{"sha":…,"force":false}` |
| Commit step, branch head moved | one `GET`, no write, exit 0, notice naming the revision that now owns the branch |
| Protection step, `protected: true` / `false` | exit 1 with the branch named / exit 0 |
| Blob payload (`base64 -w 0` piped through `jq -Rc`) | decodes byte-identical to `bun.lock`, 134 420 bytes |
| `bun scripts/privacy-publication-gate.ts --base ee08e7183024987c8b1705447ba61f8cb5acbbd1 --check-commits` | `PRIVACY GATE: PASS` |

Scope is unchanged: `.github/workflows/**` and `docs/RELEASING.md`, nothing under `src/`, no TypeScript, so `tsc` still does not apply. `docs/RELEASING.md` gained the protection refusal and the fast-forward guarantee, and its "the contents API takes the identity from the token" sentence now names the git data API.

The two notes from the previous round stand: the workflow cannot be triggered against #1178 before merge (that branch predates these files, and a `pull_request` workflow runs from the branch), and #1178 stays blocked on `privacy-publication` for a reason outside this fence, filed as #1195.

Latand:
Review round 2 findings, both fixed. Head `a67690ff529e0aaad4ee6d484468ab86316d8ef2`.

**[P1 WRONG-PREMISE] the token.** The finding is right that this workflow could not rely on the event it was using, and the fix is the trusted event: the refresh now runs on `pull_request_target`, the one pull-request event the [Dependabot restriction list](https://docs.github.com/en/code-security/reference/supply-chain-security/dependabot-on-actions) omits. Its single exception is a pull request whose *base* branch Dependabot created, which a pull request against `main` cannot be.

One detail of the finding does not survive checking, and it matters only because it would have pushed the fix toward a two-workflow `workflow_run` writeback. `permissions` **can** raise a Dependabot `pull_request` token above read-only — GitHub's own [auto-merge recipe](https://raw.githubusercontent.com/github/docs/main/content/code-security/tutorials/secure-your-dependencies/automate-dependabot-with-actions.md) is `on: pull_request` with `permissions: contents: write`, and the downgrade rule in `workflow-syntax` is scoped to pull-request events *from a forked repository*. So the choice here was between two workable events rather than between a broken one and a rebuild, and `pull_request_target` wins on its own merits: it is exempt outright, the fence is read from base-branch content so a head cannot edit the condition guarding it, and the workflow file comes from the base branch — a Dependabot pull request opened before this merges is covered on its next event instead of needing a rebase to pull the file onto its own branch. Two workflows plus an artifact handoff would have bought none of that and collided with P2.

What the event costs is a head checkout beside a privileged token, and the mitigations are named where they live: the guard admits only a Dependabot-authored branch of this repository, `--lockfile-only` unpacks nothing so no bumped code runs, and `allow-unsafe-pr-checkout` stays unset on purpose — `actions/checkout` refuses a fork head under this event without it, which keeps a second line behind the job's own condition.

**[P2 OVER-BUILT] the commit.** The blob → tree → commit → ref sequence is gone. `git commit -- bun.lock` carries that file and nothing else, and `git push origin` is refused by the server both for a branch that moved and for a branch a protection rule covers, which is what the API sequence was reimplementing by hand. The step is 8 lines where it was 22.

The one thing the API was buying that plain git does not is the author: `git` needs an identity, and this repository's privacy gate rejects any address literal, including the bot's. So the address is assembled from the bot's own user record and `GITHUB_SERVER_URL` — two lines, no literal, and a run on GitHub Enterprise attributes to that server's bot rather than to github.com's.

### Evidence

The commit-and-push step was extracted from the YAML and run **verbatim** against a local bare repository seeded with a Dependabot-shaped branch, from a genuinely shallow (`--depth 1`) clone, which is what the workflow's `fetch-depth: 1` produces.

| Case | Result |
| --- | --- |
| Branch unchanged | Push accepted from the shallow clone; remote head is `chore(deps): refresh bun.lock`, author and committer both the Actions bot at its standard no-reply address, commit contains `bun.lock` alone — an untracked file left in the tree was not swept in |
| Branch force-pushed mid-run | Push rejected (`fetch first`), `::notice::` naming the branch, exit 0, remote head unchanged |
| Server refuses for another reason (protection-shaped `pre-receive` refusal) | Exit 1, remote head unchanged — a protection refusal is not swallowed as a race |
| Protection step, `protected: true` / `false` / unreadable | Exit 1 with an `::error::` naming the branch / exit 0 / exit 1 |
| Protection step against #1178's real head branch | `protected=false`; the live Dependabot branch is admitted |
| Dispatch step, all three accept | Three `gh workflow run` calls in order, the tracker one carrying `publication_number` |
| Dispatch step, one workflow's copy on the branch predates the trigger | `::warning::` naming it, the other two still dispatched, exit 0 |
| `actions/checkout` v7 at the pinned SHA | Its fork refusal fires only when the head repository id differs from the base's, so leaving the opt-in unset costs this job nothing and keeps the refusal armed |
| `python -c 'import yaml'` over all five workflows | Parse; `on:`, job ids and job-scoped permissions read back |
| `bash -n` over all 26 `run:` blocks in the repository | Clean |
| `git diff --check` | Clean |
| `bun scripts/privacy-publication-gate.ts --base <merge-base> --check-commits` | `PRIVACY GATE: PASS` |

Scope is unchanged: `.github/workflows/**` and `docs/RELEASING.md`, nothing under `src/`, no TypeScript, so `tsc` still does not apply. `docs/RELEASING.md` now explains the event, the ordinary push, and what a dispatch does on a branch that predates this.

Two notes stand from earlier rounds, and one is new:

- The workflow still cannot be triggered against #1178 before merge — a `pull_request_target` run executes the base-branch copy, and `main` does not carry the file yet. After merge, reopening #1178 raises `reopened` and runs it, with no Dependabot cooperation needed.
- #1178 remains blocked on `privacy-publication` for a reason outside this fence, filed as #1195.
- **New:** this PR can no longer show `refresh-lockfile` as *skipping*. That evidence came from the workflow running on `pull_request`, so this branch ran its own copy; under `pull_request_target` the run comes from the base branch and no such check appears here. The property replacing it is the stronger one — the condition can no longer be edited by the head it fences.

PR #1208, opened 2026-08-27
ci(privacy): let the base-ref assertion accept the dispatch fallback

Latand:
## Problem

`privacy-publication` is failing on every open pull request (#1203, #1204, #1205, #1206), so nothing can merge. The failure is not a privacy hit — it is the gate's own test suite:

```
(fail) privacy publication gate > resolves the current protected base tip for long-lived pull requests
1494 | expect(workflow).toContain("PRIVACY_BASE_REF: ${{ github.event.pull_request.base.ref }}");
```

## Cause

`ac197031` (PR #1196) gave the workflow a `workflow_dispatch` default:

```yaml
PRIVACY_BASE_REF: ${{ github.event_name == 'workflow_dispatch' && github.event.repository.default_branch || github.event.pull_request.base.ref }}
```

The behaviour the test guards is unchanged — the base ref still comes from the pull request's base branch and is still resolved to that branch's current tip — but the assertion pinned the whole env line as an exact substring, so it went stale the moment the fallback landed.

## Change

Assert the guarantee instead of the literal line: `PRIVACY_BASE_REF` must be an expression that reads `github.event.pull_request.base.ref`. The `base.sha` prohibition, the fetch line and the `rev-parse` line are untouched, so the "current tip, never the frozen base sha" contract still fails loudly if it regresses.

## Verification

`bun test scripts/privacy-publication-gate.test.ts` under an isolated `HOME`/`XDG_CONFIG_HOME`/`LLV_STATE_DIR`/`TMPDIR`: **116 pass, 0 fail** (was 115 pass / 1 fail).

### DIFF
--- a/scripts/privacy-publication-gate.test.ts
+++ b/scripts/privacy-publication-gate.test.ts
@@ -1491,7 +1491,9 @@ exec "$LLV_TEST_REAL_GIT" "$@"
     const workflow = readFileSync(join(import.meta.dir, "..", ".github", "workflows", "privacy-publication.yml"), "utf8");
 
     expect(workflow).not.toContain("github.event.pull_request.base.sha");
-    expect(workflow).toContain("PRIVACY_BASE_REF: ${{ github.event.pull_request.base.ref }}");
+    expect(workflow).toMatch(
+      /PRIVACY_BASE_REF: \$\{\{[^}]*github\.event\.pull_request\.base\.ref[^}]*\}\}/,
+    );
     expect(workflow).toContain(
       'git fetch --no-tags --force origin "+refs/heads/${PRIVACY_BASE_REF}:refs/remotes/origin/${PRIVACY_BASE_REF}"',
     );


Latand:
Merging this with an administrator bypass of the `privacy-publication` check, and recording why.

That job's `Verify privacy gate behavior` step runs `bun test scripts/privacy-publication-gate.test.ts` in the **trusted** checkout — `ref: github.event.repository.default_branch` — and the test reads the workflow file next to itself, i.e. `main`'s copy of both. The stale assertion therefore fails against `main`'s workflow no matter what a pull request contains, and no pull request can turn the check green, including this one. The fix has to be on `main` before the check can pass anywhere.

Evidence it is the only failure and that the fix is correct: `bun test scripts/privacy-publication-gate.test.ts` on this branch under an isolated `HOME`/`XDG_CONFIG_HOME`/`LLV_STATE_DIR`/`TMPDIR` gives 116 pass / 0 fail, against 115 pass / 1 fail on `main`. The other checks on this PR (`audit`, `privacy-tracker-audit`) are green, and the gate's actual scanning step — `Scan changed publication surfaces and commit messages` — is not what failed.

Once this is on `main`, re-running `privacy-publication` on #1203, #1204, #1205 and #1206 picks up the fixed trusted copy.

Issue #1246, opened 2026-08-28
P0 fix(deps): every candidate answers HTTP 500 after the dependency bump, so production cannot deploy


Latand:
## Production cannot deploy

Every deployment attempted after the dependency bump landed fails the same way: the candidate container starts, and then answers **HTTP 500 on every route**. Production is pinned and has not advanced since.

## Evidence

Three consecutive failed deployments, read from the durable deployment records. In each, the candidate's own health evidence is identical:

```
rootStatus: 500, authenticatedStatus: 500, unauthorizedStatus: 500, assets: 0
detail: "candidate readiness timed out: Viewer deployment capability gate failed"
```

`assets: 0` matters: the probe derives the asset list from the returned HTML, and a 500 page references none. So this is not a partial failure of one route — the application is not serving at all.

## Timeline

| When (UTC) | Revision | Outcome |
| --- | --- | --- |
| 16:39 | last good | **succeeded** — the last revision before the bump |
| 21:59 | the bump merges | — |
| 23:37 | bump + TypeScript re-pin | failed, candidate 500 |
| 00:46 | one commit later | failed, candidate 500 |
| 00:56 | same revision, clean retry | failed, candidate 500 |

Two of these are the same revision deployed twice, ten minutes apart, with the same result — this reproduces, it is not a flake or a leftover from an interrupted release switch.

## What the bump changed

- `next` 16.2.10 → 16.3.2
- `react` / `react-dom` 19.2.4 → 19.2.8
- `@modelcontextprotocol/sdk` 1.29.0 → 1.30.0
- `entities` ^7 → ^8 (major)
- `@types/node` ^20 → ^26 (major)
- `eslint` ^9 → ^10 (major)
- `typescript` ^5 → ^7, already reverted separately because the image stopped building

The framework minor is the first thing to examine. `AGENTS.md` opens by warning that this Next version carries breaking changes against common assumptions and that its own docs under `node_modules/next/dist/docs/` are the reference. A minor bump of it reached the default branch through automation without any deploy ever exercising it.

The TypeScript re-pin is evidence in itself: the bump had already broken the image build once and was repaired at the build layer only. The runtime was never verified.

## Two defects, not one

**1. The bump breaks the running application.** Root cause unknown until the candidate's own logs are read; the candidate container and image are both removed when a deployment fails, so this needs either a retained artifact or a local reproduction under the image's runtime.

[...]

Issue #1241, opened 2026-08-28
Re-take the npm bump once the image's Bun can load Next 16.3's app-page runtime


Latand:
Split out of #790 / PR #1240.

The candidate built from `9fc92c90` answered HTTP 500 to every request and could not be promoted. Cause: the image serves with `bun --bun node_modules/.bin/next start` (`Dockerfile:250`) on the Bun the image pins (1.3.3), and that Bun cannot load Next 16.3.2's compiled app-page runtime.

```
bun 1.3.3, next 16.3.2
  require next/dist/compiled/next-server/server.runtime.prod.js     -> loads
  require next/dist/compiled/next-server/app-route.runtime.prod.js  -> loads
  require next/dist/compiled/next-server/app-page.runtime.prod.js   -> TypeError: Expected CommonJS module
                                                                        to have a function wrapper
bun 1.3.3, next 16.2.10 -> loads
bun 1.4.0, next 16.3.2  -> loads
```

Every built server module reaches that runtime through the shared chunk graph, so the server binds its port and then every route module throws while loading. `bun run build` fails for the same reason on a machine whose only node is Bun's shim.

PR #1240 pins `next` and `eslint-config-next` back to 16.2.10 to unblock shipping. That reverts one part of the npm bump. Two things are left:

1. Move the image's Bun forward (three `npm install -g bun@1.3.3` lines in `Dockerfile`), verify a candidate serves, then re-take `next` 16.3.x. Verification must include loading `.next/server/app/page.js` and one API route module under the image's Bun, since a build that succeeds under node says nothing about a server that runs under Bun.
2. `bun run lint` is broken by the `eslint` ^9 -> ^10 half of the same bump: every file fails with `Error while loading rule 'react/display-name': contextOrFilename.getFilename is not a function` from `eslint-plugin-react`. Either move that plugin forward or pin eslint back.

Acceptance: a deployment of a revision carrying the re-taken bump reaches `succeeded`, and `bun run lint` runs to completion.

Latand:
Closing per backlog triage #1353: superseded by work shipped on main — see the evidence row in #1353.

Issue #1221, opened 2026-08-27
Orchestrator board — 2026-08-27: what shipped, what is in flight, what is queued


Latand:
Durable record of what this orchestrator seat is driving, because until now the plan lived only in a session-scoped timer and would have died with the session.

## Shipped today

| PR | What it fixes |
| --- | --- |
| #1203 | Agent sessions dialog lists structured hosts and kills them, bound to a re-verified process identity (#1199) |
| #1204 | `docs/orchestrator.md`, README section, seeded dock shot (#1160, #1171) |
| #1205 | Scheduled Telegram report spawns were refused at admission since the semantic-title rule; also a stale mobile fixture (#1188, #1198) |
| #1206 | `suggest_replies` reply-draft pills that fill the composer (#1202) |
| #1208 | The privacy gate's own self-test was red on every PR, freezing every merge in the repository |
| #1209 | Dependabot sign-off trailer, and publish verification that raced registry propagation (#1195, #1159) |
| #1210 | Dock re-resolution retries while binding is pending; mandate card honours a stored version (#1189, #1190) |
| #1212 | `close` retires an attempt whose turn a provider limit ended, so a dead lane can leave the board (#1141) |
| #1220 | The trailer bound shipped in #1209 was stricter than git's — 100 commits on `main` flagged, 99 of them false |

## In flight

| Issue | Lane | PR | State |
| --- | --- | --- | --- |
| #1216 | `48b77bca` | #1219 | P0. Promote times out on hot-state activation and rolls back; **production cannot advance**, and everything above is waiting behind it |
| #1213 | `0308036b` | #1217 | A delivery ending in `delivery-uncertain` shows "Delivering" forever, with no signal and no recourse |
| #1214 | `74bb548a` | #1218 | The compact control sends `/compact` on the Claude structured path, per the operator's decision |

## Queued, nobody on it

- **#1207** — OpenClaw as a third engine. Design written and twice reviewed; the standing decision is to cut hosting into a Deferred section and ship phase 1 (viewing) alone. Lane parked.
- **#747** — retire finished and idle structured hosts automatically. Today 65 live hosts were counted, 15 of them belonging to closed lanes, reaped by hand.
- **#1184** — a host resumed after a viewer restart idles forever on an interrupted tool call while the board shows it as throttled.
- **#301** — paginated Codex resume (PR #886, long-running).
- **#1201** — native Windows support. Design stage only.

## Two environment failures that shaped the day

[...]

PR #1253, opened 2026-08-28
chore(deps): restore next and eslint-config-next 16.3.2 on Bun 1.4.0, and refuse an undeclared dependency change

### LINKED ISSUE #1248: chore(deps): restore Next 16.3.2 on Bun 1.4.0, and stop a merge from silently reverting a dependency group
## 1. Restore Next 16.3.2 by moving the image to Bun 1.4.0

#1246 was closed by pinning the framework back to 16.2.10, which unblocked production. That pin is a hold, not a destination — it leaves the repository a minor version behind and will collide with the next dependency group.

The incompatibility is precisely bounded, which makes the forward path clear:

| | Next 16.2.10 | Next 16.3.2 |
| --- | --- | --- |
| Bun 1.3.3 (current image) | loads | `Expected CommonJS module to have a function wrapper` |
| Bun 1.4.0 | — | loads |

Only `app-page.runtime.prod.js` is affected; `server.runtime.prod.js` and `app-route.runtime.prod.js` load under both.

The work: move the image's three Bun pins to 1.4.0, exercise the whole served runtime under it, then restore Next 16.3.2 and `eslint-config-next` to match. This is a wider runtime change than the pin it replaces, so it deserves its own verification rather than riding along with a dependency bump.

**Acceptance:** a deployment reaches `succeeded` with Bun 1.4.0 and Next 16.3.2, and the verification exercises the *served* application — a green build does not count, which is the whole lesson of #1246.

## 2. A merge can silently revert a dependency group

The pin that fixed #1246 was not chosen by anyone. A pull request branched before the dependency bump had its branch updated onto the default branch, and the merge resolved `package.json` and the lockfile to the branch's older side. Nobody reviewed that revert, and the pull request's own description said nothing about dependencies.

Here the accident happened to be exactly right. The same mechanism could just as easily undo a security update, and nothing in the process would surface it: the diff that reverts belongs to a pull request about something else entirely.

**Acceptance:** a pull request that changes `package.json` dependencies or the lockfile without saying so is surfaced before merge — the shape of the guard is open (a check that compares the merge result's dependency set against the base, a label, or a required note), but silence must stop being a valid outcome.

Latand:
Closes #1248
Refs #1241 — this is item 1 of that issue (move the image's Bun forward, then re-take Next 16.3.x, verified by loading the built server modules under the image's Bun). Item 2 there, `bun run lint`, is deliberately untouched here; see the gates section.

Two independent commits, in the issue's order, so either can be reverted alone.

## 1 — `next` / `eslint-config-next` back to 16.3.2, the image's Bun to 1.4.0

The image's three `npm install -g bun@…` pins move to 1.4.0; `next` and `eslint-config-next` go back to 16.3.2. Every other member of the dependency group stays where it is — React 19.2.8, MCP SDK 1.30.0, `entities` 8, Node types 26, ESLint 10, TypeScript 5. The lockfile movement is entirely a consequence of 16.3.2's own ranges: every changed entry is `next` itself, its `@next/*` platform packages, `eslint-config-next`, or something Next pulls — `sharp` 0.34.5 → 0.35.4 with its `@img/*` binaries, `@swc/helpers` 0.5.15 → 0.5.23, and Next's nested `postcss` 8.4.31 → 8.5.23. No unrelated package drifted.

### Runtime verification — exactly what was run

A green build is what missed this the first time, so the build is not the evidence. Everything below was run on a clean `.next` built from this revision under Bun 1.4.0.

**1. The compiled framework bundles, `require()`d under both Bun versions.**

| | `server.runtime.prod.js` | `app-route.runtime.prod.js` | `app-page.runtime.prod.js` |
| --- | --- | --- | --- |
| Bun 1.3.3 + Next 16.3.2 | loads | loads | **`Expected CommonJS module to have a function wrapper`** |
| Bun 1.4.0 + Next 16.3.2 | loads | loads | loads |

**2. The built `.next` server modules, same artifacts, two runtimes.** The probe first installs `globalThis.AsyncLocalStorage`, which the Next server itself installs before loading any server module — without it a bare `require()` stops on Next's own runtime-context invariant instead of on module format, so the probe would not be measuring loading.

```
bun 1.3.3   FAILS  .next/server/app/page.js                                          Expected CommonJS module to have a function wrapper
            FAILS  .next/server/app/api/workflows/route.js                           (same)
            FAILS  .next/server/app/api/runtime/deployments/capabilities/v1/route.js (same)
            FAILS  .next/server/middleware.js                                        (same)
            FAILS  next/dist/compiled/next-server/app-page.runtime.prod.js           (same)
bun 1.4.0   LOADS  all five
```

**3. The served application, started with the image's own command** — `bun --bun node_modules/.bin/next start`, isolated `HOME`/`XDG_CONFIG_HOME`/`LLV_STATE_DIR`/`TMPDIR`, loopback port. Under **Bun 1.4.0**:

- `GET /` → **200**, 16 259 bytes of real HTML (`<title>Agent Log Viewer</title>`).
- `GET /api/runtime/deployments/capabilities/v1` → **200**; `GET /api/workflows` → **200**.
- All 8 `/_next/static/**` assets the served HTML references → **200** each.
- `GET /api/timeline` → 400 `{"error":"project required"}` — the route's own contract, which only a loaded route module can produce.
- Zero `Expected CommonJS` occurrences in the server log.

**4. The counterfactual, which is what makes 3 mean something.** The *same build*, the *same start command*, only the Bun version changed — served under **Bun 1.3.3**:

```
500  /
500  /api/runtime/deployments/capabilities/v1
500  /api/workflows
Expected-CommonJS occurrences in the server log: 4
```

That is the production failure of #1246 reproduced end to end: the server binds its port, reports `✓ Ready`, and then every route module throws while loading.

**5. The same thing from inside the image.** `docker build --target build` — `node:22.16.0-bookworm-slim`, `npm install -g bun@1.4.0`, `bun install --frozen-lockfile`, `bun run build` exactly as the Dockerfile does — then served from that container over a mapped loopback port. In-image `bun --version` → `1.4.0`; `/` **200**, capability route **200**, `/api/workflows` **200**, all 8 referenced assets **200**, zero `Expected CommonJS` in the container log, and the built page, route, middleware and `app-page.runtime.prod.js` all load under the in-image Bun.

### What else the runtime change touches — checked, not assumed

- **Worker sidecars.** They run under the in-image Bun via `bun-container`, from TypeScript source with the built `.next/server/*-worker.js` as fallback. All four stdin-driven workers — file scanner, files response, account-migration controller, resource collector — load and reach their own input handler (`… worker received an invalid request`) in **both** forms and on **both** Bun versions. They never pull in the app-page runtime, so they were never part of this failure. The wakatime sidecar was left out of the load probe because it starts a sync loop rather than reading stdin.
- **Container entrypoints.** The runtime host (`src/runtime-host/main.ts`) and the proxy (`src/proxy.ts`) both resolve and parse their full import graph under Bun 1.4.0, and the shipped `dist/mcp-server.mjs` imports cleanly under it. Resolve-and-parse only — nothing was executed, because starting a runtime host from a lane is not a thing to do.
- **The lockfile across both Bun versions.** `bun install --frozen-lockfile` is accepted by **both** 1.3.3 and 1.4.0 against this lockfile, leaving `bun.lock` and `package.json` untouched. CI cannot break on the lockfile the 1.4.0 image produces.
- **CI stays on Bun 1.3.3, deliberately.** All four workflows pin 1.3.3, and the privacy provenance chain is pinned to it by construction: `scripts/generate-privacy-placeholders.ts` refuses any `Bun.version` other than `1.3.3`, and the gate only accepts manifests declaring `bun-1.3.3`. Moving those workflows would break placeholder regeneration for no gain — CI never serves the application, which is the only place the incompatibility lives.
- **The published package.** `bunx agent-log-viewer` starts the viewer with `bun --bun`, so a user on Bun 1.3.x would meet this exact failure. `engines.bun` moves `>=1.2.0` → `>=1.4.0`. It is declarative; Bun 1.3.3 installs the manifest without complaint, which is what CI still does.
- **The pin assertion in `scripts/dockerfile-permissions.test.ts`** now requires three `bun@1.4.0` pins and refuses `bun@1.3.3`, so a silent slide back to the broken pairing fails a test.

**No deployment, promote, rollback or restart was run** — the acceptance in the issue is the deployment that follows this, not anything from this lane.

## 2 — a merge cannot silently change dependency versions

`supply-chain.yml` already compared the merge result against the base (under `pull_request` the checkout is the merge ref, so `HEAD` is what would land), but it only asked that `bun.lock` move alongside `package.json` — which a revert of both satisfies. It now also asks that somebody said so: **every package whose version the merge would change must be named in the pull request's own title or body.**

Aimed at the case that bit, not at ordinary bumps: a real dependency change names its packages already (Dependabot lists each one), so the gate is quiet for changes people make on purpose and fires on the one nobody wrote down. When it fires it prints the package, both versions and the section, so the fix is one sentence — or dropping a revert the branch never meant to carry.

The rule moved out of the inline `bun -e` block into `scripts/dependency-declaration-gate.ts` so there is one tested definition of a dependency change; the #1156 lockfile rule is carried over unchanged, and a dispatched run (no pull request payload) keeps it while skipping the declaration half — and no longer prints a success line for the half it skipped.

Its test builds the accident out of real commits — a branch predating a bump, updated onto the default branch with `package.json` and `bun.lock` resolved to its own side, then merged as GitHub merges it — and asserts the gate refuses that merge naming `next 16.3.2 -> 16.2.10`, and accepts the identical merge once the description declares it.

One boundary, stated rather than left implicit: the declaration half reads `package.json`, so a merge that reverts a *lockfile-only* refresh — the kind `dependabot-lockfile.yml` commits, where no manifest version moves — is covered by the #1156 lockfile rule alone and not by the declaration rule. Comparing resolved lockfile *content* instead would fire on every dependency pull request with hundreds of transitive entries, which is the noise the issue says not to build.

Run live against this branch's merge into `main`, it behaves the same way:

```
description that says nothing   exit 1   next 16.2.10 -> 16.3.2 in dependencies,
                                         eslint-config-next 16.2.10 -> 16.3.2 in devDependencies
this pull request's own body    exit 0   Every changed package is named in the pull request description
no pull request payload         exit 0   declaration rule not applied; lockfile rule still enforced
```

This pull request is its first live subject, and the `audit` check on it is the gate running on itself.

## Gates

- `bunx tsc --noEmit --incremental false` — exit 0, no output.
- `bun test` by path under an isolated `HOME`/`XDG_CONFIG_HOME`/`LLV_STATE_DIR`/`TMPDIR`: `scripts/dependency-declaration-gate.test.ts`, `scripts/dockerfile-permissions.test.ts`, `src/proxy.test.ts` — **23 pass, 0 fail** under **both** Bun 1.3.3 and Bun 1.4.0, since the framework moved a minor under both runtimes.
- `bun scripts/privacy-publication-gate.ts --base <merge-base> --check-commits` — **PASS** (exit 0), including with the tracked fingerprint catalog wired in via `LLV_PRIVACY_KNOWN_VALUE_FINGERPRINTS_FILE`.
- `bun run lint` is broken, and this branch is not the cause — verified rather than assumed. Run in a throwaway worktree checked out at the merge base (`next` 16.2.10, `eslint-config-next` 16.2.10, deps installed from that revision's lockfile) it fails identically: exit 2, `Error while loading rule 'react/display-name': contextOrFilename.getFilename is not a function`. The cause is the `eslint` ^9 → ^10 half of the earlier bump against `eslint-plugin-react` 7.37.5, and this branch changes neither: `eslint` is `^10` on both sides and `eslint-plugin-react` resolves to 7.37.5 on both sides. Fixing it means moving a dependency the issue says to leave alone, so it stays where it is and stays tracked as item 2 of #1241. No workflow runs `lint`.


Latand:
## Scope cut: the dependency-declaration gate is reverted, the runtime restore stays

The second commit is reverted. Its premise was mine and it was wrong.

I reported that the pin back to Next 16.2.10 arrived by accident — that a branch updated onto the default branch had resolved `package.json` and the lockfile to its older side, with nobody choosing it. This review's third finding contradicted that, and the record agrees with the review: the pin is a deliberate, declared commit in PR #1240, `fix(deps): pin Next back to 16.2.10 so the candidate can serve again`, whose body diagnoses the Bun/Next loader mismatch and explains the change. It was reviewed and it was intentional.

So the guard defends against an incident that never happened. Its own review also found it refused ordinary declared bumps (`bumps next.`, `Upgrade Next.js to 16.3.2`) and printed a fix that could not clear the check, because editing a description does not re-trigger the workflow. A gate built on a false premise, which then fails on the ordinary case, is not worth repairing.

Reverted rather than force-pushed, so the attempt and this reasoning stay in the history.

## What remains, unchanged

Commit 1: `next` and `eslint-config-next` restored to 16.3.2 with the image's three Bun pins moved to 1.4.0. The review found no findings against it and called it ready, having reproduced the verification independently:

| | Bun 1.3.3 | Bun 1.4.0 |
| --- | --- | --- |
| `require()` of `app-page.runtime.prod.js` and the built routes | all fail, `Expected CommonJS module to have a function wrapper` | all load |
| `bun --bun next start`, then `GET /` | 500 | 200, 8/8 referenced assets 200 |
| inside the image build stage | — | `/` 200, capability 200, all seven modules load |

That is the bar this issue set: a green build does not count, because a green build is exactly what shipped the outage.

Corrections have been posted to #1246 and #1221, which carried the same mistaken account.

Issue #1228, opened 2026-08-27
test(telegram): the Daily Report durable-marker render test fails on main — a spawn now requires a title


Latand:
## Symptom

`src/components/TelegramReports.render.test.tsx` → "after a reload, a run the history row cannot name is still linked from the durable marker" fails on `main` (verified at the current default-branch tip, with the file restored from `HEAD` so no local change is involved):

```
error: title is required for every new spawn
      at beginSpawnRequestInFile (src/lib/agent/registry.ts:3811)
      at <anonymous> (src/components/TelegramReports.render.test.tsx:284)
```

## Mechanism

The test builds the durable report-run receipt by calling `AgentRegistry.beginSpawnRequest` directly with a launch profile from `emptyLaunchProfile({ cwd })` — which carries no title. `beginSpawnRequestInFile` refuses a titleless spawn (`src/lib/agent/registry.ts:3811`, `durableSemanticTitle` / `SPAWN_TITLE_REQUIRED_ERROR`), a rule introduced by #916. The Daily Report's real launch path always supplies one (`reportConversationTitle`), so the production behaviour the test is about is fine; the fixture is what predates the rule.

## Repro

```
bun test src/components/TelegramReports.render.test.tsx
```
→ 16 pass, 1 fail, with the trace above.

## Expected

The fixture gives its reservation a title (the same semantic title a report launch uses) so the test exercises the marker-relink behaviour it is named for, instead of failing before it gets there.

Found while working on #1133; unrelated to that change, which leaves this test exactly as it was.

Latand:
Same class, three more files red on the current default branch (verified at merge base `7d66f146` by swapping the pre-change sources back in — the failures are identical with and without an unrelated branch applied):

- `src/lib/delivery.test.ts` — 2 fail: `codex send follows verified tmux rollback ownership and creates one structured successor after termination`, and the `claude send …` twin. Both throw `title is required for every new spawn`.
- `src/lib/runtime/structuredSpawn.integration.test.ts` — 3 fail: `claude stale-live recovery converges registry and runtime on one durable conversation`, its `codex` twin, and `issue 367: a post-kill relaunch that exhausts admission never leaves a live queued placeholder`.
- `src/lib/runtime/legacyClaudeRecovery.cleanCi.integration.test.ts` — its single test fails with the same `title is required for every new spawn`.

Noted from the #611 lane, which had to baseline these before trusting its own run. Nothing in that lane touches spawn titles.

Issue #565, opened 2026-07-22
Structured spawn to a non-active account deadlocks: first message held forever behind a waiting-turn migration that can never start


Latand:
## Symptom

A fresh structured spawn targeted at a **non-active** managed account never delivers its first message. The board card sits forever on "Structured host ready. First message queued." (`spawnCard.queued`), the host process is alive and idle, the transcript file is never created, and the conversation shows nothing.

Observed live on prod 2026-07-22 ~08:09 UTC: launch `a2bae7b2-7ca7-4a45-8bf0-dafa9809ce08`, engine claude, account `botfatherdev-2`, conversation `conversation_aa94c250-7442-4e71-a6f8-cccc836ff031`, while the active claude account was `botfatherdev-last` with a still-draining manual migration intent from the previous day (`dadda37f-5b0e-4f0b-85ea-02c73af17548`).

## Root cause — three rules deadlock each other

1. **Intent sweep stamps the newborn conversation with `waiting-turn`.** When a conversation registers under an account other than the intent target, it gets a migration whose phase is picked by turn readiness: `turn.state === "busy" || turn.state === "unknown"` → `waiting-turn` (`src/lib/agent/registry.ts:3593`, also `:5104`, `:4768`, `:4817`). A brand-new spawn's turn is `{state: "unknown", source: "empty"}` — no message has ever been delivered — so it is stamped `waiting-turn`.

2. **Migration can't advance until the turn ends.** `successorCreationReady` requires `turn.state` ∈ {`terminal`, `idle`} (`src/lib/accounts/migration/coordinator.ts:243`). `unknown` never qualifies, and nothing will ever move it: the turn only changes when a message is delivered.

3. **Delivery is blocked by the migration.** `reserveHeldDelivery` keeps the spawn's initial message in `held` while the migration phase is in `["waiting-turn", "requested", "preparing", "successor-starting", "verifying"]` (`src/lib/agent/registry.ts:5339-5352`). Held → no delivery → no turn → migration waits forever → held forever.

The spawn receipt stays `path-pending`, the held delivery stays `held` with `attempts: 0`, and the #334 convergence pass doesn't help because the host is alive and healthy.

## Repro

1. Have engine routing active on account B with a draining migration intent (lazy sweep still open).
2. `POST /api/spawn` a claude structured launch with the account picker set to account A (≠ B).
[...]

Issue #344, opened 2026-07-17
Structured Claude continuation ignores the newly selected account


Latand:
## Problem

Selecting a fresh Claude account updates `engineRouting.claude.activeAccountId`, while existing structured conversations retain the original generation `accountId`. When the original account reaches its session limit, sending a follow-up or pressing resume keeps targeting that exhausted account. The visible result is `structured resume host claim is unavailable` and the conversation cannot continue.

Observed on 2026-07-17:

- active account: `botfatherdev-last` with fresh capacity
- affected conversations: `conversation_00000000-0000-0000-0000-000000000000` and `conversation_00000000-0000-0000-0000-000000000000`
- both generations remain bound to `accountId: default`
- `default` reports 100% session usage
- auto-balance remains idle because quota selection evaluates the current active route and does not hand off exhausted conversations owned by a previously active account

## Required behavior

1. A user-selected Claude account must be usable for the next turn of an idle or quota-terminated structured conversation.
2. Preserve the stable Viewer conversation identity, transcript history, board placement, lineage, title, pending deliveries, and idempotency ownership.
3. Keep a single visible conversation through the handoff.
4. Fence the old host before publishing the successor host.
5. Queue the triggering message during handoff and deliver it once through the selected account.
6. Surface an actionable terminal failure when authentication or history adoption fails.
7. Apply equivalent semantics to Codex where its provider permits continuation under a selected managed account.

## Acceptance

- Start a structured Claude conversation on account A.
- Exhaust or simulate quota exhaustion on A.
- Select authenticated account B.
- Send a follow-up from the existing conversation card.
- The same Viewer conversation continues under B and the message is delivered once.
- Restart during each handoff phase and prove recovery.
- Add focused runtime, registry, delivery, and account-routing tests.
- No UI changes are required for this issue.

Latand:
Production verification completed for the same-conversation selected-account rebind.

Release evidence:
- Commit: `7960dba5917c851f7dedeb3e4476cedffed17822`
- Deployment: `deploy_00000000-0000-0000-0000-000000000000`
- Serving container: `llv-deploy-7bed75dbc27e946d0fdd9e26`
- Runtime deployment phase: `succeeded`

Implemented contract:
- Dead Claude structured recovery resolves the currently selected account.
- Claude resumes through the registered absolute source transcript with the selected account credentials.
- Viewer conversation ID, native session ID, transcript path, lineage, and card identity remain stable.
[...]

PR #1277, opened 2026-08-29
fix(flows): publish-readiness consults provider throttle, and a parked relay is held visibly (#611)

### LINKED ISSUE #611: Same-account continuation stalls queued when a builder is parked at a quota-warning/prompt during review relay (forces manual preserve-commit + pipeline restart)

Latand:
Closes #611.

## The defect, where triage left it

A structured host that is process-alive and claim-owned was treated as
publish-ready **without consulting provider throttle state**
(`src/lib/runtime/structuredRecovery.ts`). Recovery handed that same host back
with `spawned:false`, and the review relay enqueued the round's findings into it
(`src/lib/flows/engine.ts`). When the builder was parked behind a provider
limit, the continuation was queued against a host that could not start a turn:
it stayed `queued`, the relay of reviewer findings stopped, and the lane went on
looking alive. Nothing errored, which is why it cost a preserve-commit, a fresh
pipeline and a reviewer re-attach three times in one evening.

## What "parked" is, from the runtime's own state

New `src/lib/runtime/accountPark.ts` answers one question — *is this account
parked, and until when* — from two readings the Viewer already keeps for its own
lifecycle surfaces, never from transcript prose:

- the newest **limits provenance** for the account: a provider rejection with
  the retry deadline it asked for (the signal liveness already projects); and
- the **durable quota observation** the account controller writes into the
  registry: a governing window at 100% with a reset still ahead — the state a
  builder sits in at a quota-warning prompt.

Two deliberate refusals. An exhaustion with **no knowable reset** is not
reported as a park: a wait nobody can time out is the same unwatched stall in a
different coat. A **stale** reading is not evidence — the freshness bar is the
one auto-balance already uses, so a cold cache never invents a hold.

`candidateFor` computes readiness as `hostLive && !park`, and recovery returns a
parked live host **held** rather than published:
`StructuredRecoveryResult.hold` carries the reason, the account and the
deadline.

## Held, not reseated — and why

The park belongs to the **account**, not to the host. A successor spawned for
the same account starts parked too, so a reseat that stays on the account buys
nothing while terminating the live host throws away the builder's continuation
— exactly the loss the manual workaround had to paper over with a preserve
commit. Moving the continuation to a *different* account is an account-migration
decision with its own machinery and its own hazards (#344, #565), and it is out
of this issue's scope. Meanwhile the park carries a bounded deadline from the
same state the board renders, so "wait, then deliver" terminates by
construction. Recovery therefore holds; an explicit ownership takeover (resume /
reconfigure) still reseats exactly as before.

## The relay holds instead of enqueueing

`sendToImplementer` refuses the enqueue **before any actuation** when recovery
reports a hold, so nothing is queued, no transport is recorded, and the round
keeps the durable client-message identity it would have been sent under — the
delivery that runs after the park lapses is the same one, never a duplicate.
The tick turns that into a hold on the round: `relayRetryAt` is set to the
provider's own deadline, and the ordinary relay wait resumes delivery when it
lapses.

Explicitly not done: no timeout is widened, no item is dropped to make delivery
"succeed", and the bounded retry budget and delivery-attempt generation are both
untouched — a three-hour park cannot burn a round into `needs_decision`.

Only the relay holds. The kickoff shares `sendToImplementer` but has no resume
schedule of its own, so it keeps the queue behaviour it has always had; the flag
that turns holding on is set by the relay tick, which owns the re-attempt.

## The stall is visible while it lasts

- `round.relayHold` — durable on the round (persisted through the flow store):
  reason, account, `until`, and when the wait started.
- `flow.stateDetail` — names the account and the deadline in prose.
- `flow.block` — the read model projects the block **from the flow's own durable
  hold, first**, so the strip and the loop hub show "rate limited until HH:MM"
  even when this scanner pass attributed no quota reading to the implementer's
  transcript. A held relay that showed nothing is what let three of these pass
  unnoticed.

## Verification

- `bun scripts/privacy-publication-gate.ts --base <merge-base> --check-commits` —
  PASS.
- Tests, by path, under an isolated `HOME`/`XDG_CONFIG_HOME`/`LLV_STATE_DIR`/
  `TMPDIR`:
  - `src/lib/runtime/accountPark.test.ts` — 5 pass (new)
  - `src/lib/runtime/structuredRecovery.test.ts` — 15 pass
  - `src/lib/flows/engine.test.ts` — 48 pass
  - `src/lib/rateLimit.test.ts` — 19 pass
  - `src/lib/flows/store.test.ts`, `store-isolation.test.ts`,
    `commands.test.ts`, `commands.durable.test.ts`,
    `src/components/flows/flowModel.test.ts`,
    `src/lib/runtime/structuredMessageDelivery.test.ts`,
    `structuredReconfigure.test.ts`, `src/lib/pipelines/engine.test.ts`,
    `src/components/SwitchCard.status.render.test.tsx` — all pass.
- **RED verified**: each new behavioural test was run against the pre-change
  source of the file it covers (swapped in from `HEAD`, restored after — no
  stash). Against the old code the relay enqueues into the parked host, recovery
  reports no hold, and the board projects no block.
- The suite this repo does not run on CI carries pre-existing failures on the
  merge base, unchanged by this diff — `src/lib/delivery.test.ts` (2),
  `src/lib/runtime/structuredSpawn.integration.test.ts` (3) and
  `src/lib/runtime/legacyClaudeRecovery.cleanCi.integration.test.ts` (1) all
  fail identically with the old sources swapped back in. All six are the
  "a spawn now requires a title" class already tracked in #1228.
- ESLint cannot run in this checkout at all (`eslint-plugin-react` vs
  ESLint 10 incompatibility), on untouched files too; not introduced here.

## Left alone deliberately

The flows engine's structure, the retirement predicate and the seat tick are
untouched, as are the two live lanes' files.


Latand:
Orchestrator note, 2026-08-29 — three review rounds deadlocked on one point, so this branch continues under a successor lane rather than a fourth identical round.

The deadlock, stated fairly, is that both sides were right about their own half:

- the reviewer held that an account reading exhausted with **no knowable reset** still leaves the live owned host publish-ready, so the relay enqueues into it and #611 reproduces exactly;
- the author held that an exhaustion nobody can time out must not be called a park, because an unbounded wait is the same stall in different wording.

Neither took the third option, which the reviewer's own acceptance criteria had already named: for that state the host is **neither publish-ready nor held indefinitely**. It gets a bounded recheck or an explicit decision state, visible, saying what it waits on and that the reset is unknown — with the host, the claim, the builder's work, the idempotent relay identity and the reviewer attachment all preserved, no retry budget spent, and exactly one delivery once usable state arrives.

The successor lane implements that on this branch, together with the second finding: publish readiness currently inherits the liveness-only grace after the retry deadline, so a probe just past it returns a hold whose deadline is already in the past and every tick re-holds until the grace expires. That grace belongs to liveness; readiness resumes at the provider's deadline.

PR #1278, opened 2026-08-29
fix(monitor): a hidden draft stops parking seats, and a project's tick is the project's to govern (#1274, #1275)

### LINKED ISSUE #1274: P1: a deleted never-run draft still parks its seat, waking it hourly with no way out but starting it
## A deleted draft keeps waking its seat, hourly, forever

Reported from another project's orchestrator seat and **independently confirmed here** by reading the record.

A draft pipeline that never ran has woken its seat roughly hourly since 2026-08-28 12:28Z — fourteen times at the point it was reported, more since. Every wake reaches the same conclusion and produces nothing. The seat looks busy and does no work, which is the worst shape of failure: it is invisible until someone notices the absence of output.

## The record, read back rather than trusting the action echo

```
state           : draft
closedAt        : null
hiddenAt        : 2026-08-28T15:01:15Z
publishedCommit : null
stages          : 1
attempts total  : 0
```

`hiddenAt` is set. `state` is still `draft` and `closedAt` is still null — so the parked/stalled detector reads it as a live pipeline and parks it, every tick.

## What the API does today

Each verified by reading the record back afterwards:

| action | result |
| --- | --- |
| `close` | `ok:true`, sets `hiddenAt` **only**; `state` stays `draft`, `closedAt` stays null |
| `delete` ×3, distinct request ids | identical — `hiddenAt` refreshes, nothing else changes, stages and runs remain |
| `pause` | refused: *draft pipelines can only be started, edited, or deleted* |
| `skip-stage` | refused: *pipeline does not have a stage awaiting a decision* |
| `remove-stage` | works down to one stage, then refused: *a pipeline keeps at least one stage* |
| `update-draft {task}` | accepted — title changes, branch renames |
| `update-draft {stages}` | refused: *update-draft needs at least one field to change* — stages are not editable |
| `start` | the only verb that clears the state, and the one that must never be used |

Retitling was tried and disproved: after the rename the next tick carried the **new** title and still reported the pipeline parked. The detector reads the record, so cosmetic edits cannot quiet it.

That leaves no safe exit. The only action that would end the loop is the one that must never be taken, because this particular draft's stage carries a destructive brief whose work was already delivered elsewhere — starting it would send a builder to delete already-deleted code and rebuild an already-merged model.

## Why this is not one project's problem

The pipeline belongs to another project, and it has been visible in **this** project's active-lane listing all day as well, because a listing that filters on `closedAt == null` includes it. So a single undeletable draft pollutes every seat's view of what is in flight, not only its owner's.

## Fix

Either is sufficient and either ends the class permanently:

1. **`delete` on a never-run draft actually terminates it** — set `state` and `closedAt`, or drop the record. A draft with zero attempts and no published commit has nothing to preserve.
2. **The parked/stalled detector ignores pipelines with `hiddenAt` set.** Hiding something and then continuing to report it as work is a contradiction on its own terms.

## Secondary, lower priority

Allow editing stages on a draft. An obsolete stage that cannot be deleted currently sits there with a live destructive brief and no way to render it inert.

## Acceptance

- A never-run draft, once deleted, stops appearing in active-pipeline listings and stops waking any seat.
- A seat's wake reasons never cite a pipeline the operator has hidden.
- Whichever path is chosen, a draft that *is* real work still starts normally.

### LINKED ISSUE #1275: feat(monitor): a seat must be able to stop its own ticks, with a reason and a bound
## An agent cannot quiet the loop the Viewer arms for it

The operator's words, relayed from another orchestrator seat: **an agent should be able to stop this monitor when it decides to.**

Today it cannot. The two halves of that are worth stating together, because each is defensible alone and the pair is not:

- The tick contract delivered with every wake says: *"Do not schedule yourself. The Viewer ticks this seat; a self-scheduled monitor is refused practice."* That rule is correct and was introduced deliberately — a session-scheduled monitor dies with its session, and every rotation silently dropped it.
- There is no lever to quiet the loop the Viewer arms instead. The seat record carries `activatedAt, conversationId, designatedAt, engine, intent, mandate, model, path, predecessorConversationId, project, promptVersion, runtimeIdentityFrozen, seatEpoch, state`. A scan for any key matching tick, sched, interval, wake, cadence, poll or stall returns nothing, and there is no action to pause, throttle or disable ticking.

So an agent is forbidden from arming its own loop and has no way to stop the one arming it. Only the operator can intervene, and only by hand.

## Why it matters in practice

The failure this exposes is quiet rather than loud. A seat woken repeatedly on a condition it cannot discharge — see #1274, where an undeletable draft woke a seat fourteen times in fourteen hours — looks busy and produces nothing. Each wake costs a paid turn. The operator finds out by noticing an absence, which is the hardest thing to notice.

`update_task → blocked` is the stop for a *task*. There is no equivalent for the tick itself, so a seat that has correctly concluded "there is nothing here for me, and the reason will still be true next hour" has no way to say so.

## What is requested

An agent-callable way to suppress ticking for its own seat. Shape is open; what it must have:

- **A reason, in the agent's own words**, stored with the suppression. A tick that goes quiet without a recorded reason is indistinguishable from a broken tick, and that is worse than the noise it replaced.
- **Visibility on the board**, so a human sees that a seat is deliberately quiet and why, and can re-enable it.
- **A bound.** Either an explicit expiry (`tickSuppressedUntil`) or something that lapses when the seat's situation changes — suppression that outlives its reason recreates the problem in the opposite direction, and a permanently silent seat is a worse defect than a noisy one.
- **Scoped to the seat that calls it.** A seat must not be able to quiet another project's seat.

## What this is not

Not a cadence setting, and not a per-project configuration surface — the design for the tick deliberately excludes both, and this should not become one by the back door. It is a stop, in the same spirit as `update_task → blocked`: an explicit act with a reason, not a knob.

## Acceptance

- A seat can suppress its own ticking, with a reason, and the suppression is visible on the board.
- The suppression is bounded and the bound is recorded.
- A suppressed seat is distinguishable, in the interface, from a seat whose tick is broken.
- No seat can suppress another's.

Latand:
Two halves of one failure: a seat woken forever on something it could not discharge, and no way for it to say so.

## #1274 — a discarded draft terminates instead of parking its seat

`delete` on a draft set `hiddenAt` and nothing else. The board stopped drawing the lane; every reader that asks the record whether it is live kept answering yes, because `state` was still `draft` and `closedAt` still null. So the parked detector read a live pipeline and parked it at every tick, and the wake it produced named the one verb the operator must never press: `pause` and `skip-stage` refuse a draft, and `start` runs the stage that was just thrown away.

Both fixes from the issue are here, because they answer different halves of the acceptance:

1. **`delete` (and `close` on a draft, the same act) closes the record** — `state: "closed"`, `closedAt`, `hiddenAt`, and a state detail saying what happened. A draft has never run, so there is no host to tear down, nothing published and no attempt to preserve; the row and its archive path stay, so nothing is dropped.
2. **Hidden stops reading as live** in the two places that were reporting it as work: the seat tick's open-lane predicate, and the bounded list projection (so a hidden lane leaves every project's active listing and comes back under `includeClosed`). This is what retires the records discarded *before* this lands — including the one burning turns now, which is already hidden and would otherwise need a second `delete` call.

A draft that is real work still starts normally, asserted directly.

## #1275 — a project's seat tick is the project's to govern

Durable per-project settings (`state/seat-tick-settings.json`): whether the Viewer wakes that project's seat at all, and how often. A project with no row reads the defaults — the same hour the tick has always used — so nothing has to be configured for the tick to work, and every pre-existing seat-tick test passes unchanged as the evidence for that.

- **`seat_tick_settings`** is the MCP surface: a read with no change fields, a change with them, `project` defaulting to the caller's own. Naming another project is allowed rather than refused; the row, the board card and the payload all record who changed whose tick.
- **Off means off** for as long as the caller says. `untilMinutes` is an optional expiry that lapses back to the default — a convenience for whoever reads the board, not a leash.
- **A reason is required** whenever the settings leave the default, and it is what the board card shows. Restoring the default needs none.
- **Board visibility**: one card per project carries the setting, its reason and who set it while it stands, is kept in step when the setting changes, and is closed by the first check that reads the project back on its defaults.
- **Off is not broken**: a quiet project is still checked, still reads the board and still writes its journal line every check. Only the wake is suppressed.
- The wake brief that says *do not schedule yourself* now names the lever on the schedule the Viewer arms instead.

ADR 0001's limit 4 said the wake interval had no override and that a reason allowed to jump it would be a decision to reopen the ADR. That decision is recorded as an amendment in the ADR: still no environment override, no exempt reason kind, no reset on rotation — the project's own recorded number in place of a constant.

## Verification

- Touched tests, by path, under an isolated `HOME`/`XDG_CONFIG_HOME`/`LLV_STATE_DIR`/`TMPDIR`: `pipelines/engine.test.ts` (208), `pipelines/listProjection.test.ts` (22), `pipelines/store.test.ts`, `pipelines/visibility.test.ts`, `pipelines/taskBinding.test.ts`, `pipelines/controller.test.ts`, `pipelines/terminalReap.test.ts`, `api/pipelines/[id]/route.test.ts`, `monitor/seatTick.test.ts` (43), `monitor/seatTickSources.test.ts` (34), `monitor/seatTickController.test.ts` (37), `monitor/seatTickSettings.test.ts` (10, new), `monitor/report.test.ts`, `monitor/cards.test.ts`, `monitor/run.test.ts`, `monitor/classify.test.ts`, `monitor/evidence.test.ts`, `mcp/bindings.test.ts` (43), `mcp/server.test.ts` (31 pass / 2 fail, see below), `mcp/presentation.test.ts`, `mcp/toolAllowlist.test.ts`, `mcp/orchestratorTools.test.ts` — all green.
- The seat-tick hidden-lane assertion was verified RED against the old predicate before the fix, then restored.
- `mcp/server.test.ts` re-run on a second, busier pass: the two receipt-churn tests (`a pending mutation remains claimed while read receipts churn`, `mutations retain replay and conflict protection after read receipt churn and restart`) exceed bun's 5 s per-test timeout. Each drives 501 sequential durable receipt writes, so the verdict tracks disk speed rather than this branch: **the same two fail identically at the merge base**, checked one file per invocation in a throwaway worktree at `7d66f14`. The other 31 pass.
- Pre-existing failures on the merge base, unrelated and unchanged by this branch (verified in a throwaway worktree at the base): `src/app/api/tasks/route.test.ts` (reads a `pipelines.json` mirror that is not written) and `src/lib/mcp/writerConcurrency.test.ts` (needs a built `dist/mcp-server.mjs`). `bunx eslint` cannot run in this checkout at all — `eslint-plugin-react` fails on an untouched file too.
- Follow-up commit `restoring the default tick keeps nothing of the setting it ended`: the restore branch cleared the interval and the expiry but carried the old reason forward, against the comment above it, so the stored row and the tool's own payload named a reason for a tick that is ticking again. The effective reading already nulled it, so nothing on the board showed it. Fixed, with the assertion verified RED against the previous line and green after.
- No deployment, promotion, rollback or restart of anything of the operator's. No Windows platform files touched.

Closes #1274
Closes #1275


Latand:
Orchestrator note on the review verdict, 2026-08-29.

The last review round returned **WRONG-PREMISE / OVER-BUILT**: that this branch adds the cadence and per-project configuration surface #1275 expressly excludes, and should be cut back to bounded own-seat suppression.

That verdict was faithful to #1275's text — and that text is stale. The requester superseded the exclusion on 2026-08-29: he wants exactly this surface. In substance, his direction was that a seat has no way to say "I am finished here", so when work ends he simply stops replying and nothing records that it is over; the monitor should therefore carry a state the agent itself can turn off and back on, hold an additional prompt the agent writes, and fire from the scheduler rather than re-arming after every response. He called it critical. It is recorded on #1280, which also notes that this branch already delivers most of it.

So the verdict is set aside as anchored to a requirement that no longer stands, rather than overruled on its merits. A successor lane is re-reviewing this branch against the current direction, with the presence of cadence and per-project settings explicitly not a defect in itself. Nothing here needs re-implementing unless that review finds a real one.

The one piece #1280 still carries beyond this branch is the agent-authored prompt travelling with the wake.

Latand:
Re-review under the corrected frame is in, 2026-08-29.

The reframed reviewer explicitly did **not** repeat the earlier verdict: *"Cadence and per-project settings were assessed as required by the superseding direction. I found no separate OVER-BUILT issue reaching the finding threshold."* So the WRONG-PREMISE reading is closed on its merits, not merely set aside.

One real gap remains, and it is exactly the piece #1280 says is still outstanding:

> **P1 — the scheduler cannot deliver the requester's agent-authored additional prompt.** The settings and MCP schemas expose toggles, cadence, expiry and `reason` metadata only (`src/lib/monitor/seatTickSettings.ts:64`, `src/lib/mcp/server.ts:2144`); when the scheduler fires, the controller builds the wake with no configured prompt (`src/lib/monitor/seatTickController.ts:510`) and the message carries only fixed contract text and gathered items (`src/lib/monitor/report.ts:96`).

Acceptance the reviewer set: a bounded, redacted, replaceable and clearable prompt field, appended to later scheduler-fired wakes, with current behaviour preserved when it is absent, and tests proving persistence across checks and across seat rotation.

A lane is adding that to this branch so the whole capability merges and deploys as one piece, rather than shipping the toggle now and the prompt later.

Note for whoever re-runs the gates: the reviewer could not run typecheck, the privacy gate or the MCP tests because its worktree had no `node_modules`. Those gates are unverified on this branch so far and must pass before merge.

Latand:
Continuing on this branch, 2026-08-29 — the lane ran out of review rounds, not out of agreement. The findings have been converging each round rather than repeating.

Confirmed and not in question: the schema exposes the prompt, the binding persists it through the settings record, later calls reread it, stored prompts are bounded and redacted, and a project with no settings row behaves exactly as before.

Two findings are being closed.

**Replacing or clearing a prompt can collide with a wake already in flight.** The wake's idempotency key is built from the seat epoch, the reasons, the board fingerprint and the prior wake stamp — the prompt joins the text afterwards. So a wake carrying one prompt is held, the stored prompt is changed, and the next unchanged check produces the same key with different text; delivery rejects the changed payload and keeps rejecting it until the old wake settles. The fix folds a bounded prompt digest into the identity of prompt-bearing wakes only, leaving promptless keys and the scheduling untouched.

**A maximum-length prompt can truncate the wake's own contract.** The prompt goes in ahead of the contract text and the whole message is then clamped at 4,000 characters. Measured: a production-shaped five-item wake is 2,902 characters with every clause present; adding an accepted 1,000-character prompt lands on exactly 4,000 and drops the final clause. That contract is what tells a seat not to schedule itself and what its stop is, so quietly losing a clause is worse than refusing the prompt. The fix reserves the prompt and the contract before bounding the dynamic content, with a maximum-prompt, full-agenda regression asserting every clause survives.

Latand:
Cannot merge yet: main moved and this branch now conflicts, 2026-08-29.

The work here was reviewed clean with zero findings at `8e371ecc` and that verdict stands — nothing about the feature is in question.

Exactly one file conflicts: `src/lib/mcp/bindings.ts`. Both sides touched it — this branch added the seat tick settings binding with its prompt field, and main gained the MCP reconnect work. Everything else merges cleanly.

A lane is resolving it so both halves survive, then re-running the gates on the merged result, since the approval predates the merge. The risk being watched for is a resolution that quietly drops one side.

Issue #1282, opened 2026-08-29
fix(startup): a structured host launched while the runtime-host successor waits for the singleton fence is never claimed, and every recovery verb is then refused


Latand:
A redeploy today left one pipeline lane permanently stuck. The evidence points at a narrow startup window; the mechanism below is stated as suspected where it is inferred rather than observed.

## Timeline, 2026-08-29

| time (UTC) | event |
|---|---|
| `03:24:05` | lane's builder writes its last transcript event — a tool result, no assistant answer yet |
| `03:25:50` | new viewer container starts (redeploy) |
| `03:26:37` | runtime-host successor starts and logs: *"waiting for the singleton fence; waited 0s of 600s; the predecessor has not released it yet"* |
| `03:28:15` | the viewer process relaunches the severed session — **inside that fence wait** |
| `03:34:39`, `03:37:42` | two other lanes' hosts are launched — both healthy afterwards |
| `04:02` | the `03:28:15` host has burned 4.7 s CPU total, written nothing since its own launch, and is parked in `epoll_wait` on stdin |

The two hosts launched six and nine minutes later, after the wait window, work normally. The one launched during it never did anything.

## Suspected mechanism

A structured host launched while the runtime-host successor is still waiting for the singleton fence is never claimed by the host that ends up owning traffic. The process exists and holds its stdin socket, but no owner ever writes a turn into it. Everything downstream then follows from that:

- the conversation keeps `turn.state: busy`, inherited from the predecessor that was severed mid-turn;
- `interrupt` and `kill` are refused with *"structured runtime host is unavailable"* — the current runtime host does not own it;
- a `send` is accepted and sits `queued` behind the open turn, which nothing is running;
- the pipeline stage stays `running`, so `retry-stage` is refused with *"pipeline does not have a stage awaiting retry"*.

Every recovery verb is blocked by a different guard, and the only thing that moved the lane was killing the pid by hand from a shell.

## Also observed while recovering

After the process was terminated, the Viewer went on reporting `proc: running, pid <the dead pid>` for that conversation. A dead pid reading as a running process is its own defect and belongs to #1281's evidence-based liveness decision.

## Wanted

1. **No unclaimable launch.** A structured host is not launched into a window where nothing can claim it — either the launch waits for the fence like everything else, or the launch is retried by whoever takes the fence. Whichever is chosen, an unclaimed host must be detectable rather than silent.
[...]

Issue #1280, opened 2026-08-29
feat(monitor): the seat's monitor is an explicit agent-managed record — on/off, its own prompt, fired by the scheduler


Latand:
Operator direction, 2026-08-29, after reading how the tick behaves today. It revises the boundary #1275 drew.

## The problem, in the operator's terms

A seat has no way to say **"I am finished here."** When the work is done he simply stops replying, to avoid spending quota — and nothing anywhere records that the work is over. So the Viewer keeps deciding on its own whether the seat is owed a wake, and the seat keeps producing empty turns that say "no change". Marking every finished seat by hand is not an answer.

## The direction

Two things are currently in tension and both are half right:

- The tick contract forbids a seat from scheduling itself. That was introduced for a real failure: a schedule that lives inside a session dies with the session, so every rotation silently dropped the monitor, and while a session-held schedule keeps a turn open the Viewer's own check finds the seat busy and drops it — two clocks deadlocking on one seat (#1245).
- The restart nudge tells every seat to "re-arm any scheduled work" (#1276), which only makes sense if a seat is supposed to have scheduled work.

The resolution is not "a seat has no schedule". It is that **the schedule is a durable record the Viewer owns and the agent controls**:

- **A state: on or off.** The agent turns its own monitor off once it concludes there is nothing left to watch, and on again when there is. The operator can do the same from the board.
- **A prompt the agent writes.** What this monitor should look at when it fires, in the agent's own words, rather than only the fixed reason list the tick derives.
- **Fired by the scheduler, not by turn completion.** Codex's Goal re-arms itself after every response; that is the shape to avoid. This one fires from the Viewer's clock and from nothing else.
- **Durable in the Viewer** — surviving a restart, a deploy and a rotation. That is exactly what "re-arm any scheduled work" is reaching for and what a session can never provide.

## Relation to #1275

#1275 asks for a stop: a suppression with a reason and a bound, and it explicitly rules out a cadence setting or a configuration surface. The operator's direction supersedes that boundary — he wants the fuller record, agent-managed. The lane implementing #1275 may land the stop first; this issue is the shape it should grow into, and the stop should arrive as the `off` state of this record rather than as a separate mechanism to reconcile later.

## Relation to #1276

Once the record exists, the restart nudge becomes honest: a seat whose monitor is `off` is not resumed and not messaged on a redeploy, and the message that survives names the turn that was actually severed instead of asking for a re-arm that no longer means anything.

## Acceptance

[...]

Issue #1276, opened 2026-08-29
Every Viewer restart nudges every orchestrator seat, including idle ones with nothing owed — a resumed host and a wasted turn per seat, per restart


Latand:
Reported by the operator, 2026-08-29: dormant projects keep coming back to life on the board. A project nobody has touched in days shows fresh activity, its orchestrator seat spends a turn, and the turn says nothing happened.

## What happens

`src/lib/runtime/startup.ts` — on every Viewer boot, `enqueueOrchestratorRestartRecoveries` delivers this to **every** active orchestrator seat:

> Viewer restarted and severed your structured host. You were re-hosted automatically. Resume your checkpoint loop and re-arm any scheduled work.

The target predicate `orchestratorRestartRecoveryTargets` selects a seat whose registry entry was `structuredHost && host === null && (status === "live" || status === "idle")`. **`idle` is in that set**, so a seat with no turn in flight — nothing severed, nothing to recover — is resumed and spends a model turn saying so.

## Evidence, this machine, 2026-08-29

- 11 active orchestrator seats. One restart is therefore up to 11 resumed structured hosts and 11 model turns.
- One dormant project's seat: no open lane, its only board card `blocked`, two PRs parked on an operator decision. Its transcript across the last two Viewer restarts:
  - `2026-08-28T20:11:40Z` — recovery text delivered; the seat answers "checkpoint unchanged".
  - `2026-08-29T03:32:38Z` — recovery text delivered; the seat answers "checkpoint unchanged, third restart in a row".
- Across that same window the seat tick's own verdict for that project was `quiet — nothing owed` at **every** five-minute check: 45 consecutive rows in `state/seat-tick/runs.ndjson`, and its `state/seat-tick.json` row has not recorded a wake since `2026-08-28T16:32Z`. The tick was right and stayed silent; the restart path woke the seat anyway.
- The board half of the complaint follows directly: the nudge stamps a fresh turn on a long-dormant project, so freshness ordering floats dead projects back up.

## Why it is wrong now

1. **There is nothing left to re-arm.** The prompt dates from when a seat's monitor lived inside its own session, so a restart severed the host and killed the in-session schedule with it. Since #1245/#1261 the Viewer owns the clock: the tick's state is durable, survives restarts, deploys and rotations, and wakes the seat when something is actually owed.
[...]

PR #1286, opened 2026-08-29
fix(runtime): decide a turn's liveness from evidence, stop stranding a host on redeploy, and nudge only severed seats (#1281, #1282, #1276)

### LINKED ISSUE #1281: fix(liveness): decide "this turn is working" from evidence, not from the status word — and never queue a send behind a turn nothing is running
Operator direction, 2026-08-29: the Viewer does not actually know when an agent is working, and both directions of the mistake are live.

- **A long tool call reads as stalled.** A step that legitimately takes ten minutes — generating a large file — looks identical to a dead turn, and anything acting on "stalled" then acts wrongly.
- **A severed turn keeps reading as busy**, forever, because the status word outlived the process that earned it.

## A live specimen, this machine, 2026-08-29

A pipeline builder host was severed by a redeploy at `03:25Z`, immediately after a tool result and before the assistant message that would have answered it. The Viewer relaunched the session at `03:28:45Z`. Thirty-four minutes later that process had:

- consumed **4.7 s of CPU in total**,
- written **nothing at all** to the transcript since before it was launched — its last transcript event is the pre-redeploy tool result,
- sat in `epoll_wait` on its stdin socket for the entire time.

Meanwhile the registry read `turn.state: busy` (inherited from the severed predecessor), the conversation read `activity: stalled, proc: running`, and the pipeline stage read `running`. The consequences compose into a deadlock:

- `retry-stage` was refused — *"pipeline does not have a stage awaiting retry"* — because the stage still reads running;
- `kill` answered *"structured runtime host is unavailable"*;
- a `send` was accepted and sat `queued`, because the queue waits for the open turn to end, and no process is running that turn.

**So a resumed host that inherits an open turn from a severed predecessor can wait forever for a prompt that is queued behind the very turn it inherited.** Nothing errors; the lane simply stops.

## What the decision should read

Name the evidence instead of trusting a status word:

- the last transcript event **and its kind** — an unanswered tool call is a different fact from a settled assistant message;
- CPU consumed since the process was launched — a host that has burned essentially none and written nothing **since its own launch** has not started, whatever the registry says;
- whether the process exists at all, and whether it is the process the registry believes owns the turn;
- whether a delivery is outstanding for it, and how long it has been outstanding.

A host writing to its transcript, or burning CPU, is working however long the gap between messages — that is the answer to the ten-minute tool call. A host doing neither since its own launch, under an open turn it inherited rather than opened, is severed — that is the answer to the specimen above.

## Why it matters beyond one lane

- **#1276** turns on exactly this distinction: the restart nudge fires for a seat whose entry says `idle` as well as `live`, and tightening or widening that status set only papers over the fact that neither word is evidence about the seat.
- The seat tick's own stall rule, `#1274`/`#1275`, and every "is this lane moving" judgement rest on the same call.
- The recovery verbs must not be mutually blocking: a turn that cannot be shown to be running must be settleable, so `kill` and `retry-stage` stay available on precisely the hosts that need them.

## Acceptance

- A turn whose host is writing to its transcript, or consuming CPU, is never reported stalled — including across a ten-minute gap with no message.
- A resumed host that has written nothing and burned no CPU since its own launch is reported severed, not busy, and its inherited open turn is settled so that `kill` and `retry-stage` both work on it.
- A send is never left queued behind a turn no live process is running.
- Regression covering: (a) long-running tool call → alive; (b) resumed host that never received its prompt → severed and recoverable; (c) host writing normally → unchanged.

Latand:
Closes #1281, #1282 and #1276 — one failure seen from three sides.

## The suspected cause in #1282 is refuted

#1282 asked whether a structured host launched while the runtime-host successor
waits for the singleton fence can end up unclaimed. From the code, it cannot:

- `src/runtime-host/main.ts` acquires the fence (line 43) **before** it opens
  the journal and long before `serveRuntimeHost` listens (line 102). A
  generation that is *waiting* has taken nothing from the one that is serving —
  the predecessor still holds the fence, still owns the socket, and still
  answers. The wait window is a window in which the runtime host is available.
- A claim is durable, not in-memory: the Viewer publishes `session-status` into
  the shared SQLite journal, and a successor's `claimHostEpoch()` reconciles
  only never-materialized `registering` placeholders. A `hosted` session
  survives the epoch bump untouched.
- Engine hosts live in Viewer processes and are never signalled by the
  succession (`main.ts` says so where the hand-over is implemented).

What the code *does* show, and what this PR fixes, is the state the incident
actually described. The Viewer re-hosts a conversation whose turn was severed
mid-flight (`entry.status === "live"` plus an open turn), the fresh host ends
its inherited turn on its own side and sits idle, and nothing ever writes a
turn into it: the Codex continuation is engine-filtered and the orchestrator
nudge is seat-only, so a Claude worker gets neither. Meanwhile the durable
conversation keeps reading `busy` — its transcript ends on an unanswered tool
result and nothing will append to it again — and each recovery verb is refused
by a different guard.

The exact string the operator saw for `kill`/`interrupt` ("structured runtime
host is unavailable") could not be reproduced from this tree; on `main` that
message has one source, `structuredControls.ts` with no runtime client. The
durable equivalent is reproducible and is what this PR removes: a kill on a
host no controller owns hits `terminateInactiveStructuredHost`, which refuses
while the recorded process is alive, so `drainControl` reports the group
blocked and the kill effect — plus every message queued behind it — never
settles.

## The decision (#1281)

`src/lib/runtime/liveness.ts` decides "is this turn being worked on" from
evidence, and its verdict carries the evidence in words:

- the last transcript event **and its kind** (an unanswered tool call is not
  the same fact as a settled assistant message), plus the artifact's own clock
  so a host appending to a transcript this reader cannot parse still counts as
  writing;
- whether the process the registry believes owns the turn exists, and is still
  that process;
- CPU consumed **since that process launched**, derived from the same
  start-time token the registry already stores to fence a pid against reuse
  (`processCpuMs` is new on the proc backend; the portable backend answers
  `null`, which is read as "no evidence" and never as zero);
- whether a delivery has been outstanding for it, and for how long.

`working` is granted on any positive sign — including a ten-minute gap after a
tool call this host wrote. `severed` needs positive evidence of absence and is
refused whenever the platform cannot supply it. Everything else is `unknown`,
which every consumer treats as "do not act".

## The recovery verbs (#1282)

- **retry-stage.** The pipeline's `conversationHostUnavailableSince` port now
  consults the decision instead of stopping at `entry.status`. A stage whose
  host is proven severed leaves `running` after the existing grace, so
  `retry-stage` works without closing the pipeline.
- **kill.** When no registration in this process owns the host and the evidence
  says severed, the recorded process is reaped — fenced on its start identity,
  so a reused pid is reported severed and never signalled — and the row
  retires. A host that is working, or that cannot be shown not to be, is left
  strictly alone.
- **interrupt.** A control for a conversation with no host used to block its
  group forever. Proven severed, it settles as `interrupted` and the drain
  moves past it, so the message queued behind it stops waiting on a turn no
  process is running.
- **detection.** An adopted host the controller never claimed fails the startup
  pass (the hosts stay in the retry set, which re-registers them) instead of
  being reported as adopted; and a startup completion that resumes inside a
  generation the publication has already left hands its hosts to the successor
  instead of answering "done" and registering nothing.
- A carried-over `proc: running` is revalidated against the pid, so a dead
  process stops reading as a running one.

## The seats (#1276)

The nudge target set is no longer a status word. A seat is messaged only when
the evidence says a turn of its own was severed and the transcript leaves that
turn open; an idle seat receives no message and is not adopted, so no process
is started for it. The surviving message names the turn it is about ("the
severed turn is the one whose last transcript event is a tool-result at …") and
drops "re-arm any scheduled work" — since #1245 the Viewer owns the clock and a
seat has no lever for that sentence; it returns when #1280 gives it one. The
seat-epoch fencing for a revoked or rotated seat is unchanged and still proven.

## Scope left alone

- `src/lib/monitor/**` (#1274/#1275), `src/lib/runtime/structuredRecovery.ts`
  and `src/lib/flows/engine.ts` (#611) are untouched; nothing here needed them.
- The seat tick's wake reasons and interval are unchanged.
- Retrying a stage does not reap a severed host process that is still resident;
  the operator's kill now does, and automatic retirement still owns the rest.
  Worth a follow-up if it shows up again.

## The second review's four findings, closed

- **Unreadable transcript evidence no longer borrows the row's turn word.**
  `readTurnLiveness` used to substitute `conversation.turn.state` when the tail
  could not be parsed, which put the status word this decision exists to stop
  trusting back at the centre of it: an unreadable transcript plus a stale
  `busy` produced `severed` with no last event and no kind behind it, and a
  stale `terminal` produced `settled` the same way. Missing, corrupt or
  unstable transcript evidence now leaves the turn `unknown`, and `unknown`
  authorises no kill, no retry and no restart nudge. Negative cases cover all
  three seams: the decision itself (both words), the kill (a live host with an
  unreadable transcript is never reaped), the nudge (a seat with an unreadable
  transcript is owed none), and the retry (a stale `terminal` no longer settles
  a stage whose host is gone).
- **An identity that cannot be revalidated is not evidence.** A present pid
  whose recorded start identity the kernel would not report — or that the row
  never recorded — bypassed the mismatch branch and could still become
  `severed` from an outstanding delivery, on a process nothing proved was the
  recorded one. It answers `unknown`.
- **Nothing is launched that a pass cannot hand to a delivery controller.**
  `controllerBoundEarly` gated the unclaimed-host assertion but not the
  adoption loops, so a pass with no runtime client started CLI processes and
  reported success with the only check that would have caught it behind the
  same condition. That pass now defers: nothing is launched, the boot's own
  recovery evidence and retained hosts are kept, and the startup retry loop
  runs it again once a client exists. The regression asserts zero launches
  while unclaimable and exactly one claimed host, with its single nudge, after
  recovery.
- **The kill and retry regressions no longer stub what they prove.**
  `src/lib/pipelines/severedStageRetry.test.ts` drives the whole incident
  through production seams: a live child that has written nothing since its own
  launch, the operator's kill admitted as a runtime operation and executed by
  the delivery controller, the registry retirement, a pipeline tick reading its
  own liveness port, and `retry-stage` producing exactly one replacement
  attempt without closing the pipeline.

One case is honest about its own limit: the stale-`busy` mirror at the stage
port pins the shape rather than the defect, because a `busy` word only ever
reached a verdict past the 90s observation window and a pid spawned inside a
test cannot be on the far side of it. The stale-`busy` cases that go red
without the fix are the decision, the kill and the nudge; the comment says so.

## Verification

- Touched suites by path, under an isolated `HOME`/`XDG_CONFIG_HOME`/
  `LLV_STATE_DIR`/`TMPDIR`: `liveness` 21, `severedHostReap` 4,
  `stageHostLiveness` 5, `severedStageRetry` 1, `startup` 74,
  `structuredDeliveryQueue` 39, `structuredDeliveryRebind` 11,
  `resourceScopeProcess` 2, `pipelines/engine` 207, plus `instrumentation`,
  `structuredControls`, `structuredRecovery`, `structuredHostRetirement`,
  `registryPersistence`, `startupCredentialIsolation`, `stageHostTeardown`,
  `pipelines/controller`, `scanner/discover`, `proc/portable` — all green.
- Red path re-checked for every new regression, in a throwaway worktree at this
  branch's head with one file at a time reverted to its pre-fix state — no
  stash. Against the pre-fix `liveness.ts`: the two stale-word decision cases,
  the identity case, the unreaped kill and the un-nudged seat all fail. Against
  the pre-fix `startup.ts`: the unclaimable-adoption case fails. Against the
  merge base's `engine.ts`, the end-to-end case fails where the incident did —
  the stage is still `running` after the tick; against the merge base's
  delivery controller it fails one step earlier, on a kill that never settles.
- Pre-existing failures on the merge base, confirmed by running the same files
  at the merge base in a throwaway worktree and unrelated to this change:
  `structuredDelivery.integration` (2), `structuredSpawn.integration` (3).
- `bun scripts/privacy-publication-gate.ts --base <merge-base> --check-commits`
  — PASS.

Nothing was deployed, promoted, restarted or rolled back.


Latand:
Orchestrator decision, 2026-08-29 — five build attempts and five review rounds returned to the same question, so it is settled here rather than sent round again.

**The reviewer is right, and its prescription is what gets implemented.** Incomplete, corrupt or unreadable transcript evidence short-circuits to `unknown` with no `since`, before any branch can return `severed` — including the branch that fires because the recorded process is gone, and the one that fires from an old delivery plus low CPU.

The counter-argument was weighed and rejected, and the reasoning is worth keeping: a missing process proves that **that pid** is not running, but it cannot distinguish a turn severed mid-work from a turn that finished long ago under a row nobody updated. With the transcript unreadable those two are the same observation, and the consequences are not the same. A kill is harmless either way. A **retry re-runs work that may already be complete**, and a **continuation nudge re-prompts a seat about a turn that is over** — which is precisely the class of failure this change exists to remove. So `unknown` has to reach the consumers instead of a verdict they are entitled to act on.

Three existing tests encode the old behaviour on purpose — one expects the stale-terminal case to become retryable, another sits inside the observation window where it cannot fail, and two startup tests preserve adoption for corrupt transcripts. Those are changed, not worked around, and negative regressions are added asserting zero kill, zero retry, zero launch and zero continuation under unreadable evidence.

Also flagged and being removed: this branch had picked up `src/lib/scanner/**` changes that belong to another lane.

Everything the earlier rounds confirmed stays: unverifiable identity already returns unknown, startup launches nothing without a claimable controller and exactly one host after recovery, the ten-minute tool call stays alive, an idle seat gets nothing on a redeploy.

Latand:
Cannot merge yet, and the reason is an integration failure rather than anything wrong with this branch on its own, 2026-08-29.

The `macOS process identity` job — the only oracle for that platform, since nobody here has a Mac:

| commit | what it is | result |
|---|---|---|
| `7d1bfa90` | this branch's approved head | pass |
| `eab408a6` | current main | pass |
| `714bd4be` | the merge of the two | **fail** |

One test fails, *"on macOS the reader returns the exec-time argv of a live child"*; the rest of that job passes (49 pass, 2 skip, 1 fail) and every other check on the merged head is green.

So neither side is broken alone — something this branch changed in the shared process-reading contract (`src/lib/proc/linux.ts`, `portable.ts`, `types.ts`) and something that landed on main disagree only once combined.

A lane is establishing which two changes disagree before touching anything, then fixing it so the merged head passes without weakening what that test protects: the reader must return the argv the child was executed with, which is what stops process identity being confused by a later mutation. Skipping or relaxing the assertion would not count, and neither would a macOS-only special case that drifts from the contract the Linux and portable readers keep.

Also recorded for whoever picks it up: there is no Mac here, so that job is the only evidence, and claims about it have to come from reading the run rather than from a local pass.

PR #1287, opened 2026-08-29
fix(scanner): keep relocated worktree transcripts visible

Latand:
## Summary

- Rebinds a Claude conversation current generation to an existing same-session continuity path when the registered file is gone and the surviving file exists.
- Keeps a continuity artifact secondary while the current transcript still exists, preserving the existing migration and fork boundary.
- Covers live and removed worktree grouping, archive folding, readable parent projection, and child lineage without `parentRemoved`.

## Verification

- `bun test src/lib/agent/registry.test.ts` (123 pass)
- `bun test src/app/api/files/route.test.ts` (92 pass)
- `bun test src/lib/scanner/describe.test.ts` (34 pass)
- relocation regression under `LLV_AGENT_REGISTRY_SQLITE=dual-write`
- `bunx tsc --noEmit --incremental false`
- `bun scripts/privacy-publication-gate.ts --base 7d66f146ae828739560cc578ea29362e351956dd --check-commits`

Closes #566

PR #1288, opened 2026-08-29
fix(mcp): reconnect control calls after Viewer restart

### LINKED ISSUE #1283: fix(mcp): a redeploy severs every running agent's MCP channel — the runtime is pinned to the release's candidate port, not the stable proxy
Reported by two orchestrator seats and confirmed on this machine, 2026-08-29.

## Symptom

After a redeploy, an agent that was already running loses every `mcp__viewer__*` tool. One seat reported it as *"the Viewer is down"*, then corrected itself to *"the Viewer keeps moving to a new port"* — it had watched the port go `18024 → 19559 → 19616` across deploys. Its board reads over HTTP still worked; its MCP channel did not, so it could no longer spawn or message anything.

Agents fall back to raw `curl` against the API, which is slower, needs the token, and loses every guarantee the MCP surface provides.

## Mechanism

- Each deploy runs the new viewer container on a **fresh candidate port**, and `state/viewer-release.json` records that port as `endpoint`. The stable address is the runtime-host proxy on `8898`, which forwards to whatever the current candidate is.
- The MCP launcher (`~/.agents/tools/llv-mcp-runtime/bin/mcp-server.mjs`) reads that release target **once, at process start**, to pick the release package and its endpoint.
- So an MCP child is pinned to the release that existed when its agent launched. The next deploy moves the endpoint, and the pinned child keeps addressing a port nothing listens on. Nothing tells it; the tools simply stop answering.

Verified: `127.0.0.1:8898` is held by the runtime-host container and proxies the API correctly, while `19616` is the current deploy container's own port — the very value written into `endpoint`.

## Wanted

1. **An agent's MCP channel survives a redeploy.** Either the MCP runtime addresses the stable proxy rather than the candidate port, or it re-reads the release target and re-targets when a call fails — whichever keeps the guarantee that a running agent does not silently lose its control surface.
2. **A severed channel is loud.** If a channel genuinely cannot be re-established, the agent gets an error that says so, instead of tools that stop returning.
3. **Regression** driving a deploy that changes the candidate port and proving an already-running agent's MCP calls still land.

## Related

- #1282 — the same redeploy strands structured hosts; this is the same event severing the control channel.
- #1276 — the same redeploy resumes idle seats.

Latand:
Closes #1283

## Mechanism

The managed release target selects the exact MCP package and hot-state revision. MCP control HTTP already defaults to the stable runtime-host address on port 8898, so the candidate-port theory did not explain the same-port reboot.

The severing boundary was the long-lived MCP process's control fetch. A retiring Viewer generation could accept a request and stop answering it. Control mutations had no deadline, and no control call retried after its peer disappeared. The replacement generation could bind the identical port while the original MCP call remained attached to the retired connection. Once the MCP client abandoned that stuck stdio server, the whole tool surface disappeared.

The regression drives the packaged stdio MCP process through that sequence. At the base commit, a fresh probe reached the replacement generation while the already-running MCP call never did and timed out after seven seconds.

## Fix

- Bound each network attempt through response-body consumption.
- Retry transport failures, proxy 502/504 responses, and the runtime proxy's empty 503 within the MCP call budget.
- Preserve JSON 503 domain failures as their original product errors.
- Reuse every mutation endpoint's existing idempotency identity during retry.
- Return a named reconnect-exhaustion error while keeping the stdio tool surface alive.

## Verification

All stateful tests ran with isolated HOME, XDG_CONFIG_HOME, LLV_STATE_DIR, and TMPDIR.

- `bun run build:mcp`
- `bun test src/lib/mcp/stdio.integration.test.ts`
- `bun test src/lib/mcp/deploymentReads.test.ts`
- `bun test src/lib/mcp/bindings.test.ts`
- `bun test bin/mcp-server.test.ts`
- `bunx tsc --noEmit --incremental false`
- `bun scripts/privacy-publication-gate.ts --base 7d66f146ae828739560cc578ea29362e351956dd --check-commits`
- `git merge-tree --write-tree origin/main HEAD`


Latand:
Approved and queued for merge, 2026-08-29.

Review round 2 returned pass with zero findings at `9fde654e`, which is this PR's head and the lane's published commit. The lane completed on that verdict.

It is **not** being merged this moment on purpose. #1287 was already integrated against the current main and is waiting on CI; merging this one first would push that branch behind again and cost it a second update-and-CI cycle. So the order is #1287 first, then this one updated against the resulting main and merged when its checks are green.

Nothing is deployed either way — the operator is batching this release.

Latand:
Brought up to date, 2026-08-29, now that the two ahead of it have landed.

Approved at `9fde654e` with zero findings. The new head `a0c221e4` is a merge commit whose two parents are exactly that reviewed head and the current main; every file that differs between the reviewed head and the merge result arrived from main (the provider-throttle work and the relocated-transcript work). None of this lane's own code changed and no conflict resolution touched it.

So the approval stands for this lane's diff, and only CI on the integrated head remains. It was queued at the time of writing; merging once it is green needs no further round.

PR #1292, opened 2026-08-30
fix(delivery): an accepted send settles, and its receipt says what became of it (#1131)

### LINKED ISSUE #1131: P1: send_message 'queued' can be silently lost — receipts must settle to delivered/failed and be queryable
## P1 — send_message answered "queued" and the message never arrived; no way to see the loss

Reported by a peer orchestrator seat (another project on this machine), 2026-08-24 ~07:00Z, prod `9355e4eb`.

- MCP `send_message` to a Codex conversation returned outcome `queued` (operationId `6b8ecd7b-be21-465b-be2d-32f83442362e`). The message **never appeared** in the recipient rollout (recipient created 2026-08-24T09:13:49 local). The recipient — a production deployer holding a paused blue/green cutover — sat idle for ~40 minutes until the seat's own silence-monitor noticed.
- An identical retry (operationId `e1ed85cb-…`) delivered instantly, so the payload and the target were fine; the queued delivery itself was lost.

`queued` as a terminal answer hides message loss on a critical control path. The delivery journal exists (#1065 made flow relays follow it) — MCP callers need the same truth:

1. `send_message` receipts must settle: a queued delivery either becomes `delivered` (confirmed by the journal/transcript echo) or `failed` loudly; a caller must be able to query the operation by `operationId` (or the receipt must be pushed to the conversation card) instead of trusting `queued` forever.
2. A queued delivery that cannot be handed to a live host within a bounded window is retried by the delivery controller itself or terminalized with a visible error on the sender's side — never silently dropped.
3. Regression: fixture where the host claim/queue drops the payload after admission → the operation reports failed within the bound; the retry path stays idempotent by clientRequestId (no double delivery — coordinate with #1113).

Refs: #1065 (journal-backed relay settlement), #1114 (delivery controller unavailability), #1113 (idempotent redelivery), the held-delivery diagnosis of 2026-08-15.

Latand:
Closes #1131

## The failure

`send_message` answers `queued` the moment the runtime host admits the operation, and until now that could be the last thing anyone ever said about it. A delivery the drain never executed rested there forever: the reservation stayed `delivery-uncertain`, the receipt stayed non-terminal, and nothing anywhere turned that into an answer. A deployer holding a paused cutover sat idle for forty minutes on exactly this, and the only thing that noticed was the sender's own suspicion.

## What changed

**Every accepted send settles.** `src/lib/runtime/sendSettlement.ts` sweeps the reservations that have not reached a terminal state, asks the delivery journal what became of each, and ends it as `delivered` or `failed`. A send the journal already delivered is reconciled rather than reported lost; an unexecuted one is declared lost. The sweep runs in the release that owns traffic, beside host retirement, because it fences operations that this process's own delivery queue would otherwise still execute.

A send queued behind the recipient's active turn is progressing, not lost, so it is exempt from the settlement window — with a ceiling, so a host wedged mid-turn cannot make `queued` permanent again. A send whose journal answer cannot be read is deferred, never settled: an unreachable runtime host is a reason to know less, not a licence to declare a message lost while it may still be sitting in a live outbox.

**A receipt is queryable.** A new `message_receipt` tool answers what became of one operation id — `delivered`, `failed` or `in-flight` — read from the durable delivery record rather than from what the send call guessed at the time. It also says what is safe to do next: `not-needed`, `safe`, or `verify-first` when `duplicateRisk` is set — and that a resend is a NEW `send_message` under a NEW `clientRequestId`, because the settled operation still owns the old key and repeating it replays the settled answer instead of delivering. `send_message` and `send_message_to_orchestrator` now report `settled` on their own answers, so acceptance is not mistaken for arrival, and both tool descriptions say so.

**A loss is visible without a watchdog.** The settlement journals the `delivery_expired` lifecycle event as it declares the loss, rather than leaving it for whichever poll happens next — so it is in the lifecycle digest, which orchestrators already read, and on the recipient's card as a failed send. No new surface, nothing for the sender to watch.

## Why this cannot deliver the same instruction twice

Declaring a send failed is only honest if the send can no longer happen, so the settlement terminalizes the **operation in the runtime journal first**, and writes `failed` on the reservation only once that succeeded. Two properties of the journal make that a real fence:

- terminalizing an operation marks its outbox row completed inside the same `BEGIN IMMEDIATE` transaction, so the effect is gone from every later effect batch; and
- the delivery queue must move an effect to `delivering` **before** it calls `host.send`, and the journal refuses that transition out of a terminal state — so an executor that had already batched the effect is stopped at the transition rather than at the send.

A send the executor already took (`delivering`) may have reached the recipient, and nothing here can prove it did not, so it is terminalized as `uncertain` and the receipt says the fate is unknown and a resend may duplicate. **Nothing is retried automatically** — a resend is the caller's decision, made against a receipt that states whether it is safe. Nothing widens a timeout: the settlement deadline is a deadline for producing an answer, not a wait anything sits in.

A retried send is judged by its current attempt, not by the ancestor it replaced: reading the ancestor would report a delivered message as a loss and would then try to fence an operation that is already terminal.

## Verification

All fixture-built; nothing was sent into a real conversation, and every suite ran by path under an isolated `HOME`/`XDG_CONFIG_HOME`/`LLV_STATE_DIR`/`TMPDIR`.

- `src/lib/runtime/sendSettlement.test.ts` — 10 pass. Real `AgentRegistry` and real `RuntimeJournal` per fixture: a delivered send reconciles; a dropped send settles failed and the real `StructuredDeliveryQueue` then delivers nothing (with a control test proving the same port, queue and host **do** deliver an unfenced send, so the empty recipient is the fence's doing and not the harness's); an uncertain first attempt settles failed and refuses to call a resend safe; a retried send is judged by its leaf; an in-turn send waits and stops being exempt at the ceiling; an unreadable journal defers; the loss is journaled as it is declared. The caller's own replay of a settled loss is answered with the settlement rather than a second delivery — admission replays the terminal receipt, no effect is admitted, and the queue delivers nothing; a scratch control run confirmed each of those three assertions turns on the settlement (without it the replay is `queued`, the effect batch holds the send, and the queue delivers the text).
- `src/lib/mcp/bindings.test.ts` — 47 pass, including `settled: false` on an accepted send, `message_receipt` answering in-flight then failed, and a refusal for an id nothing admitted.
- `src/lib/mcp/server.test.ts` 33 pass · `schemaParity.test.ts` 18 pass · `presentation.test.ts` 3 pass · `src/instrumentation.test.ts` 40 pass.

Pre-existing and not from this branch: `src/lib/delivery.test.ts` fails 2 ("title is required for every new spawn") — 34 pass / 2 fail on this branch and the identical 34/2 at the merge base in a throwaway worktree. `bunx eslint` cannot run in this checkout at all — `eslint-plugin-react` throws loading `react/display-name` under the installed ESLint — which is also unrelated to this change.

Nothing was deployed, promoted, restarted, or rolled back.

### DIFF
--- a/README.md
+++ b/README.md
@@ -220,7 +220,8 @@ do not render as Viewer cards.
 The MCP surface includes:
 
 - conversations and the board: `board_snapshot`, `list_conversations`,
-  `get_conversation`, `send_message`, and `conversation_action` for
+  `get_conversation`, `send_message`, `message_receipt` (what became of an
+  accepted send, by its operation id), and `conversation_action` for
   `interrupt`, `kill`, `resume`, `compact`, and `dialog-key`;
 - review flows: `list_flows`, `get_flow`, and `flow_action`;
 - pipelines: `create_pipeline`, `list_pipelines`, `get_pipeline`,
--- a/src/instrumentation.test.ts
+++ b/src/instrumentation.test.ts
@@ -515,6 +515,36 @@ test("automatic host retirement starts with the release that owns traffic, not w
   expect(started).toEqual(["pipelines", "host-retirement"]);
 });
 
+test("send settlement starts with the release that owns traffic, and a failure to start it costs only itself", async () => {
+  /* #1131: the sweep fences operations in the runtime journal that THIS
+     process's delivery queue would otherwise still execute, so it belongs to
+     the process that owns traffic — and, like every controller here, it must
+     not take the rest of the boot down when it cannot start. */
+  const started: string[] = [];
+  await startCurrentReleaseControllers(
+    { LLV_ACCOUNT_CONTROLLER_DISABLED: "1" },
+    {
+      loadFlowPipelineController: async () => ({ startFlowPipelineController: () => { started.push("pipelines"); } }),
+      loadAccountMigrationController: async () => ({ startAccountMigrationController: async () => { started.push("account"); } }),
+      loadSendSettlement: async () => ({ startSendSettlement: () => { started.push("send-settlement"); } }),
+      loadSeatTick: async () => ({ startSeatTick: () => { started.push("seat-tick"); return true; } }),
+    },
+  );
+  expect(started).toEqual(["pipelines", "send-settlement", "seat-tick"]);
+
+  const withoutSettlement: string[] = [];
+  await startCurrentReleaseControllers(
+    { LLV_ACCOUNT_CONTROLLER_DISABLED: "1" },
+    {
+      loadFlowPipelineController: async () => ({ startFlowPipelineController: () => { withoutSettlement.push("pipelines"); } }),
+      loadAccountMigrationController: async () => ({ startAccountMigrationController: async () => { withoutSettlement.push("account"); } }),
+      loadSendSettlement: async () => { throw new Error("runtime state unavailable"); },
+      loadSeatTick: async () => ({ startSeatTick: () => { withoutSettlement.push("seat-tick"); return true; } }),
+    },
+  );
+  expect(withoutSettlement).toEqual(["pipelines", "seat-tick"]);
+});
+
 test("the seat tick starts with the release that owns traffic — one clock, one process", async () => {
   /* #1245: the thing that drove orchestrator sessions was a prompt an agent
      scheduled for itself, so it died with the session and every rotation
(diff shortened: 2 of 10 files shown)


Latand:
Superseded by #1297, which carries this work forward on a continued branch after a redeploy severed the original lane's host. Closing so the board shows one live line of work for the send-settlement fix rather than three.

PR #1293, opened 2026-08-30
feat(accounts): bind accounts to projects, and let a pipeline stage honour the binding

### LINKED ISSUE #1279: feat(accounts): bind accounts to projects (many-to-many) so a switch can only ever pick an account that project is allowed to use
Operator direction, 2026-08-29.

## What is missing

Every account in the registry is usable by every project, and the active account is one global selection. The operator's rule is per-project: **one of the Claude accounts must be used only on one specific project**; the remaining accounts may be used anywhere. There is no way to express that today, so the only thing enforcing it is the operator remembering — and an automatic switch under load will happily cross the line.

## Shape

- **A many-to-many binding between accounts and projects.** An account may be allowed on many projects; a project may allow many accounts.
- **Every selection for work belonging to a project draws only from that project's allowed set.** An account outside the set is never chosen — not as a fallback when the allowed ones are throttled either. That case reports "no allowed account has capacity" and parks, rather than quietly crossing the boundary.
- **Switching inside the allowed set stays automatic.** When one allowed account runs out of capacity, work moves to another allowed account without asking.
- **A project with no explicit binding keeps today's behaviour** — every account allowed — so nothing changes on the first deploy and the constraint is opt-in per project.
- **An agent can manage it.** An orchestrator reads the current bindings and changes them through the same control surface it uses for everything else, not only through the accounts UI. Reading back the record is what proves a change, never the action's echo.
- **Visible on both sides.** The accounts UI shows which projects an account is bound to; the project shows which accounts it may use, and which one is currently carrying its work.

## Why it matters now

Account choice is already automatic under capacity pressure, and the seat/pipeline machinery switches without asking. Once one account is reserved for one project, an automatic switch is a boundary violation the system currently cannot even represent, let alone refuse.

## Acceptance

- A project bound to a subset of accounts never runs on an account outside that subset, including when every allowed account is throttled — that state is reported, not worked around.
- A project with no binding behaves exactly as today.
- An orchestrator can list bindings, add one, and remove one, and each change is confirmed by reading the record back.
- Both directions of the relation are visible in the interface.
- Regression: a project bound to one account, with that account throttled and an unbound account idle, does not switch and reports the refusal.

## Related

- Account identity and switching work: #86 (a switch rebinds ownership, never copies), #954.
- The privacy rule applies to this issue's own follow-ups: accounts are named "account A / account B" in public artifacts, never by handle.

Latand:
Closes #1279

Every account was usable by every project and the active account was one global selection, so a per-project reservation was enforced only by the operator remembering it — and an automatic switch under capacity pressure would cross the line without asking.

The harder half was reach. `/api/spawn` could pin an account and a pipeline stage could not, so nearly all real work was unpinnable and honouring a reservation meant abandoning the Pipeline API and spawning raw. A settings table the launcher never consults would have satisfied the issue text and none of the requirement, so the binding is wired into the paths that actually launch work.

## The model

Many-to-many, and per engine: an account may be allowed on many projects, a project may allow many accounts. **A project with no binding for an engine allows every account of that engine** — the behaviour the Viewer has always had — so the first deploy changes nothing and the constraint is opt-in per project. The per-engine split matters because account ids are engine-scoped: binding a project to one Claude account must not silently forbid every Codex account from it.

`null` (no binding) and `[]` (bound to accounts that no longer exist) are deliberately different answers. The second one refuses rather than falling back.

## What consults it

- **A pipeline stage takes an `account`** (`PipelineStageInput.account`, published in the `create_pipeline` tool schema and settable through `override-stage`). It is checked at create, at `add-stage`, at `override-stage`, and again at the launch itself — `spawnPipelineAgent`, which is the seam that actually decides. A refusal there throws, so the stage parks with the reason on the record.
- **The automatic rate-limit switch draws from the allowed set only.** Review-loop reviewer selection passes the flow's project into `resolveHeadlessSpawn`. Every allowed account out of capacity reports *no allowed account has capacity* and parks; the idle account the project forbids is not a candidate and is not consulted.
- **Raw `/api/spawn`** refuses a pin outside the set (409), and its Claude health pass runs over the allowed accounts, so the fallback a failing probe picks comes from the same set the pin was checked against.
- **The orchestrator handoff digest** resolves its account against the project it is summarising for.
- **A workflow stage** resolves through the same project-fenced seam. This one was found while auditing the launchers rather than the store: it chose the engine's globally active account, so a bound project could still run its work outside its allowed set on the surface next to the pipeline one. It now parks with the reason when every allowed account is out of capacity, and takes the identical old branch when the project is unbound.
- **Reseating a rate-limited conversation** draws its successor from the allowed set. This was the sharpest remaining hole: `chooseReseatTarget` picked the healthiest account in the whole catalog, so a project bound to one account, that account throttled and an unbound account idle, moved the work onto the unbound one — the acceptance regression, happening on the surface the operator reaches for precisely when capacity is tight. It now reports the shortage and does not switch.
- **Reconfiguring a live conversation onto a named account** is refused when the project does not allow it, the same answer the spawn pin gets. Left open, it was a one-call door around every fence above.

Both resolve the conversation's project through one helper, which takes a fallback the callers already hold — the scanner's project for the transcript, or the cwd from its head. An **adopted** conversation carries an empty launch profile (no project, no cwd), so a fence reading only the profile would have resolved to `null` and silently not applied to most live conversations.

Switching *inside* the allowed set stays automatic, exactly as today.

## Managing it

`account_project_binding` (MCP) and `GET`/`POST /api/account-project-bindings` list, add and remove. **Every answer is a read of the store taken after the write** — the mutation returns the re-read record, and the store refuses to report `ok` when that read does not contain the change (`NOT_CONFIRMED`, kept distinct from a refused write). A caller never has to trust an echo.

## Visible both ways

The accounts panel names the projects each account is bound to. The project header names the accounts the project may use and marks which of them are carrying its work. A carrier outside the allowed set — a session that started before the binding existed — is still shown and still marked, so the fence never looks retroactive.

## Verification

- New, run by path under an isolated `HOME`/`XDG_CONFIG_HOME`/`LLV_STATE_DIR`/`TMPDIR`: `projectBindings.test.ts` (8), `projectSelection.test.ts` (8), `projectAccountsView.test.ts` (5), `pipelines/stageAccountBinding.test.ts` (8), `mcp/accountProjectBinding.test.ts` (4), `api/account-project-bindings/route.test.ts` (4), `ProjectAccounts.render.test.tsx` (6) — all green.
- `accounts/reseatBinding.test.ts` (5) and `reconfigureBinding.test.ts` (4) cover the two seams above, including the acceptance regression at the reseat path. The reseat fence was confirmed **red without it**: the unfenced selector, given that exact fixture, returns the forbidden account.
- The acceptance regression is `projectSelection.test.ts`: one project bound to one account, that account throttled, an unbound account idle → `exhausted` is reported and no switch happens. The same fixture **with the binding removed does switch**, which is what makes the first assertion a fence rather than an accident of the fixture.
- `workflows/engine.test.ts` gains two cases for the launcher fence (23 in the file, all green). Both were confirmed **red against the unfixed engine** — the capacity case leaves the workflow `implementing` on a forbidden account instead of `needs_decision` — so the green is attributable rather than assumed.
- Existing suites re-run by path, all green: `pipelines/engine.test.ts`, `pipelines/store.test.ts`, `pipelines/controller.test.ts`, `flows/engine.test.ts`, `flows/reviewerPolicy.test.ts`, `accounts/headlessSelection.test.ts`, `orchestrator/handoffDigest.test.ts`, `mcp/server.test.ts`, `mcp/bindings.test.ts`, `mcp/schemaParity.test.ts`, `mcp/toolAllowlist.test.ts`, `mcp/presentation.test.ts`, `api/spawn/route.test.ts`, `hooks/useEngineAccounts.test.ts`, `AccountsPanel.dom.test.tsx`, `AccountsPanel.render.test.tsx`, `pipelines/pipelineModel.test.ts`, `i18n.test.ts`, both `routeExports.test.ts`, and every `ProjectDashboard.*` file.

Two things a reviewer should know rather than rediscover:

- `delivery.test.ts` has two failing cases (`{codex,claude} send follows verified tmux rollback ownership…`, failing on `title is required for every new spawn`). Both are **red at the merge base** (`8dba5cea`), verified in a throwaway worktree at that commit; they are unrelated to the account fence added to that file.
- `ProjectDashboard.launchClaim.dom.test.tsx` has one failing case. It is **red at the merge base** (`8dba5cea`) — verified in a throwaway worktree at that commit, and again on this branch with this PR's only edit to that component removed.
- `bunx eslint` cannot run in this checkout at all: `eslint-plugin-react` throws `contextOrFilename.getFilename is not a function` under ESLint 10, including on files this PR does not touch. Pre-existing and unrelated.

Nothing was deployed, restarted, or re-authenticated; no real account handle or id appears anywhere in the change.

## Follow-up round: two seams where the project resolution was not the careful one

Both found by auditing the launchers and the view against the fence, not the store, and both confirmed red before the fix.

- **A moved draft re-reads the binding.** `update-draft` can change a draft's `repoDir`, and a repository is a PROJECT. The allowed set travels with the project rather than with the plan, so a pin that was legal where the draft was written can be illegal where it lands — and the move did not re-read the binding, so a stage pinned to an account the destination forbids became a stored draft. The launch seam still refused it, so no boundary was crossed; what broke was the create-time promise that a plan the launch can only ever park is never stored. The move now runs the same reading create does, before any field is assigned, so a refusal leaves the draft where it was rather than half-moved. Red before: the move returned success with the forbidden pin in place.
- **An adopted conversation is still attributed to the project it carries.** The project view's carrier projection resolved a conversation's project from ownership and launch profile alone. An adopted conversation carries an empty launch profile — no project, no cwd — which is exactly why both fence seams pass a fallback; without one here, the account actually carrying a project's work was never marked, so the display side under-reported precisely the sessions an operator opens the strip to understand. The projection moved into `projectAccountsView` with the same fallback, and resolves a project only for BUSY conversations that have an account, so the transcript read is paid per open turn rather than per conversation the registry has ever known.

Companion cases pin the unchanged behaviour in both: a move whose pin the destination project allows still lands, and a conversation that names its own project never reaches the transcript at all.

### Verification for this round

Re-run independently rather than carried over, by path, under an isolated `HOME`/`XDG_CONFIG_HOME`/`LLV_STATE_DIR`/`TMPDIR`:

- `bunx tsc --noEmit --incremental false` — exit 0.
- `pipelines/stageAccountBinding.test.ts` (10), `accounts/projectAccountsView.test.ts` (8), `api/account-project-bindings/route.test.ts` (4), `api/routeExports.test.ts` (2), `pipelines/engine.test.ts` (208), `accounts/projectBindings.test.ts` (9), `accounts/projectSelection.test.ts` (8), `accounts/reseatBinding.test.ts` (5), `reconfigureBinding.test.ts` (4), `workflows/engine.test.ts` (23), `flows/engine.test.ts` (50), `mcp/schemaParity.test.ts` (18), `mcp/server.test.ts` (33), `mcp/accountProjectBinding.test.ts` (4), `mcp/toolAllowlist.test.ts` (17), `api/spawn/route.test.ts` (55), `accounts/headlessSelection.test.ts` (9), `orchestrator/handoffDigest.test.ts` (13), `hooks/useEngineAccounts.test.ts` (39), `i18n/i18n.test.ts` (18), `AccountsPanel.render.test.tsx` (26), `ProjectAccounts.render.test.tsx` (6) — all green.
- The two pre-existing failures noted above were re-confirmed at the merge base in a throwaway worktree this round, by the same test names: `delivery.test.ts` 2 fail / 34 pass, and `ProjectDashboard.launchClaim.dom.test.tsx` 1 fail / 1 pass — identical counts on the branch and at `8dba5cea`.

`chooseAutoBalance` was checked as a possible unfenced automatic switch and is referenced only by its own test, so no production path selects an account through it.

## Follow-up round: a damaged record refuses, and every write is one locked transaction

Two review findings, and both inverted the feature the same way — a project that reserved an account ended up allowing every account.

**A record that could not be read became an empty binding list**, and an empty list means unbound, which means every account. The condition where the reservation matters most — damaged, truncated, unsupported or unreadable — was the condition that silently removed it. The record now has three states and only one is unbound: ABSENT (no file, or no directory that could hold one) still allows every account, exactly as before; READABLE is enforced; anything else throws `AccountProjectBindingsUnreadableError` naming the record and the fault. The throw parks a pipeline stage with the reason on its record, parks a workflow stage rather than escaping into the tick that drives every other workflow, refuses a reseat, a switch and a raw spawn, and answers 409 from the bindings API. A mutation refuses without writing, which also leaves the damaged record intact for repair instead of replacing it with one row.

**The read-modify-write was unlocked across processes.** Two writers read the same record, each appended its row, and the later rename erased the earlier one while both reported success — a project whose sole binding was the lost row is open to every account again, the same inversion arrived at by a race. Read, mutation, write and confirmation now run inside the cross-process account mutation lock every other account write already takes, and the API route awaits the async form so an operator's click queues instead of being turned away. Two pieces of machinery went with it: the bespoke mtime+size cache, which could not see a same-millisecond same-size write by another process, and the hand-rolled temp-file-and-rename, replaced by the shared durable JSON write.

`GET /api/accounts` is deliberately left to fail while the record is damaged rather than serve an account whose project list reads as empty, and the project strip degrades to absent rather than to "nothing is restricted".

### Verification for this round

Re-run independently, by path, under an isolated `HOME`/`XDG_CONFIG_HOME`/`LLV_STATE_DIR`/`TMPDIR`:

- `bunx tsc --noEmit --incremental false` — exit 0.
- `accounts/projectBindings.test.ts` (27), `accounts/projectBindings.interprocess.test.ts` (3), `api/account-project-bindings/route.test.ts` (5), `accounts/projectAccountsView.test.ts` (8), `accounts/projectSelection.test.ts` (8), `accounts/reseatBinding.test.ts` (5), `mcp/accountProjectBinding.test.ts` (4), `reconfigureBinding.test.ts` (4), `pipelines/stageAccountBinding.test.ts` (11), `pipelines/engine.test.ts` (208), `workflows/engine.test.ts` (23), `accounts/accountMutation.test.ts` (7), `mcp/schemaParity.test.ts` (18), `mcp/server.test.ts` (33), `AccountsPanel.render.test.tsx` (26), `ProjectAccounts.render.test.tsx` (6) — all green.
- `bun scripts/privacy-publication-gate.ts --check-commits` — PASS against both the merge base and the default branch tip.
- **The new tests were watched going red.** All three multiprocess cases fail against the pre-fix store in a throwaway worktree at the previous commit: sixteen concurrent adds leave three accounts that answered `ok: true` with no row on record, concurrent adds and removes produce two `NOT_CONFIRMED` answers, and a record damaged mid-flight answers `ok` to the intruding add instead of refusing it.
- `accounts/manager.interprocess.test.ts` has one failing case, and it is **red at the merge base** (`8dba5cea`) with the identical assertion and 3s child-readiness timeout — verified in a throwaway worktree at that commit. `mcp/server.test.ts` failed one liveness case on a single run while another suite ran alongside it, then passed three consecutive full-file runs and in isolation.

Nothing was deployed, restarted, or re-authenticated.


Latand:
Follow-up commit `18e52b5d`: bindings are stored and read under the project's **canonical** id.

`projectForCwd` answers with the scanner's project, which can be an alias *source* for a project whose repository identities have since converged, while the accounts panel and the MCP tool both canonicalize before writing. Written from one spelling and read from the other, the fence would have silently not applied to the pipeline it was meant to fence. Both sides now go through the same key, with a case in `projectBindings.test.ts` that writes from the alias source and reads from the target.

Re-run after the fix, all green: `projectBindings` (9), `projectSelection` (8), `projectAccountsView` (5), `pipelines/stageAccountBinding` (8), `mcp/accountProjectBinding` (4), `api/account-project-bindings/route` (4), `pipelines/engine` (208), `flows/engine` (50), `orchestrator/handoffDigest` (13). `bunx tsc --noEmit --incremental false` clean; privacy gate PASS at HEAD.

Latand:
Follow-up commit `ae4c4ed3`: **the automatic side, closed at the one seam that never asked.**

### What the audit found already done (nothing re-done)

- A damaged record no longer fails open — only a *confirmed absent* pathname is unbound (`ENOENT` re-checked with `lstat`, so a dangling link is damage).
- The structured reconfigure consults the binding before dispatch, to **attribute** rather than to refuse.
- A binding survives its project being renamed under it — rows canonicalize on the way out as well as in.

### The finding that was still open

`POST /api/tasks/[id]/spawn` — the board's task **spawn** and **retry** buttons — never read the record. The client sends no `accountId` at all, so the account is entirely the Viewer's own pick, which is exactly what a pool binds. Reproduced against the previous commit: a project bound to one account still launched its task on the machine's default, with a pane, two assignment writes and a durable receipt a retry could replay.

It now resolves through `resolveProjectSpawnAccount` — the same project seam every other launch uses. An unbound project keeps the branch it always took; a bound project draws from its allowed set only; an allowed set with no capacity is reported; an unreadable record refuses.

The direct launch (`POST /api/spawn`) already refused on a damaged record, but as a server fault — which reads as a Viewer crash and leaves nothing to repair from. It now gives the same conflict the reseat, the binding route and the task launch give, carrying the record's name.

### Earlier findings re-judged against the corrected rule

| finding | verdict |
|---|---|
| structured reconfigure dispatches a forbidden / damaged-record switch | **void** — a named switch is a control; the override is required behaviour |
| legacy reconfigure dispatches the same through conversation migration | **void**, same reason |
| the override journal and attribution UI are an over-built bypass | **void** — attribution is the price of the capability, not a bypass |
| a binding disappears when its project alias arrives after the row was written | closed by `ae9b8a50` |
| persisted project keys with surrounding whitespace open the canonical project | closed by the same canonicalize-on-read |
| damaged-record exceptions escape the reseat and legacy-switch contracts | closed earlier; the remaining two boundaries (direct launch, task launch) are closed here |

### Every path where the system picks an account by itself

pipeline stage · flow reviewer (pane + headless) · workflow stage · one-click reseat · direct spawn · task spawn · orchestrator handoff digest — each draws from the pool, reports and parks on no capacity, and refuses on an unreadable record. The engine-wide auto-balancer (`chooseAutoBalance`, `createMigrationIntent`) has no production caller; the only live limit-triggered migration is the conversation reseat, which is fenced.

### Gates

`bunx tsc --noEmit --incremental false` clean. Privacy gate **PASS** at HEAD against the merge base. Tests by path, isolated `HOME`/`XDG_CONFIG_HOME`/`LLV_STATE_DIR`/`TMPDIR`: task-launch binding 4, spawn-route binding 2 (both new, each verified RED against the previous commit in a throwaway worktree), task spawn route 12, spawn route 55, projectBindings 32, projectSelection 8, projectAccountsView 12, accountOverrides 7, reseatBinding 5, reseatCommandBinding 2, projectBindings.interprocess 3, stageAccountBinding 11, mcp/accountProjectBinding 4, account-project-bindings route 5, accounts route 21, tmux route 28, reconfigureBinding 6, structuredControls 31, workflows/engine 23, flows/engine 50, pipelines/engine 208, handoffDigest 13, schemaParity 18, ProjectAccounts 9, AccountsPanel 26 — all green.

Two pre-existing reds, both red at the previous commit as well and untouched by this change: `accounts/manager.interprocess.test.ts` (1 of 4, a controller selection race), and `eslint` itself cannot run in this tree (`react/display-name` fails to load under the installed plugin pair).

Latand:
Follow-up commit `1775bc4c`: **the automatic side, closed on the side that refuses as well as the side that crosses.**

### What the audit found already done on this branch (nothing re-done)

- **A damaged record no longer fails open.** Only a *confirmed absent* pathname is unbound — `ENOENT` is re-checked with `lstat`, so a dangling link counts as damage, and every other errno raises. The old fail-open expectation is gone.
- **The structured reconfigure consults the binding** before dispatch, to attribute the choice rather than to refuse it.
- **A binding survives its project being renamed under it** — rows canonicalize on the way out as well as in, so a row written before its own alias existed is still that project's row.
- Every launch seam already draws from the pool: pipeline stage, flow reviewer (pane and headless), workflow stage, one-click reseat, direct spawn, task spawn, orchestrator handoff digest. Re-verified this round; the engine-wide auto-balancer (`chooseAutoBalance`, `createMigrationIntent`) still has no production caller, and the only live limit-triggered migration is the conversation reseat, which is fenced.

### Earlier findings, re-judged against the corrected rule

| finding | verdict |
|---|---|
| structured reconfigure dispatches a forbidden / damaged-record switch | **void** — a named switch is a control, and the override is required behaviour |
| legacy reconfigure dispatches the same through conversation migration | **void**, same reason |
| the override journal, actor plumbing and attribution UI are a prohibited bypass | **void** — attribution is the price of the capability |
| capacity-pressure reseat can cross a project's allowed set | closed — the reseat resolves through the project's pool and reports when it holds no successor |
| unreadable or malformed binding state reads as no binding | closed — every read raises, and only a confirmed-absent pathname is unbound |
| task-assignment and workflow launchers bypass the binding | closed |
| binding mutations lose each other with no interprocess lock | closed — read, mutate, write and confirm under the shared account mutation lock |
| a binding disappears when its project alias arrives after the row was written | closed |
| persisted project keys with surrounding whitespace open the canonical project | closed by the same canonicalize-on-read; normalizing a key *binds* a project, which narrows rather than widens |
| damaged-record exceptions escape the reseat and legacy-switch contracts | closed — both answer a conflict naming the record |

### What this commit fixes

**A continuity hint was travelling as a pin.** The task launch passed the account a *first* launch happened to resolve — the one an assignment keeps so retries land where the work started — into the project seam as `requestedId`. Nobody ever named that account, so a project bound elsewhere since then **refused its own task launch while the pool it was given sat idle**. The automatic path was declining to draw from the pool, which is the one thing the rule asks it to do.

`ProjectSpawnRequest` now separates the two. A **preference** orders the allowed candidates and loses to the fence; a **named** account is still refused rather than quietly substituted. An unbound project is byte-identical, down to still reporting an account that no longer exists as unknown rather than starting somewhere else.

**A bound project whose pool yields no usable account is now reported at the direct launch** instead of raised. It answered `500` with the resolver's message — a Viewer fault for a state the binding itself created. It now answers `409` in the wording every other seam uses, with the resolver's own reason after it so the operator learns which account to repair. An unbound project keeps the throw it always had: its failure is about the machine's accounts, not about a boundary anyone drew.

Both new tests were verified RED against `d4bcad9b` by reverting only the source line under test, and both continuity tests stay green under the old and new code.

### Where the line was drawn

The override covers a deliberate **switch** that names an account, which is what the corrected rule names. A **launch** that names an account — a pipeline stage's `account` pin, `POST /api/spawn` with `accountId` — is still refused when the project forbids it, and refused at create/override time for a stage so the plan cannot be stored unrunnable. Say the word if a named launch should be carried out and attributed like a named switch; it is the same seam and a small change.

### Gates

`bunx tsc --noEmit --incremental false` clean. Privacy gate **PASS** at HEAD against `origin/main`, with commit checking and known-value fingerprints.

Tests by path under an isolated `HOME`/`XDG_CONFIG_HOME`/`LLV_STATE_DIR`/`TMPDIR`, 28 files, **628 pass / 0 fail** when each file runs alone: task-launch binding 6, spawn-route binding 5, task spawn route 12, spawn route 55, tmux route 28, accounts route 21, account-project-bindings route 5, projectSelection 8, projectBindings 33, projectBindings.interprocess 3, managerProjectBinding 7, accountOverrides 7, projectAccountsView 12, reseat 2, reseatBinding 5, reseatCommandBinding 2, spawnHealth 13, flows/engine 52, workflows/engine 23, pipelines/engine 208, stageAccountBinding 12, structuredControls 31, reconfigureBinding 6, mcp/accountProjectBinding 4, schemaParity 18, handoffDigest 15, ProjectAccounts 9, AccountsPanel 26.

One failure appears only when the 28 are run in **one** invocation, and it is pre-existing: a file-order-dependent `LLV_STATE_DIR` leak between two suites this branch does not touch, reproduced identically at the merge base `8dba5cea`. Filed as #1299.

Latand:
Follow-up round `b99f0b8d..ebb97f34`: **the eleventh selector goes through the admission, and attribution follows acceptance.**

A deploy severed the previous host mid-turn. Its work was preserved as `b99f0b8d`, marked rescued and unreviewed and carrying no tests. This round reviewed it, held every behaviour it changed down with a test, and closed the one surface it left saying the opposite of what happened.

### The eleventh selector: settlement-time enrollment

A spawn that SETTLES while an engine-wide drain is running is enrolled into it at settlement. Nobody named an account for that enrollment — the spawn was reserved on its birth account before the drain existed — so for an **automatic** intent it is an automatic selection, and it was the one that asked nothing: no binding, no capacity, and it built its own migration record **beside** the shared admission rather than going through it. A spawn reserved before the drain therefore migrated one project onto a target only another project allows, and an exhausted target or an unreadable record passed the same point untouched.

It now calls `admitAutomaticAccountTarget` — the same function the drain loop and the lazy active-account reseat call, with the same project key, the same observations and the same bindings. A **manual** intent NAMED its target: that is a control, it is carried out, and enrollment there reads no binding at all, exactly as before.

A refusal **parks**. The spawn has already happened, so what is refused is the routing change: the settlement stands and the conversation stays visibly on the account it was born on. That is also why no read failure escapes the admission call — evidence this process cannot turn into a pool must not be able to lose a spawn, so every failure answers "not admitted", the same direction the fence fails everywhere else.

**What was cut.** The settlement built a fourteen-field `ConversationMigration` literal of its own, the fourth copy of a record the two paths that queue the same move already construct through `conversationMigrationForIntent`. It is gone; settlement now calls the shared construction after admission. The two are field-for-field identical at this call site — the guard requires no existing migration, so every "carry the previous record's value" branch in the shared version falls through to the literal's own expression — so the cut moves no behaviour and removes the place a fifth copy would have drifted.

`projectSelection.ts` names it as (11) in the inventory, so the SET stays the thing under review rather than the next member of it.

### Attribution follows acceptance

Both switch seams appended the override record **before** the switch was authenticated, reserved and dispatched, and a storage failure was reduced to a bare `false` nobody read. For a feature whose whole point is to permit the out-of-pool choice **and show it**, that failed in both directions: the project view could show a choice that never happened — a switch onto a signed-out account, and an appending journal has no compensating delete — and could silently lose one that did.

- **The legacy switch** records after the reseat is durably queued, outside the `try`, because nothing about writing the record may turn an accepted switch back into a failure.
- **The structured reconfigure** builds the attribution as a closure and calls it only once the host holds a receipt — on the success path, and on the transport-failure path where `operationStatus` proves the command was accepted. A reconfigure the host refuses records nothing.
- **The journal names its failure.** `appendOverride` returns the reason instead of `false`; the notice carries `recorded: false` and `recordFailure`; the Viewer logs it so the failure is on the record even when nobody is watching the answer. There is no retry, on purpose: the write already queues behind the journal's file transaction for as long as that transaction waits, so what reaches the catch is a durable failure of the state directory itself.
- **The badge stopped saying the opposite.** It answered every out-of-pool switch with "recorded as your choice", including the ones whose record did not land — the state directory being unwritable is exactly when the operator has to hear it, and instead they were told the thing this journal exists to guarantee. A switch that was recorded says so as before; one that was not is an error toast naming what happened.

The choice itself still stands in every case. A record that would not write is not a decision anybody made.

### Verification

`bun install` was not needed. Everything below ran by path under an isolated `HOME`/`XDG_CONFIG_HOME`/`LLV_STATE_DIR`/`TMPDIR`.

- `bunx tsc --noEmit --incremental false` — exit 0.
- Touched: `agent/registry.migrationBinding.test.ts` (14), `accounts/accountOverrides.test.ts` (8), `reconfigureBinding.test.ts` (8), `runtime/structuredControls.test.ts` (33), `AccountBadge.dom.test.tsx` (7) — all green.
- Regression set, all green: `accounts/projectSelection` (8), `accounts/projectBindings` (33), `accounts/reseatBinding` (5), `accounts/reseatCommandBinding` (2), `accounts/projectAccountsView` (12), `accounts/managerProjectBinding` (19), `accounts/headlessSelection` (9), `pipelines/stageAccountBinding` (12), `agent/registry` (123), `agent/registry.admission` (11), `agent/registry.reseat` (24), `agent/migration` (5), `agent/reconfigure` (4), `accounts/migration/controller` (8), `accounts/migration/intentLiveness` (4), `conversation/actions` (6), `ProjectAccounts.render` (9), `AccountsPanel.render` (26), `i18n` (18).
- `bun scripts/privacy-publication-gate.ts --require-known-values --check-commits` — **PASS** against both the merge base and the default-branch tip.

**Every new case was watched going red**, in throwaway worktrees, against the sources it was written for:

| subject | red against | result |
|---|---|---|
| `registry.migrationBinding.test.ts` | pre-rescue registry (`66f1ece3`) | 11 pass / **3 fail** — disjoint pool, exhausted target, unreadable record. Allowed, unbound and manual-intent pass either way, which is what makes those three a fence rather than an accident of the fixture. |
| `accountOverrides.test.ts` | same | 7 pass / **1 fail** |
| `reconfigureBinding.test.ts` | same | 6 pass / **2 fail** — the refused switch, and the failure to record |
| `structuredControls.test.ts` | same | 32 pass / **1 fail** — the refused reconfigure. The durable-receipt case is the acceptance path that must not change and passes either way. |
| `AccountBadge.dom.test.tsx` | HEAD with only the badge toast reverted to its pre-fix form | 6 pass / **1 fail** — the unrecorded switch. The recorded one and the in-pool one pass either way. |

Three pre-existing failures were re-confirmed at the merge base (`8dba5cea`) in a throwaway worktree this round, by identical test names and counts, and none of them is in this lane's files:

- `agent/spawnProjection.test.ts` — 31 pass / 4 fail here, **31 pass / 4 fail at the merge base**, all four failing on `title is required for every new spawn`.
- `delivery.test.ts` — 35 pass / 2 fail here, **34 pass / 2 fail at the merge base**; the two failing names are identical and the extra pass is a case this branch added to that file in an earlier round.
- `accounts/migration/coordinator.test.ts` — 105 pass / 9 fail here, **105 pass / 9 fail at the merge base**, the same nine names.

Nothing was deployed, promoted, rolled back or restarted, and no real account was switched, retired or re-authenticated.

Latand:
Orchestrator decision, 2026-08-30: the accounts-binding continuation exhausted its rounds with two P1s left — automatic migration retries re-enrolling disjoint/exhausted/unreadable targets without shared admission, and a dangling state-directory symlink read as confirmed absence. Next round is structural: inventory every autonomous account-selection site (settlement, startup adoption, retry, drain, reseat) in one table with a test that walks it, then fix the two findings. Held until the delivery-drain outage in #1314 is resolved.

Latand:
Closing as superseded by #1328, merged to main on 2026-08-30 as 51a3db57. That PR closed #1279, the issue this one targeted.

Checked against current main (4718eb53):

- This PR's head `ebb97f34` is an ancestor of #1328's head `624d1242`. #1328 continued this branch for 18 more commits.
- #1328's squash commit applies exactly that head's change set: same patch-id, same tree.
- All 25 files this PR adds are on main.

The branch is kept.

PR #1294, opened 2026-08-30
fix(delivery): every accepted send settles when it is asked about, and an uncertain one stays uncertain (#1131)

Latand:
Successor to the lane on #1131 (supersedes #1292, whose commits this branch carries — close that one rather than merging it).

Two rules govern this fix, and the second matters more than the first:

1. **`queued` is never a final answer.** Every accepted send reaches delivered or failed, and a caller holding an operation id can ask what became of it.
2. **An uncertain send stays uncertain.** Once actuation has started the state is absorbing: never reclassified as unexecuted, never advertised as safely resendable. Delivering a deployment instruction twice is a worse incident than the silence this issue was opened for.

## Settlement happens where the question is asked

The previous rounds ended a send with a background sweep — an interval in the release that owns traffic, a retrying startup controller, a batch ceiling, a candidate classifier. All of it worked only while the runtime host it had to ask was reachable, which is the one condition a lost send does not meet: lasting unavailability left `queued` as the effective final answer for the whole outage. A fix for "this can hang forever" had acquired its own way to hang.

`message_receipt` is the settlement now. It reads the journal's current answer, and past the deadline it **ends** the send from whatever evidence exists:

- an operation the journal still holds open is fenced (terminalized `failed`, clearing its outbox row in the same transaction) and reported `lost` — the one case that may answer `resend: "safe"`;
- one an executor already took terminalizes `uncertain` and reports `verify-first`;
- a journal with no record of it at all, and a runtime host that cannot be reached at all, both settle unverified — an absent or unreadable record is not a proof of absence.

The deadline is the reservation's own acceptance time, so nothing has to be running for it to pass. A send queued behind the recipient's live turn is progressing rather than lost and stays exempt, up to the ceiling that stops a wedged host from making `queued` permanent again.

An answer given during an outage has to stay true afterwards, so the settled record is itself the fence: the delivery queue reads it before actuating a message effect and hands the engine nothing. That is what makes "one terminal receipt, zero late deliveries" hold across a socket that was gone for hours, with nothing healthy in between. It is asked under the effect's **own** operation id, so a deliberate retry — admitted as a new operation — is not fenced by the attempt it replaces.

Deleted with the sweep: the interval and its process-global timer, the startup controller and its backoff loop, the batch ceiling, the sweep's report types and candidate classifier, and the delivery-queue port helper that had no fence to consult. `viewerInstrumentation.ts` and `instrumentation.test.ts` are back to their state on the default branch.

## Actuation is what makes a fate unknown

Kept from the previous round, and unchanged: a send handed to a host that then failed to answer terminalizes `uncertain` whether that host died or is still answering health checks, because the only thing that could tell the two apart is the confirmed-delivery record the call just failed to get. The thread/read timeout keeps its `queued` retry — it is reissued as the same operation, which the host dedupes by queue-entry id. An effect whose durable receipt already says `delivering` was actuated by an executor that never settled it, so the drain terminalizes it unverified instead of delivering it again.

## The legacy pane path is absorbing too

A legacy write that began and never came back left a `delivery-uncertain` reservation, and the same client request reassigned it and typed the payload a **second** time — one retry putting the same instruction in front of an agent twice, on a path with no journal that could ever say whether the first landed. The replay is answered from that uncertainty instead: the reservation's operation id, `verify-first`, and nothing typed. The regression drives `deliverConversationMessage` twice and asserts one host write. The artifact-replay machinery that existed only to feed that retry is gone.

## Two acceptances nobody could ask about

- A hold answers `/api/runtime/send` with its reservation's operation id rather than `held: true` alone, so the composer's own path is queryable like every other acceptance, and both terminal outcomes are covered through the receipt query.
- A `bridge_directive` the runtime merely accepted parks its answer against that operation id. The manager's open decision request clears itself once the send is recorded delivered, and stays standing when it is recorded failed — the trade the previous round took only half of, which left a queued directive's ask open forever after the message did arrive. The parked ref passes the same project-and-seat fence a recorded answer does, and the ask projection joins the two on every read, so nothing polls and nothing can close an ask twice.

## Verification

- `bunx tsc --noEmit --incremental false`: passes.
- Touched files by path, under an isolated `HOME`/`XDG_CONFIG_HOME`/`LLV_STATE_DIR`/`TMPDIR`: 349 pass, 4 fail. All four fail identically at the default branch tip in a clean worktree (two legacy `verified tmux rollback ownership` spawn-title cases in `delivery.test.ts`, and the two `migration-held delivery switches from the source host to the published … successor` integration cases, whose successor ledgers are empty at the base commit too and whose deliver callback is never reached). Adjacent suites (`structuredDeliveryController`, `api/tmux/route`, `api/files/route`): 116 pass, 4 fail, identical at the base tip.
- Both new fences were verified red: disabling the record fence fails the recovery regression; disabling the parked directive answer fails the ask regression.
- Privacy gate with commit checking, against both the merge base and the default-branch tip: pass.
- Nothing was deployed, promoted or restarted.


Latand:
Round update, at `200c19f5ad6fe2dca1f4c790e8f6d5fe30dab1f8`. Findings 1, 3, 4, 5 and 6 are addressed; finding 2 is refuted with base evidence. The PR body is rewritten to the fix as it now stands.

**6 (over-built) drove the round.** The sweep is gone: the interval and its process-global timer, the startup controller and its backoff loop, the batch ceiling, the report/candidate types, and the delivery-queue port helper that had no fence to consult. `viewerInstrumentation.ts` and `instrumentation.test.ts` are byte-identical to the default branch again. What remains is one journal-to-registry projector (the controller's transition) and the receipt reader.

**3 is closed by that same move, not around it.** `message_receipt` is the settlement now: past the deadline it ends the send from whatever evidence exists, including none — a runtime host that cannot be reached still yields one terminal, unverified answer. The settled record then fences the delivery queue, so recovery after an outage of any length produces zero deliveries. The regression drives that end to end over a real journal and a real queue, and it is red without the fence.

**2 is wrong-premise.** Both `migration-held delivery switches from the source host to the published … successor` cases fail identically at the merge base `8dba5cea` and at the default-branch tip `95079734` in a clean worktree, where the `delivering` guard does not exist; the successor ledgers are empty there too, and the deliver callback is never reached, so the drop is upstream in migration reconciliation. Filed as #1295 along with the six other pre-existing reds. The guard did have a real regression the round found separately — keyed on the presentation operation id it fenced a deliberate retry (`repeated terminal retry clicks produce one replacement engine write`); it is keyed on the effect's own operation id now.

**1** moves the legacy fix into `deliverConversationMessage` where the finding asked for it: a two-call fixture, one host write, `verify-first` with the operation id, nothing typed on either replay. **4** parks a queued directive's answer against its operation id, so the ask clears on delivered and stays open on failed. **5** returns the hold's operation id from `/api/runtime/send`.

Gates: `tsc` clean; touched files by path 498 pass / 8 fail, all eight identical at the default-branch tip; privacy gate with commit checking passes against both the merge base and the tip. Nothing deployed, promoted or restarted.

Latand:
Superseded by #1297, which carries this work forward on a continued branch after a redeploy severed the original lane's host. Closing so the board shows one live line of work for the send-settlement fix rather than three.

PR #1297, opened 2026-08-30
fix(delivery): an accepted send settles without a healthy runtime, and an owned one is never called abandoned (#1131)

Latand:
Successor to the send-settlement lane on #1131 (supersedes #1294, whose commits this branch carries — close that one rather than merging it; it in turn supersedes #1292).

Two rules govern this fix, and the second matters more than the first:

1. **`queued` is never a final answer.** Every accepted send reaches delivered or failed, and a caller holding an operation id can ask what became of it.
2. **An uncertain send stays uncertain.** Once actuation has started the state is absorbing: never reclassified as unexecuted, never advertised as safely resendable. Delivering a deployment instruction twice is a worse incident than the silence this issue was opened for.

## Termination no longer needs a healthy runtime

A send admitted while the runtime host was unreachable is **accepted** — durable reservation, operation id handed back — but no journal operation exists for it. The settlement deadline covered only the state a reservation reaches by being ACTUATED, so that acceptance rested unsettled until a drain that only runs while the runtime is healthy picked it up: the original bug, rebuilt inside its own fix, on the one path whose only terminal answer required the health the lost send did not have.

The deadline now covers the accepted-but-unactuated reservation too and ends it from the durable record alone — `unverified`, because an outage is a reason to know less and never a proof that nothing executed. The record is terminal, so the recovered runtime hands the engine nothing.

## An owned send is not an abandoned one

A `delivering` row was read as proof of an abandoned executor on its own. It is not: a send another executor is actuating right now looks exactly the same, and during a release succession two executors are briefly alive over one journal — so the fence that exists to prevent a second delivery was terminalizing first deliveries still in flight.

The executor stamps itself and the writer claim it delivers under onto the transition, and a later pass calls the row abandoned only where that ownership has actually **changed**: a claim that has moved on cannot write to the engine and cannot answer for the send either, so nothing but this pass is left to end it. Its own abandoned row is never "someone else's", so ownership can delay an answer and never withhold one — and past the deadline a receipt query ends it regardless. The evidence is the writer claim the registry already keeps; the carrier is the receipt reason the journal already persists per transition.

## Settling a send stopped being able to re-deliver it

An ambiguous legacy send was absorbing while its reservation said `delivery-uncertain`, and a receipt query past the deadline ends it as an unverified failure — whose exact-replay contract is "revive and deliver again", because an ordinary failure is recoverable that way. The query that made the send terminal was therefore also what made an exact replay type the message a second time. A terminal failure that says the message may already have arrived is now absorbing wherever `delivery-uncertain` was, asked of the record's own disposition rather than of one state name.

That evidence has to exist for the ordinary send too, and it did not: the durable row was skipped exactly for the reservation that generates its own operation id, which is the common shape. Deleted, so every accepted send has one row keyed by the id its caller holds, and within one bounded retention the rows that say a send may have arrived are the last to be dropped.

## The remaining roads by which `queued` could be final

- The **operation read API** reported the journal's raw state and nothing else: `queued` while the socket was up, 503 once it was gone. It reads the same settlement `message_receipt` does, and answers from the durable record when the journal cannot be reached. Absent plane and unreachable plane stay distinct answers.
- A send **no structured delivery owned** was admitted straight into the journal with no reservation behind it — nothing to settle it, nothing to answer for it. Refused at send time instead. Controls on the same path reserve nothing and are untouched.
- The **ambiguous legacy send** answers with its operation id and `verify-first` from the first attempt, not only from replays, and the MCP control client carries both through instead of flattening the error body into a message.

## The crash regression runs a real process

The dead-executor regression faked the death by throwing inside a transition. A second interpreter now runs the production queue over the production journal, records its engine write where the parent can count it, holds the send open, and is killed with SIGKILL in exactly that gap. One engine write in total, an absorbing unverified receipt, an empty outbox — and without the fence the successor writes the instruction a second time.

## What was deleted

- The owner-row exception for generated operation ids, on both the load and compaction paths.
- The reservation-less direct send admission.
- The state-name test for absorbing replay, replaced by the record's own disposition.

No sweep, no timer, no background reconciler: settlement happens where the question is asked, and the deadline it applies is the reservation's own durable acceptance time.

## Not fixed here, and why

Two migration-successor regressions in the structured-delivery integration suite fail on this branch. They also fail on `main` (verified at its tip) and at this lane's merge base, and they never reach the delivery queue: committing a migration cancels every non-terminal held delivery for the conversation, so the reconciler is never asked to deliver one. That cancellation is deliberate and separately asserted — preserving held deliveries at commit turns 18 currently-green reseat assertions red and still does not make these two pass, because the coordinator's committed branch drains only assigned and uncertain reservations. Reconciling the two contracts is a change to the account-migration coordinator, which another lane owns.

Refs #1131


Latand:
Review verdict recorded, and the follow-up lane is running, 2026-08-30.

This branch got the review it had been missing. Three findings, and they are the same rule broken in three places — **absence of evidence taken as evidence**, which is the pattern that has recurred across this codebase all night.

1. An absent or failing ownership-claim read becomes `null` and is then treated as an ownership mismatch, so a transient gap in the claim projection is enough to terminalize a send another executor is actuating right now. The fix is tri-state: a matching claim stays live, an explicit differing claim proves abandonment, and unreadable evidence leaves the row with its executor until the receipt deadline.
2. The operation query can answer a terminal send alongside a queued receipt in the supported half-outage where the journal read succeeds and its fence write fails. One consistency rule has to decide, and it has to be stated.
3. A retry's accepted operation id has no durable settlement record, so a caller holding the id it was handed can ask about something nothing will ever settle.

What the review confirmed and what must not regress: a send admitted while the runtime was unreachable terminates from the durable record alone; an ambiguous legacy send settled to failed can no longer make a replay revive and re-deliver it; the crash regression kills a real process; and machinery was removed rather than added — no sweep, no timer, no standalone settlement controller.

The two migration-successor regressions fail here **and at the merge base**. They turn on a contract conflict owned by another lane and are tracked separately; they are not this branch's debt and the follow-up lane is instructed not to touch them.

PR #1305, opened 2026-08-30
fix(delivery): resume reclaimed hosts before delivery (#1300)

### LINKED ISSUE #1300: fix(delivery): a message to a live conversation whose host was reclaimed is refused as transient, when the resume that would fix it already exists one layer down
Reported by a peer orchestrator seat on another project and relayed on the operator's instruction, 2026-08-30. Reproduced and diagnosed here.

## What the caller sees

`send_message` to a live conversation fails, twice, with:

```
structured host ownership is unavailable; retry after runtime synchronization
```

The retry used the **same** idempotency key, per the tool's own guidance never to re-send an unsettled operation under a new id, and came back `replayed: true` with the identical error. So the replay path is correct; the underlying condition simply never clears.

The target was a long-lived session deliberately parked waiting for a human answer for about 24 hours.

## What is actually true, from the record

The conversation is live, and its host is **gone and terminal** — not synchronising:

```
activity: idle    proc: killed    pid: none
turn:  terminal, since ~22 hours before the failure
last transcript write: ~28 hours earlier
```

**So the error message is wrong in the way that matters.** It describes a transient window and invites a retry, while the real condition is a permanently reclaimed host that no amount of waiting will restore. A caller following the message retries forever.

## The capability already exists, one layer down

`POST /api/tmux` with `{"action":"resume"}` on the same conversation answered:

```
{"ok":true,"outcome":"resumed","spawned":true,"structured":true}
```

with no content injected. So the Viewer can re-establish a host for exactly this conversation — `send_message` just refuses instead of doing it. `send_message_to_orchestrator` already documents that it resumes a dead selected conversation; the general path does not, and there is no reason for the two to differ.

## Wanted

1. **Deliver by resuming.** A message to a live conversation whose host was reclaimed re-establishes the host and delivers, the way the orchestrator variant already does.
2. **Let a caller ask.** A way to learn whether a conversation is currently deliverable, so "wait and retry" can be told apart from "this is gone, re-spawn". Today the caller's only options are to guess or to abandon the session and lose its accumulated context.
3. **Say which it is.** Distinguish a genuine synchronisation window from a reclaimed host in the error itself. The present wording reads as transient and, hours later, still is not.

## The wider family, which is the reason to fix this properly

The same seat reports that `spawn_agent` has returned `ok` with a conversation id **three times** while no session or transcript ever appeared — detectable only by polling for a transcript rather than trusting the answer. And during the resume above, the call answered `spawned: true` while no process was visible for some minutes afterwards.

That is one shape, not three bugs: **the delivery layer reports success, or a retryable state, while the runtime side never materialises.** #1123 is the spawn half of it.

## Related

- #1123 — a structured builder host wedges at spawn with no transcript ever created.
- #1296 — a stranded host still reads as live, and a redeploy strands hosts.
- #1298 — an unreadable evidence source silences the whole wake.

Latand:
## Impact

A message can continue a long-lived conversation after its runtime host was reclaimed. The sender receives one durable operation and can follow it through message_receipt while recovery and delivery complete.

## Cause

The structured send path treated a missing or unhosted runtime projection as unavailable ownership. A durable conversation whose host record showed no deliverable process never reached the recovery path, even though conversation resume already supported that state.

## Fix

- Reserve the instruction before starting recovery.
- Resume a durable conversation whose current host record is reclaimed.
- Treat a durable dead record as reclaimed even when its last process identity remains.
- Route an unhosted runtime projection through the same durable recovery decision.
- Keep the durable delivery queue as the operation’s sole actuator after recovery.
- Hold a reservation while an accepted resume is still publishing its process.
- Add conversation_deliverability, read directly from the durable registry, with deliverable, synchronizing, reclaimed, superseded, and unknown conditions.
- Give reclaimed recovery and host synchronization distinct errors.

## Exact-once safety

Recovery never reissues a send with an unknown fate. A delivery-uncertain reservation blocks recovery across absent and dead runtime projections, a publishing resume starts no second recovery, and the queue delivers the reserved operation once. Recovery failures preserve the accepted operation ID for receipt verification.

## Verification

- bunx tsc --noEmit --incremental false
- 180 focused tests across the structured delivery and MCP files
- privacy-publication gate with commit checks
- separate full-diff self-review: no findings

Closes #1300


Latand:
Orchestrator decision, 2026-08-30: round 4 found a real race — account-migration reconciliation can terminalize a reclaimed conversation's zero-attempt reservation before recovery establishes ownership, so `send_message` reports held while nothing can ever land. One more fix round on exactly that finding is owed, with the production-tick test the reviewer specified. It is held until the delivery-drain outage in #1314 is resolved, because a new stage host currently dies before its first message arrives.

PR #1393, opened 2026-09-01
fix(seat): survive Viewer restarts with MCP re-hosting (#1346 #1354)

### LINKED ISSUE #1346: orchestrator seat loses its viewer MCP connector on viewer restart and never reconnects for the session's lifetime
Observed on the orchestrator seat during the 2026-08-31 night: a viewer restart (deploy promote) severed the seat's structured host mid-turn; the host was re-hosted automatically, but the session's viewer MCP server connection dropped and every `mcp__viewer__*` tool vanished from the session permanently — later re-hosts of the same session did not bring the connector back.

Consequence: the seat spent the rest of the night driving spawns, pipelines, retries, closes, flows and account switches over raw HTTP with a token — functionally equivalent (same endpoints, same idempotency), but it abandons the MCP conventions the seat is supposed to observe (clientRequestId discipline, typed schemas, tool-call receipts in the transcript), and the operator reasonably asks why the orchestrator is shelling curl at its own product.

Expected: when a structured host is severed and re-hosted, its MCP server connections are re-established with it — the viewer MCP is a local stdio child, so a re-host should relaunch it; alternatively the harness should retry the connector on first failed use instead of tombstoning the toolset for the session.

Repro shape: orchestrator session with the viewer MCP attached → deploy promote restarts the viewer container → host severed mid-turn and re-hosted → MCP toolset gone; tool discovery reports no matching tools for the rest of the session across further re-hosts.

Acceptance: after a mid-turn sever + re-host, the seat's next turn can call viewer MCP tools again without operator intervention; a regression test covers connector re-establishment on re-host.

### LINKED ISSUE #1354: orchestrator mandate bakes the launch-time viewer port into the prompt; the seat clings to a dead URL after every restart
Operator report (2026-08-31, phone): a freshly seated orchestrator's mandate opens with "Drive work through the production Viewer API at http://127.0.0.1:<ephemeral port>". The port is whatever the process owned at scaffold-render time; a viewer restart or redeploy changes it, so the durable seat prompt permanently names an address that stops answering — the seat "loses" the API through no fault of its own, or worse, talks to a stale instance that still squats the old port.

## Mechanism

`src/lib/roles/defaults.ts:32` interpolates `VIEWER_BASE_URL` into the orchestrator prompt scaffold at render time. The mandate is durable (it survives restarts, rotations, redeploys by design), but the interpolated URL is launch-ephemeral. Live evidence: one seat's mandate carries port 18815, another's carries 19164, while the production viewer answers on its stable port.

## Requirements

- The mandate must never embed a launch-ephemeral address. Options, in preference order: (a) instruct the seat to use the viewer MCP tools, which need no URL at all; (b) name only the stable production endpoint from configuration; (c) tell the seat to resolve the current endpoint dynamically at each use (a well-known state file or discovery route) instead of remembering one.
- Existing seated mandates with baked ports should keep working: whatever resolution the fix chooses must also apply when an old mandate's URL stops answering (the seat's tooling falls back to the resolver rather than erroring on the dead port).
- A test pins that the rendered scaffold contains no interpolated host:port literal.

Latand:
## Summary

- preserve the engine class receiver during startup adoption so Codex and Claude re-hosts complete
- replay the full Viewer stdio MCP definition and stable resolver environment into replacement harnesses
- render the orchestrator scaffold around Viewer MCP tools with no host-port literal
- route retired launch URLs through the stable runtime-host listener while candidate probes stay pinned to their exact endpoint

## Verification

Pinned head: `99edaca2fa8615a26f6191deb9f6bdfdc1c492bc`  
Pinned merge base: `d4ee411937a986d32800e53b1101a9efca5168c3`

- `bunx tsc --noEmit`: pass
- six touched test files in one explicit by-path invocation: 202 pass, 0 fail
- mutation check: the connector-definition, retired-port, and URL-free scaffold regressions each went red under the pre-fix behavior
- `bun scripts/privacy-publication-gate.ts --base d4ee411937a986d32800e53b1101a9efca5168c3`: pass
- focused self-review across correctness, test sensitivity, silent failures, types, comments, and simplification: no findings

Fixes https://github.com/Latand/live-log-viewer-next/issues/1346
Fixes https://github.com/Latand/live-log-viewer-next/issues/1354

### DIFF
--- a/src/lib/codexHeadlessConfig.test.ts
+++ b/src/lib/codexHeadlessConfig.test.ts
@@ -13,7 +13,7 @@ test("headless Codex threads allow only the registered Viewer MCP server", () =>
     },
   })).toEqual({
     mcp_servers: {
-      viewer: { enabled: true, default_tools_approval_mode: "approve" },
+      viewer: { command: "agent-log-viewer-mcp", enabled: true, default_tools_approval_mode: "approve" },
       docs: { enabled: false },
     },
     features: { plugins: false, apps: false, multi_agent: false, realtime_conversation: true },
--- a/src/lib/codexHeadlessConfig.ts
+++ b/src/lib/codexHeadlessConfig.ts
@@ -55,7 +55,10 @@ export function headlessCodexThreadConfig(
           ? configuredApproval
           : null;
       return [name, {
-        ...(name === "viewer" && viewerMissing ? server as JsonObject : {}),
+        /* A replacement app-server must receive the launch definition again.
+           Its predecessor owned the stdio child, so an enable flag alone
+           leaves a resumed thread with no connector process to call (#1346). */
+        ...(name === "viewer" ? server as JsonObject : {}),
         enabled: enabled.has(name),
         ...(approval ? { default_tools_approval_mode: approval } : {}),
       }];
--- a/src/lib/mcp/bindings.ts
+++ b/src/lib/mcp/bindings.ts
@@ -205,8 +205,9 @@ async function requestViewerControl(
   pathname: string,
   init: RequestInit,
   context: McpToolCallContext = {},
+  pinConfiguredEndpoint = false,
 ): Promise<{ response: Response; parsed: unknown; unreadable: boolean }> {
-  const baseUrl = viewerControlOrigin();
+  const baseUrl = viewerControlOrigin(process.env, pinConfiguredEndpoint);
   const now = Date.now();
   const deadlineAt = context.deadlineAt;
   const retryable = deadlineAt !== undefined;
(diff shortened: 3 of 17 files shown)


Latand:
**Merge hold — orchestrator decision.** Do not merge at `99edaca2` despite the green checks.

The round-4 review finding on `src/lib/mcp/controlEndpoint.ts` reproduces against this branch, and it is a live-state safety hole of the class AGENTS.md calls out:

`viewerControlOrigin()` treats an explicitly configured `LLV_VIEWER_CONTROL_URL` as a *hint*. When a test sets only that variable and leaves `LLV_STATE_DIR` / `XDG_CONFIG_HOME` / `LLV_VIEWER_DEPLOY_TARGET` unset, `releaseTargetPath()` falls back to `$HOME/.config/agent-log-viewer/state/viewer-release.json` — the operator's real release file. The configured fixture origin then fails the equality check against the release endpoint and resolution drops through `stableControlOrigin()`, which, with `LLV_VIEWER_PORT` unset, returns the default control origin.

Verified preconditions on a developer machine:
- the default release file exists and parses, with a valid loopback `endpoint`, `image`, `container` and 40-hex `revision`, so `currentReleaseOrigin()` returns non-null instead of the `ENOENT` short-circuit;
- `src/lib/mcp/deploymentReads.test.ts` (the two `deployment_status` adapter tests) starts a `Bun.serve` fixture on port 0 and sets **only** `process.env.LLV_VIEWER_CONTROL_URL`.

Both hold together, so those tests resolve away from their own fixture listener and onto the operator's running Viewer. The same resolver backs every viewer MCP binding, not just this read, so the blast radius is any test in that shape.

This is a regression introduced by the branch: the resolver must fail closed, so an explicitly configured endpoint is never overridden by ambient release state.

Both remaining HIGH findings (this one and the missing separate-process sever/re-host regression) are assigned to a successor lane that continues on this branch, so this PR updates in place. Converting to draft until that lands.

PR #1394, opened 2026-09-01
fix(delivery): dedupe redelivery and add safe uncertain exits (#1366 #1226)

### LINKED ISSUE #1226: composer: an exit from a delivery that never arrived, without a retry that can deliver twice (from #1213)
Cut out of PR #1217 (issue #1213) in its final pass. Expected #2 of #1213 —
"the operator gets a control at that point — retry, or discard — instead of a
spinner with no exit" — is deferred here.

What shipped instead: `delivery-uncertain` is a terminal, visible state in the
composer that says the message was not delivered, how long it waited and why,
and that sending it again is the only thing that will move it. That is honest
and it cannot double-deliver.

## Why the control was cut

Round four:

> `src/lib/runtime/http.ts:443` abandons the parked operation and mints a NEW
> durable operation id, and that id is the key both engine hosts dedupe on — so
> a retry can deliver the message twice.

And the operator's own weighing of it:

> The message the operator lost is one he can retype; a message delivered twice
> is one he cannot take back.

Round two also found that the exit and the attention entry disagreed about who
owned the message afterwards:

> Abandon settled the held-delivery reservation and `retryOperation` mints a
> replacement with no reservation of its own, so a retried message that parked
> again produced no attention entry, no decision line and no expiry.

## What was removed

- The `data-receipt-uncertain-retry` and `data-receipt-discard` buttons in
  `src/components/runtime/ReceiptChip.tsx`, and the `onDiscard` prop through
  `RuntimeComposerReceipts`.
- `runtimeRetryRequestInit` (the `abandonUnconfirmed: true` body) and the
  `abandonUnconfirmed` flag on `handleRuntimeRetry`.
- `runtimeOperationActionResult` / `applyRuntimeOperationAction` and the
  `handover` 409 refusal reading, with its `runtime.receipt.handoverBusy` copy.
- `discardRuntimeReceipt` and the `runtime.receipt.discard` label.
- The `serverBacked` guard that kept both controls off the composer's own
  `composer-unconfirmed:<key>` row, which names no journal operation.

## Depends on

#1225 — a terminalization that cannot race the hand-over or change the dedupe
identity. Without it this control cannot ship, which is why it is not merely
deferred UI work.

### LINKED ISSUE #1366: delivery-uncertain retries redeliver the same message; the recipient gets duplicates
Observed 2026-08-31 right after a viewer container restart. A structured send to the project orchestrator was admitted once (one clientMessageId, one held-delivery record) and ended in `delivery-uncertain` with `attempts: 3` — but the recipient's transcript shows the identical message landed as a real user record **three times**, once per attempt. The confirmation timed out each round while the delivery itself succeeded, so every retry redelivered.

Retries after an uncertain outcome are correct; what is missing is recipient-side dedup: a redelivery of the same clientMessageId/operation must replace or skip, never append a fresh user turn. Duplicates burn paid turns (the host answers each copy) and pollute the conversation record.

Repro shape: send to a conversation whose structured host is mid-respawn (right after a deploy or container restart), so delivery succeeds but confirmation times out; watch the held delivery loop through attempts and count the user records in the recipient transcript.

Related: the held-delivery race family (#1324, #1325, #1326) and delivery confirmation semantics in the structured transport.

https://claude.ai/code/session_01J2CE43BWhmGKXovHVppLyL

Latand:
Fixes #1366
Fixes #1226

Implements recipient-side deduplication and the delivery-uncertain operator exit as one lifecycle.

## Impact

A structured message produces one recipient user record when delivery succeeds and confirmation times out across repeated host respawns. Delivery-uncertain cards expose Retry and Discard. Retry preserves the active durable operation, client message, held reservation, and attention clock. Discard leaves a visible terminal card and fences every later retry.

## Design

- Codex persists a SHA-256 operation fingerprint on structured-user records and checks the canonical recipient transcript before every write.
- Rollouts larger than the bounded replay tail use an asynchronous, incrementally cached full-history fingerprint index. Stable file identity, size, and mtime fence each scan. Unreadable, malformed, changing, or oversized candidate evidence blocks redelivery.
- Unknown-fate Retry re-arms the same journal operation and held reservation. Repeated requests converge on that identity.
- Discard uses an in-transaction journal status fence, then terminalizes held, assigned, uncertain, or unverified reservation state. The discarded reason is absorbing in both the journal and registry, including interrupted cross-store settlement.
- A re-parked retry keeps its original admission clock and owns the message-delivery attention entry ahead of a generic stalled signal.
- Codex and Claude startup adoption retains the static host receiver so respawn recovery reaches recipient reconciliation.

## Acceptance regressions

- Delivery succeeds, confirmation times out, and repeated respawns still produce one recipient user record.
- Composer retry preserves one durable identity and one recipient write when the first delivery already landed.
- A retried message that parks again keeps its attention identity, decision line, and expiry, including a stalled live card.
- Discard remains visibly terminal, exposes no resend controls, and refuses later HTTP and low-level journal retry.
- Recipient dedup finds matching records before the 16 MiB replay tail and fails closed on unreadable or malformed evidence.

## Verification

All stateful tests used isolated HOME, XDG_CONFIG_HOME, LLV_STATE_DIR, and TMPDIR.

- bunx tsc --noEmit
- Nine changed test files by path: 357 pass, 0 fail, 1579 assertions
- Commit-aware privacy publication gate: pass
- git diff --check: pass
- Separate full-diff self-review: no findings


Latand:
**Merge hold — orchestrator decision.** The green checks here do not mean this is mergeable: the latest review verdict is REQUEST_CHANGES with an open P0, and the follow-up fix stage has not run yet.

Open findings at `cd5c8a35`:

1. **P0 — a partially settled Discard can still be retried and delivered.** `src/lib/runtime/http.ts` can persist `failed/delivery-discarded` in the journal and return 503 when registry settlement fails; the retry path consults only the registry reason, and the journal accepts every failed operation, clears the reason and requeues its effect. Reproduction: Discard → 503, immediate Retry → 202, one recipient write, journal ends `delivered` — which is the exact double-delivery this PR exists to prevent. Discard has to be absorbing in both durable stores.
2. **P1 — recipient dedup only scans the rollout tail (16 MiB).** A matching record earlier in a paginated rollout is missed, so a replacement host issues a fresh `turn/start`.

This branch is also `BEHIND` main; it needs a rebase onto the tip before any merge attempt, not a squash from behind.

Holding until the fix stage closes finding 1 and a fresh review passes at the new head.

Latand:
Merge-hold follow-up at [6a072679](https://github.com/Latand/live-log-viewer-next/commit/6a072679b30e8d08e5a1b0a593f288f67ad9c4c4):

- P0 closed: delivery-discarded is absorbing inside the runtime journal for same-identity and fresh-key retry. The HTTP path also rejects immediately when registry settlement was interrupted, before any reservation re-arm or queue wake.
- P1 closed: Codex recipient dedup indexes fingerprints across the complete rollout with asynchronous bounded reads, incremental stable-file caching, and fail-closed handling for unreadable, malformed, changing, or oversized candidate records.
- The re-parked reservation owns its attention identity and expiry even when the live card is generically stalled. The discarded card exposes no resend controls.

Verification at this head: 357 focused tests across all nine changed test files, 0 failures; TypeScript, commit-aware privacy gate, and diff check passed. A fresh full-diff self-review found no remaining issues. All required hosted checks are green, and GitHub reports the ready PR mergeable and CLEAN.

Latand:
Re-drafting: the round-7 push moved this to `6a072679`, but the review at that head still returns a P1 and the lane's fail-edge budget is now exhausted after 6 rounds.

Surviving finding: `src/lib/runtime/http.ts:480` and `:506` — Discard relies on a stale operation-status read with no journal-side fence for uncertain or terminal-unverified operations, so a concurrent same-identity Retry can re-arm and enter `host.send` before Discard writes the registry tombstone; Discard then reports success while the retried operation delivers.

Six rounds have each surfaced a *different* race in this same Discard/Retry seam, which reads as one missing fence rather than a series of separate bugs. A successor lane will take it with the invariant stated up front — Discard and Retry on one operation identity are mutually exclusive, enforced by a single durable fence taken before either actuates, rather than a status read followed by a write — plus a regression that actually induces the interleaving.

Issue #314, opened 2026-07-16
P0: block account deletion on filesystem history inventory


Latand:
## Problem

Managed account deletion recursively removes the account home. The home contains Claude or Codex transcript files.

The current safety blocker reads AgentRegistry conversation ownership and checks only each conversation's latest generation. It does not inventory the filesystem before home removal. The source comment also records a cross-store race between the blocker check and a new spawn.

Read-only production audit on 2026-07-16:

- Claude botfatherdev-2: 42 JSONL files, 56,844,840 bytes; 1 file / 30,033 bytes lacks matching registry path coverage.
- Codex botfatherdev: 1,324 JSONL files, 1,592,425,202 bytes; 285 files / 284,793,451 bytes lack matching registry path coverage.
- Both current accounts still have current_conversations blockers, so their present DELETE requests are rejected.
- A managed account containing only unindexed, historical, superseded, or drifted transcript files could pass the registry-only blocker and lose those files when its home is removed.

Legacy Main is outside the managed DELETE route.

## Severity

P0 data-loss risk. Keep managed account deletion release-blocked until filesystem history coverage is authoritative.

## Acceptance

- Add a read-only deletion preflight that inventories every transcript and related user-history artifact under the exact safe managed home.
- Any history artifact blocks direct deletion, independent of AgentRegistry coverage, generation order, scan cache, or migration state.
- History blockers have no force bypass.
- Provide an explicit archive/export or migration workflow before deletion.
- The archive lives outside the account home, has a manifest, file sizes, checksums, and verified readability.
- Require fresh user confirmation after the archive or migration succeeds.
- Fence spawn, resume, login, quota probes, migration, and transcript writes across preflight and removal.
- Detect filesystem changes between preflight and commit and abort safely.
- Reject symlinked homes, symlinked history files, path traversal, external mounts, and unsafe ownership or permissions.
- UI shows transcript count, total bytes, blocker reason, archive destination, verification result, and zero ambiguous force actions.
- cleanupOrphans applies the same history inventory and backup rules.
- Preserve Legacy Main, registered homes, current selection, conversations, tasks, assignments, and unrelated sidecars.
- Production-shaped tests cover unindexed transcripts, superseded generations, subagent trees, concurrent transcript creation, registry drift, archive verification failure, cleanup retry, and crash recovery.

Related: #79, #313.

Issue #313, opened 2026-07-16
Account deletion should clean provider sidecar directories safely


Latand:
## Problem

Managed account deletion removes the registered home recursively, and the API exposes cleanupPending when that filesystem step fails. Current production state still contains Claude account artifacts outside registered homes:

- registered managed Claude homes: botfatherdev-2, botfatherdev-3
- safe orphan home candidates: botfatherdev, botfatherdev-1
- adjacent sidecar directories refreshed by provider activity: botfatherdev-2.lock, botfatherdev-3.lock
- the two orphan homes are 16 KiB each and contain zero JSONL transcripts

The sidecar names fail the managed-account ID safety regex, so cleanupOrphanedClaudeHomes classifies them as unresolved. A deleted account can therefore leave or recreate an adjacent provider lock directory after its registered home disappears.

Issue #79 and PR #94 delivered managed account removal and orphan-home cleanup. This follow-up covers provider sidecars and concurrent quota/login activity.

## Safety dependency

This work is release-blocked by #314. Account home or orphan cleanup first requires filesystem history inventory, verified archive or migration, and fresh user confirmation. Sidecar handling cannot authorize home deletion or bypass a history blocker.

## Acceptance

- Deleting an eligible managed Claude account removes its provider-owned sidecar artifacts after live session, login, and quota subprocess ownership is fenced.
- Orphan cleanup recognizes only exact safe sidecar names derived from an unregistered managed account ID.
- Locks belonging to registered, active, logging-in, or quota-probed accounts stay untouched.
- A concurrent provider process cannot recreate an orphan sidecar after successful deletion.
- cleanupPending and the Accounts UI report home and sidecar outcomes precisely.
- Failed-login orphan homes remain removable through the existing safety checks and #314 archive requirements.
- Production-shaped tests cover deletion success, concurrent recreation, stale sidecar cleanup, symlink rejection, registered-account preservation, and retry behavior.
- Existing transcripts, Legacy Main account data, registered account homes, and account selection remain preserved.

Latand:
Safety dependency added after the filesystem-history audit:

This sidecar cleanup work is release-blocked by #314. Account home or orphan cleanup must first inventory transcript/history artifacts, produce and verify the required archive or migration, and receive fresh user confirmation. Sidecar handling cannot authorize home deletion or bypass any history blocker.

PR #1401, opened 2026-09-01
fix(accounts): block deletion on filesystem history inventory

Latand:
## Summary

- inventory Claude and Codex history from isolated managed homes and reject unowned, unreadable, unsafe, mounted, or changed artifacts with exact reports
- stage home cleanup reversibly with fd-anchored identity checks, preserve registry-owned history in retained archives, and remove exact provider sidecars safely
- fence login, OAuth, quota, app-server, spawn, resume, and orphan-cleanup ownership across deletion; remove every force bypass
- keep Linux cleanup anchored to opened directories and fail closed without fd anchoring on other platforms

## Verification

- bunx tsc --noEmit
- 224 tests passed across all ten changed test files, each run by explicit path with isolated state and provider homes
- 32 adjacent account-lock, interprocess, liveness, and rate-limit tests passed across four explicit test files
- privacy publication gate passed locally and on GitHub
- all hosted checks passed, including the successful Windows platform rerun
- fresh read-only self-review: no findings

## Scope

References #314 and #313. This implements the pinned filesystem-inventory, reversible-cleanup, provider-sidecar, and lifecycle-fence slice. The broader external archive manifest, checksum, fresh-confirmation, and full archive UI workflow remains tracked by #314.


Latand:
**Merge hold — orchestrator decision.** Three review rounds have now each found a different hole in the same closed history-root allowlist, and the lane's fail-edge budget is exhausted.

Surviving P0, from the review at this head: `src/lib/accounts/removal.ts:53` — the closed history-root allowlist omits documented Codex log and SQLite session data, so an inventory can come back empty and a recursive deletion then succeeds.

The pattern matters more than the individual gaps: an allowlist of known history roots **fails open by construction**, because anything it has not heard of is silently treated as safe to delete. On the one non-recoverable failure class in this codebase, that default is backwards.

A successor lane continues on this branch with the default inverted — every artifact under the account home classifies as owned, known-history, or unknown; deletion proceeds only when both of the latter are empty; an inventory that cannot complete is itself a blocker; and every refusal names the paths that caused it. Drafted shut until that lands and a fresh review passes at the new head.

Latand:
**Merging with the residual tracked as [#1408](https://github.com/Latand/live-log-viewer-next/issues/1408).**

I verified this head myself rather than relying on a verdict. `src/lib/accounts/removal.ts` now classifies every artifact under the account home as `owned` / `history` / `unknown`, blocks unless everything is provably owned, blocks on an incomplete traversal, revalidates account-home directory identity during removal, and names the paths that caused each refusal. That is the requirement in #314 and #313.

The surviving reviewer finding is a real TOCTOU — entries are classified from `lstat` and later unlinked by pathname, so something swapped into that path during the window inherits the earlier verdict. I am not holding this PR for it, for three reasons:

1. Pre-existing user history is unaffected: it is classified before the window and blocks deletion. The probe's loss was an artifact introduced *during* removal, and the external symlink target survived.
2. Before this branch, deletion consulted only registry ownership with no filesystem inventory at all. Merging is a strict improvement on exactly the failure class this work exists to prevent; holding it keeps the weaker behaviour in main.
3. Two lanes and five review rounds each found a different hole here, and the proposed remedy (anchored staging-area relocation with device/inode/type verification per component) is heavier than the mechanism around it. That trade-off deserves its own scoped change, not a sixth round on a shipping PR.

#1408 carries the mechanism, the blast-radius analysis, the acceptance regression, and a cheaper option to weigh first (re-`lstat` immediately before unlink, comparing device/inode/type against the classified record).

Latand:
**Correcting my previous comment — this is NOT being merged.**

I said I would merge with the TOCTOU tracked as #1408. That was wrong, and I found it by verifying the head instead of trusting the round summaries.

An earlier P0 is still live at this head. `src/lib/accounts/removal.ts:66`:

```
claude: new Set(["cache", "debug", "backups", "paste-cache", "plugins"]),
```

`debug` is still blanket-owned, so every regular descendant classifies as deletion-owned and a per-session `debug/<session-id>.txt` log is deleted. Review round 2 raised exactly this and proved it with an isolated probe. Unlike #1408 this needs no race at all — it is a static classification hole that loses user-history-shaped data on an ordinary deletion.

How it survived: the branch carries **one** commit beyond `c5f8e3f3` — `3829b34f`. Builder rounds 2 and 3 both reported "P0 closed" at that same SHA, having pushed nothing. Review round 3 then reviewed the unchanged source and reported a different finding (the TOCTOU), so the `debug` P0 was never re-checked and silently dropped out of the record.

The trade-off I argued last comment still stands for #1408 — a local race, blast radius limited to an artifact created during the deletion window. It does not extend to this, which is deterministic data loss. A successor lane will close it before this PR merges.

Issue #1367, opened 2026-08-31
feed: floating suggested-reply pills and the scroll-magnet glue fight upward scroll — operator sometimes cannot scroll up


Latand:
Operator report (2026-08-31, recurring): while reading a live conversation, scrolling UP sometimes does nothing — the feed either refuses to move or snaps back down. The feeling is that it happens when the suggested-reply pills (the drafts a manager seat offers via `suggest_replies`) are pinned above the composer; sometimes an upward scroll "breaks through", sometimes it does not. A scroll-snapping bug in this area was reported and closed before (#983, mobile composer focus; #1030 tracked a refusal-alert scroll leftover), and the operator says the problem still reproduces after those fixes.

## Two mechanisms to establish (both at current main)

1. **The floating pill overlay eats gestures that start on it.** `src/components/LogFeed.tsx` renders `SuggestedReplies` with `floating` inside a wrapper centered above the composer (`pillPos` at ~line 678, wrapper ~709–722), overlaying the scroller. The pill row itself is a scrollable region: `overflow-x-auto` with `flex-wrap` (`src/components/feed/SuggestedReplies.tsx` ~245–249). A wheel or touch gesture that starts over the pills targets the pill row's own overflow, which has no vertical range; without `overscroll-behavior` / scroll-chaining discipline (or pointer-events pass-through for the wrapper's dead zones), the vertical delta can be consumed and never reach the feed underneath. That matches "when the buttons hang there, I can't scroll".

2. **The scroll-magnet glue window misclassifies real user scrolls.** `LogFeed.tsx` holds a bottom "magnet" with programmatic glue: after a glue, any not-at-bottom scroll event within `GLUE_SETTLE_MS = 300` (~line 72; glue logic ~273–310) is treated as "layout still settling" and glued back to the bottom. On a busy live conversation the transcript appends many times a second, so glues can be near-continuous — a user's upward wheel/drag that lands inside the 300 ms window is undone, repeatedly. The user escapes only when their scroll happens to land outside the window. That matches "sometimes I break through, sometimes not", and it reproduces with or without pills — the pills just correlate (they float exactly when the feed is busy at the bottom).

## Wanted

- A gesture that starts over the floating pills scrolls the feed unless the pill row itself actually has horizontal range in that axis; the wrapper never swallows vertical scroll.
[...]

PR #1407, opened 2026-09-01
fix(feed): keep live-tail follow disarmed after upward intent (#1396)

### LINKED ISSUE #1396: fix(feed): live-tail follow re-pins during an upward scroll even with no pills on screen — the shipped #1367 fix does not cover the magnet itself
Operator verification on the shipped build (2026-09-01 morning, after the night deploy that carried #1367 + its successor PR #1379): the upward-scroll fix does not cover their case. With NO floating reply pills on screen, scrolling up in a live conversation gets re-pinned to the bottom by the live-tail follow, repeatedly — "I just scroll up and it keeps snapping me back down."

So the #1367 diagnosis (pills swallowing the gesture; programmatic glue erasing a pending intent tag) was real but partial: the live-tail magnet itself re-arms against an active upward scroll, most likely on each content arrival, and in a busy conversation arrivals come faster than the operator can escape.

## Wanted

1. An intentional upward scroll DISARMS follow, full stop — content arrival, resize, image load, or any programmatic scroll must not re-arm it. Follow re-arms only when the operator returns to the bottom (or taps an explicit jump-to-latest control).
2. The disarm must win under rapid content arrival: the operator's suggestion is a grace window/debounce, but the acceptance bar is stronger — during a continuous upward gesture in a conversation receiving a burst of records, the viewport must never move down.
3. Regressions: burst of tail records while an upward wheel/touch gesture is in progress → no downward snap; follow stays off across N further arrivals until bottom is reached; the jump-to-latest control still works.

Related: #1367, PR #1379 (shipped, insufficient), the feed's stick-to-bottom glue and scroll-intent tag machinery.

Latand:
Closes #1396

## Impact

An upward wheel, touch drag, or scroll key now disarms live-tail follow before fresh records can move the viewport. Follow stays released through tail arrivals, content resizes, and programmatic glue until the operator reaches the bottom or activates the jump-to-latest control.

## Cause

The shipped gesture tag released the magnet on the resulting scroll event. A live tail render could arrive between the input event and that scroll event, observe stale magnet state, and glue the viewport to each new bottom.

## Fix

- Close the synchronous per-transcript follow latch as soon as upward intent is known.
- Make every glue consult that latch so stale content renders and resize callbacks cannot move the viewport.
- Re-arm from verified user movement back to the bottom or an explicit follow control.
- Track touch direction so upward touch intent receives the same early latch as wheel and keyboard input.
- Add wheel-burst, touch-burst, durable-arrival/bottom-return, and jump-control DOM regressions while retaining the shipped #1367 pill and resize-glue cases.

## Verification

- bunx tsc --noEmit
- bun test src/components/LogFeed.suggestedReplies.dom.test.tsx src/components/feed/SuggestedReplies.dom.test.tsx with isolated HOME, XDG_CONFIG_HOME, LLV_STATE_DIR, and TMPDIR: 19 passed, 54 assertions
- bun scripts/privacy-publication-gate.ts --base 3582778e4e068ce011802c1e6af9adc29a2d1b62
- git diff --check

The wheel-burst regression was red before the fix: the viewport advanced from 800 to 900, 1000, and 1100 across three records.

### DIFF
--- a/src/components/LogFeed.suggestedReplies.dom.test.tsx
+++ b/src/components/LogFeed.suggestedReplies.dom.test.tsx
@@ -1,5 +1,6 @@
 import { afterAll, afterEach, beforeEach, expect, mock, test } from "bun:test";
 import { Window as HappyWindow } from "happy-dom";
+import type { ReactElement } from "react";
 import { flushSync } from "react-dom";
 import { createRoot, type Root } from "react-dom/client";
 
@@ -144,6 +145,7 @@ afterEach(() => {
   for (const root of roots) flushSync(() => root.unmount());
   roots.clear();
   dom.document.body.replaceChildren();
+  tailState.lines = TRANSCRIPT;
 });
 afterAll(() => {
   globalThis.fetch = previousFetch;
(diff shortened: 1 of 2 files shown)


Latand:
**Merge hold — orchestrator verification.** Not merging at `9282c4d7` despite the lane's APPROVE, because the approval does not describe this head.

Round 1 of review raised a P1: native scrollbar return cannot re-arm follow — `userReturnedToBottom` requires a wheel/touch/key-generated cause tag, so dragging the scrollbar thumb or clicking its track reaches the bottom through an untagged `scroll` event and leaves follow disarmed.

The builder's next round reported that finding fixed, but reported **the same commit** `9282c4d7`. The branch carries exactly one commit:

```
9282c4d7 fix(feed): latch upward scroll intent
```

and at that head:

- `LogFeed.tsx` has no `pointerdown` / `mousedown` / scrollbar-cause handling at all;
- `userReturnedToBottom = atBottom && userInitiated && userDelta < 0`, where `userInitiated` still requires `cause?.kind === "user"`, and that tag is only ever set from the wheel/touch/key handlers.

So the round-1 P1 is unchanged in the code, and the round-2 APPROVE reviewed byte-identical source to the round-1 REQUEST_CHANGES. The verdict is not reliable evidence here.

What this means in practice: the primary defect from #1396 — an upward scroll getting snapped back to the bottom — **is** fixed, and touch return works, so the mobile case the issue was filed for is covered. The residual gap is desktop: after scrolling up, returning to the bottom by dragging the scrollbar will not re-arm live-tail; the jump-to-latest control still will.

Holding the merge and sending a successor to close the scrollbar re-arm path on this branch, so the PR updates in place.

PR #1414, opened 2026-09-01
fix(deploy): clear completed runtime-host handoff intent

### LINKED ISSUE #1412: bug(deploy): the runtime-host handoff intent is never cleared, so every promote wedges the next one and the host lags a revision behind
## Summary

Every promote publishes the Viewer correctly, then wedges at the `host-handoff` phase. The deployment record never becomes terminal, stays `active`, and therefore blocks the deploy lane: the next request is refused with `busy`. The runtime host ends up one revision behind the Viewer, permanently, until someone intervenes by hand.

Observed twice in a row today with the same signature:

```
phase: host-handoff   terminal: false   active: 1
error: runtime-host handoff intent is owned by another generation
```

## Mechanism

`stageRuntimeHostSuccessorContainer` reads the durable handoff intent and refuses when it exists but does not match the candidate:

```
const exactIntent = intent
  && intent.revision === candidate.revision
  && intent.image === candidate.image
  && intent.successorContainer === name;
if (intent && !exactIntent) {
  throw new Error("runtime-host handoff intent is owned by another generation");
}
```

That guard is correct. The defect is upstream of it: the intent from a **completed** handoff is not cleared. The function's own contract says it should be — "after the predecessor exits and the successor acquires the fence, the successor removes the recorded predecessor and clears the intent" — but the file survives.

So each deploy leaves a live mine for the next one:

1. Deploy A stages a host successor, records the intent, and the successor container does come up and serve.
2. The intent file is never removed.
3. Deploy B stages its own successor, finds A's intent, sees it does not match its candidate, and throws.
4. B wedges at `host-handoff`, non-terminal and active, so the lane is blocked for everything after it.

Verified against the leftover file: its `revision`, `image` and `successorContainer` named exactly the runtime host container that was running and healthy at the time, i.e. it described a handoff that had already finished.

## Second defect: the wedged record cannot be driven forward

Once wedged, nothing advances the record:

- re-requesting the same revision through the deployments endpoint (the path `rebuild.sh` uses) answers `202` with `replayed: true` and does not re-execute;
- re-requesting through the MCP deploy tool with the same idempotency key does the same;
- a different revision is refused with `busy`, returning the wedged deployment's id;
- the record's owner is the live runtime host process, so it is not orphaned and no reaper claims it.

Recovery required removing the stale intent file and terminalising the record by hand. That is not an acceptable steady state for a deploy path.

## Wanted

1. Clear the handoff intent once the handoff completes, which is what the documented sequence already promises. That alone removes the recurrence.
2. Make a stale intent recoverable without hand edits: if the recorded intent describes a handoff whose successor container is already running and healthy at that revision, treat it as complete and clear it rather than refusing forever.
3. Give a wedged deployment a way forward: either a retry that actually re-executes from the durable phase, or a bounded watchdog that terminalises it with its reason. A non-terminal record that blocks every future deploy and cannot be advanced is worse than a failed one.
4. Fail the promote loudly when the host handoff does not complete. Publishing the Viewer while silently leaving the host on the previous revision hides a real split — the two only appear consistent because the Viewer serves fine.

## Acceptance

- Two consecutive promotes complete with both Viewer and runtime host on the new revision, and no intent file left behind.
- A promote whose handoff fails settles terminal with its reason and does not block the next deploy.
- A regression covers the shape: an intent left by a completed handoff does not fail the next deployment.

Latand:
## Summary

- Clear the durable handoff intent after verified successor startup.
- Recover a stale foreign intent when Docker health proves its recorded generation is serving.
- Preserve exact-match crash recovery without fence-owner predecessor discovery.
- Settle handoff failures terminal with their reason so later deployments can enter the lane.

## Recovery choice

The coordinator terminalizes a caught handoff failure immediately. The failed operation already supplies a bounded result and a durable reason. Keeping that deployment active blocks later deploys without advancing the handoff.

## Verification

- `bunx tsc --noEmit`
- `bun test src/runtime-host/hostSuccessor.test.ts src/runtime-host/deploymentAdapter.test.ts src/runtime-host/deployment.test.ts` (78 pass)
- `bun scripts/privacy-publication-gate.ts --base 4ec4e71b07a97003ca5ae300d27d15343723ecac`

Pushed SHA: `d658555bb747e58abd754ef4cb235e5923d24647`

Fixes #1412

### DIFF
--- a/src/runtime-host/deployment.test.ts
+++ b/src/runtime-host/deployment.test.ts
@@ -532,7 +532,7 @@ test("issue 1268: the predecessor leaves success non-terminal until the staged r
   store.close();
 });
 
-test("issue 1268: terminal success requires runtime-host hand-off evidence even when generation tracking is unavailable", async () => {
+test("issue 1412: a failed runtime-host handoff settles terminal and clears deployment admission", async () => {
   const store = journal("host-successor-proof-required");
   const adapter = new FakeDeploymentAdapter();
   adapter.verifyHostFailure = new Error("runtime-host startup evidence is unavailable");
@@ -546,13 +546,20 @@ test("issue 1268: terminal success requires runtime-host hand-off evidence even 
   await coordinator.waitForDeployment(receipt.deploymentId);
 
   expect(store.viewerDeployment(receipt.deploymentId)).toMatchObject({
-    phase: "host-handoff",
-    terminal: false,
+    phase: "failed",
+    terminal: true,
     error: "runtime-host startup evidence is unavailable",
   });
+  expect(store.activeViewerDeployment()).toBeNull();
   expect(adapter.calls).toContain(
     `verify-host-successor:${store.viewerDeployment(receipt.deploymentId)?.candidate?.image}:${"b".repeat(40)}`,
   );
+  const next = await coordinator.requestViewerDeployment({
+    idempotencyKey: "deploy-after-host-successor-proof-failure",
+    revision: "c".repeat(40),
+  });
+  expect(next).toMatchObject({ state: "accepted", replayed: false, revision: "c".repeat(40) });
+  if (next.state === "accepted") await coordinator.waitForDeployment(next.deploymentId);
   store.close();
 });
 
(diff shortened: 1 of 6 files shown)


Latand:
Closing as superseded: the intent-clearing half landed in #1413 and the terminal-settle / verified-handoff half landed in #1416, both on main. This branch conflicts with that merged work and carries nothing further.

Issue #1497, opened 2026-09-03
Viewer-native sign-in and per-message authorship: everyone sees who sent what


Latand:
## What the operator asked for

Verbatim, 2026-09-03, after four teammates were given access to one shared Viewer: he wants authentication **inside the Viewer** — "чтобы вот-вот было видно, кто отправляет сообщение, ну то есть люди залогинились, кто залогинился, тот отправляет сообщение. И видно, что там подписывается кто отправил сообщение… чтобы другие люди видели, кто что пишет".

Three things, in his order: people sign in to the Viewer itself; the person who is signed in is the one whose messages are sent; every message carries a visible signature of who sent it, readable by everyone else.

## Why it does not work today

Teammates now reach a shared Viewer through a reverse proxy that asks for a password. That password is checked **in front of** the Viewer and never reaches it. Inside, `operatorAuthority` answers exactly one question — *are you a human browser or an agent?* — and by design it does not ask which human:

> That question is answered below, and it is the ONLY question asked.

So all four teammates are the same principal. A message any of them sends is written with no author, indistinguishable from the operator's own. Nobody can tell who asked an agent for what, and the transcript has no field that would let them.

The external password buys **access control** and nothing else. Attribution needs identity the Viewer itself holds.

## The trap this must not fall into

`src/components/operatorCredential.ts` is the tombstone of the previous attempt, and it is required reading before designing this:

> Rounds 7–9 of #691 built one — a key printed to the launching terminal, pasted into a gate, held in a single volatile module slot, lost on reload — and the operator rejected it on stage for the reason the design never weighed: it broke the two gestures the product is FOR. A fresh tab could not open the manager and could not start a call from a card, and every reload put it back in that state.

That rejection stands. Any design here is measured first against those two gestures: **a fresh tab opens the manager, and a card starts a call** — with no paste, no ceremony, no state that dies on reload. What changed since is not the UX bar but the premise: there is now genuinely more than one human, so "same-origin therefore the operator" has stopped being true. The answer is a session the browser keeps, not a secret the human carries.

## Scope

The first shippable slice of #634, not a competing epic. #634 already fixes the model — `Member` with a stable id, sessions, owner plus additive capabilities. This issue takes the smallest end-to-end cut of it:

1. **Sign-in.** A person identifies themselves to the Viewer once per device and stays signed in across reloads and new tabs. Who may sign in is a list the owner controls. No printed key, no paste.
[...]

Issue #1495, opened 2026-09-03
`--hostname` is silently ignored on the `next start` path, so a loopback-only request binds every interface


Latand:
Found while bringing the Viewer up on a shared server. Asking for `--hostname 127.0.0.1` — which is the **default** — produced a listener on `*:8898`.

## Two launch paths, one of them honours the flag

`resolveServer` in `bin/cli.mjs` picks between a standalone build and a plain checkout:

- **standalone** (`dist/standalone/server.js` or `.next/standalone/server.js`) — Next's standalone server reads `HOSTNAME` from the environment, so `buildChildEnv`'s `HOSTNAME: options.hostname` binds where the caller asked. This path works, and the comment there says as much: *"setting it here keeps standalone bound to the requested address"*.
- **`next start`** (`commandFor(nextBin, ["start"])`) — `next start` takes the bind address as the `--hostname` **argument**; it does not read the environment variable. Nothing passes it, so the address the operator chose is dropped on the floor and the server binds all interfaces.

A checkout that has not been built into standalone — which is exactly how someone runs the Viewer from a git clone on a server — takes the second path.

## Why this is worse than a stray listener

The CLI already knows that binding beyond loopback is a decision that needs care: `isLoopbackHostname` and `nonLoopbackBind` exist precisely to treat a non-loopback bind differently. That guard reads `options.hostname` — the *requested* address — not what the process actually bound. So on this path:

- the operator asks for loopback, so the guard sees loopback and stays quiet;
- the server binds `0.0.0.0` anyway;
- the one code path meant to notice this is the one thing guaranteed not to.

The failure is silent in both directions: nothing warns, and the log line reports the address that was asked for rather than the one in use.

On the machine where this surfaced a default-deny firewall happened to be in front of it, and the access-token check answers `403` to any non-loopback `Host` header, so nothing was exposed. Neither of those is the Viewer's doing, and neither is guaranteed on the next machine.

## Fix

1. Pass the address on the `next start` path — `commandFor(nextBin, ["start", "--hostname", options.hostname])` — so both paths honour the flag. Keep the `HOSTNAME` env var for standalone.
2. Derive the non-loopback guard from what was actually bound rather than from what was requested, so a future divergence cannot slip past it again.
3. A test that starts the CLI on the `next start` path with the default hostname and asserts the listener is not reachable on a non-loopback address. Without it this regresses the moment Next changes its flag handling again.

Latand:
[...]

PR #1503, opened 2026-09-04
fix: the Viewer decides who may reach it from the connection, not from a header (#1496, #1495)

### LINKED ISSUE #1496: The auth gate decides "local" from the Host header, so any local account on the machine reaches the Viewer unauthenticated
Found while putting the Viewer on a shared server so a team can use it. On that box a second, unprivileged account ran `curl 127.0.0.1:8898` and got **200** — the full Viewer, no token, no cookie.

## The mechanism

`isRemote` in `src/proxy.ts` decides whether a request needs the access token, and it decides from two headers:

```ts
function isRemote(request: NextRequest): boolean {
  const forwardedFor = request.headers.get("x-forwarded-for");
  if (forwardedFor !== null && hasRemoteForwardedAddress(forwardedFor)) return true;

  const host = request.headers.get("host");
  return host === null || !LOOPBACK_HOSTS.has(hostWithoutPort(host));
}
```

and the caller skips the whole gate when it says false:

```ts
if (!isRemote(request)) { /* no token required */ }
```

`Host` is chosen by whoever sends the request. It says which name the client dialled, never where the connection came from. So "the request carries `Host: localhost`" is treated as "this request is trusted", and that is true for **every** process on the machine regardless of which user owns it.

The file already knows how to do this properly — `isLoopbackAddress` handles `::1`, `::ffff:127.0.0.1` and the whole `127.0.0.0/8` range correctly. It is simply not applied to the connection's own peer address; it is only used to sanity-check `X-Forwarded-For`.

## Why it matters more than it looks

On a single-user laptop this is invisible: the only local user is the operator. The moment the Viewer runs anywhere with a second account — which is exactly what "give the team access" means — the token gate protects nothing locally:

- every transcript of every project is readable, including whatever the operator has said to their agents;
- `/api/spawn` and the rest of the control surface answer, so anyone with a shell can start agents **on the operator's own Claude and Codex subscriptions**;
- nothing in the logs distinguishes that from the operator's own traffic.

It also means the loopback trust cannot be reasoned about from the outside: a reverse proxy that forwards the client's `Host` unchanged, or any hop that sets `Host: localhost`, extends the same trust to the internet. The current deployment happens to be shielded by a default-deny firewall and by the fact that its public entry rewrites Host — neither of which is the Viewer's own doing, and neither is guaranteed on the next machine.

## Fix

1. **Decide "local" from the connection, not from a header.** Use the peer address (`request.socket.remoteAddress` at the server boundary, or a Next equivalent) with the existing `isLoopbackAddress`. A header can support that decision; it must never be the decision.
2. **Loopback should not mean unauthenticated on a multi-user host.** Even a genuinely local connection is only trustworthy when the machine has one user. Consider requiring the token always and treating loopback as a convenience only when the Viewer is told it owns the machine (an explicit setting), rather than inferring it.
3. A test that sends `Host: localhost` from a non-loopback peer and asserts the token is still required — the current behaviour would pass any test written against the header alone, which is presumably how it survived this long.

## Workaround until it lands

On a shared host, keep other accounts off the port at the kernel level rather than relying on the app — an `iptables` owner-match rule on the loopback OUTPUT chain, allowing only the Viewer's own uid to reach its port. That is a root change and it is per-host, so it is a mitigation, not a fix.

Latand:
Closes #1496 and #1495. Both were found this week while putting the Viewer on a shared server where four teammates now have accounts, so neither was hypothetical.

**#1496 — "local" was decided from a header the caller controls.** `isRemote` read the `Host` header and skipped the token gate whenever it looked like loopback. `Host` says which name the client dialled; it is never evidence of where a connection came from. Measured on the shared box before the fix: a second, unprivileged account ran `curl 127.0.0.1:8898` and got **200** — every transcript, plus a control surface able to spawn agents on the operator's own subscriptions. A configured `LLV_TOKEN` now authenticates every connection, loopback included, and neither `Host` nor `X-Forwarded-For` offers a bypass.

**#1495 — the bind address was silently dropped on one of the two launch paths.** The standalone path honoured `--hostname` through the environment; `next start` takes it as an argument and got nothing, so a plain checkout started with the default `127.0.0.1` bound every interface. The sharper half was second-order: the non-loopback guard read the *requested* address rather than the bound one, so the one code path meant to notice this was guaranteed not to. The CLI now forwards `--hostname` and verifies the kernel-visible bind.

Three review rounds, each finding something real and each fixed:

1. unscoped link-local IPv6 (`fe80::…`) passed to `listen()` returns `EINVAL` on Linux, killing a correctly bound Viewer — interface scope is now preserved;
2. the bind guard attributed any `EADDRINUSE` on the port to the Viewer child, so an unrelated service holding that port on another address killed a correctly loopback-bound Viewer — occupancy is now captured before launch and only newly occupied addresses are rejected;
3. the final round confirmed both directions still hold.

Every regression test was verified RED before GREEN by reverting the code it guards, and the reviewer re-ran those reverts independently rather than trusting the report — which mattered, because a test written against the header alone passes on the unfixed code, and that is exactly how #1496 survived this long.

Gates at `110bb337`: `bunx tsc --noEmit --incremental false` exit 0; `src/proxy.test.ts` 3 pass, `bin/cli.exposure.integration.test.ts` 3 pass; privacy gate PASS. Reviewed fresh-context on Sol: **APPROVE**.

This also unblocks #1497 — identity in front of a bypass would have been theatre.

### DIFF
--- a/.github/workflows/platform-tests.yml
+++ b/.github/workflows/platform-tests.yml
@@ -78,6 +78,7 @@ jobs:
           src/runtime-host/socket.test.ts
           src/runtime-host/runtimeHostFence.test.ts
           bin/server-runtime.test.ts
+          bin/bind-guard.test.ts
 
   windows-platform:
     name: Windows (native)
@@ -146,3 +147,4 @@ jobs:
           src/runtime-host/socket.test.ts
           src/runtime-host/runtimeHostFence.test.ts
           bin/server-runtime.test.ts
+          bin/bind-guard.test.ts
--- a/bin/bind-guard.test.ts
+++ b/bin/bind-guard.test.ts
@@ -1,0 +1,262 @@
+import net from "node:net";
+import { expect, test } from "bun:test";
+
+import {
+  classifyBindProbeFailure,
+  linkLocalScopeZone,
+  newlyBoundNonLoopbackAddress,
+  nonLoopbackProbeAddresses,
+  readNonLoopbackBindState,
+} from "./server-runtime.mjs";
+
+/**
+ * The loopback bind guard asks one question per non-loopback address: is it
+ * bound after the Viewer launched, with nothing in the reading taken before the
+ * launch to excuse that? Everything here drives that question through its two
+ * platform-dependent halves — how an address is named, and what a failed listen
+ * means — because the runner that broke it (Windows) cannot be run from this
+ * checkout.
+ *
+ * Addresses below are invented. `fe80::1` and `fe80::2` are link-local by
+ * construction and belong to no machine.
+ */
+
+function interfaceEntry(overrides: Record<string, unknown> = {}) {
+  return {
+    address: "fe80::1",
+    family: "IPv6",
+    internal: false,
+    scopeid: 12,
+    ...overrides,
+  };
+}
+
+async function availablePort(): Promise<number> {
+  const server = net.createServer();
+  await new Promise<void>((resolve, reject) => {
+    server.once("error", reject);
+    server.listen(0, "127.0.0.1", resolve);
+  });
+  const address = server.address();
+  if (address === null || typeof address === "string") throw new Error("could not reserve a TCP port");
+  await new Promise<void>((resolve, reject) => server.close((error) => error ? reject(error) : resolve()));
+  return address.port;
+}
+
+test("a Windows link-local address is scoped by interface index, never by adapter name", () => {
+  const addresses = nonLoopbackProbeAddresses({
+    platform: "win32",
+    interfaces: { "Ethernet 3": [interfaceEntry()] },
+  });
+
+  expect(addresses).toEqual(["fe80::1%12"]);
+});
+
+test("a POSIX link-local address keeps the interface name as its zone", () => {
+  const addresses = nonLoopbackProbeAddresses({
+    platform: "linux",
+    interfaces: { eth0: [interfaceEntry()] },
+  });
+
+  expect(addresses).toEqual(["fe80::1%eth0"]);
+});
+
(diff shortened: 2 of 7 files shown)


Latand:
Blocked on one red gate, not on the review.

**Platform tests → Windows (native)** is red on this PR: 143 pass, 7 skip, **2 fail** ([run 33825277830](https://github.com/Latand/live-log-viewer-next/actions/runs/33825277830)).

```
(fail) the CLI rejects a ready runtime socket owned by another process
(fail) the CLI names a missing Bun prerequisite before starting the Viewer
```

Both are in `bin/server-runtime.test.ts`, both spawn `bin/cli.mjs` on a loopback default and assert its stderr. The same workflow is green on `main` for the last six runs (latest `375981cc`), so the regression is this PR's.

Working hypothesis, to be proven or killed by the lane rather than assumed: the new pre-launch `boundNonLoopbackAddresses()` probe runs before the runtime-host and Bun prerequisite checks, and every probe error becomes a fatal `bindCheckFail` — so on Windows the CLI exits with the wrong message and both `toContain` assertions miss. The two candidate Windows causes are the `%<interfaceName>` IPv6 zone added in `cb56492a` (Windows wants a numeric zone index) and `EACCES` rather than `EADDRINUSE` for an exclusive-use conflict.

A successor lane is running from this exact head (`110bb337`) to fix it without weakening the guard: a probe that cannot evaluate one address must neither abort startup nor count as evidence of exposure, while a newly occupied non-loopback address still stops startup. This PR gets fast-forwarded onto that work once it passes review, and merges when Windows is green.

Issue #1569, opened 2026-09-08
fix(monitor): fence automatic seat ticks against active turns at dispatch


Latand:
Follow-up to #1245 and the approved idle-turn-delivery design at 2d3a6f4b110a9109846e0cc0852a9fc82e80a465.

At main 84885e7127f2c819ac423191cabca009aacf92b6, an isolated controller regression produces one tick send while the seat turn is busy when liveness is absent, stalled, or gone. All three assertions expecting zero sends fail. The existing controller also omits a delivery policy; structured admission defaults to interrupt-active. The historical September interruption is corroborating evidence; its exact pre-check inputs remain unavailable.

Automatic checks may observe state while a seat works. Routine wake generation and engine input require positive idle evidence. Tools, awaited work, unknown state, stale lifecycle projections, and resumed turns under completed pipelines must remain protected. Actionable obligations remain owed until an admissible wake actually lands. Preserve unresolved attempts under their original immutable key through restart and rotation.

Scope: tick precheck, turn admission, delivery policy propagation, final host fence, and focused regression tests. No package/changelog, composer queue UI, production lifecycle, or incident-card closure. Required proof includes dispatch races, queued events after completion, busy versus stalled, unknown delivery retention, targeted tests, TypeScript, privacy and CI. Independent review and root-owned production verification follow the implementation.

Latand:
Implementation is ready for independent review in #1571.

Head: abb0537955ffdd50ffe9814c24a5dcd28f24ed7e
Base: 01601310a5c887bf41bfd317caa7996b74ddde69
Causal baseline: 84885e7127f2c819ac423191cabca009aacf92b6; the intervening main commit changes release metadata and its publication test only.

Isolated proof: the baseline controller sent a tick in each of three protected busy states; the baseline Codex delayed-response fixture wrote three starts before acknowledgement. The repaired cases produce zero busy-turn input and one serialized start respectively. Bun 1.4.0 targeted suites passed 658 tests; final focused identity, tool-wait and policy-persistence follow-ups also passed. TypeScript and whole-diff privacy checks passed. PR CI is running.

Production remains unmodified. Incident cards require root verification of their own conditions; neither this issue nor the PR claims they are resolved in production.

Latand:
REVIEW_READY: #1571 at f6efbae409f2b016c57581b7eaeeb599750cbb4e, base 01601310a5c887bf41bfd317caa7996b74ddde69.

[...]

Issue #1245, opened 2026-08-28
feat(monitor): a native, durable orchestrator tick that survives rotation — engine-independent and proactive


Latand:
## Operator requirement

The orchestrator's self-monitor must be a **native Viewer capability**, not something an agent sets up for itself inside its own session.

Stated requirements:

- It does not depend on Claude or Codex. It is ours.
- It ticks automatically, from the start, with no configuration by the user.
- When there is work, it wakes itself on an interval — roughly hourly.
- It can stop itself when it concludes a task cannot be done.
- Its default mandate is to carry every task to completion.
- It can read the open issues.
- Where it has access to logs, it can read them and notice errors.
- It is project-independent — the same mechanism works for any project.
- When there is no task, it is **proactive**: it gathers what exists, works out what matters, and proposes a list of actions.
- Once a task is planned in the Viewer, the task gets executed.
- It watches the board itself: that work is really moving and nothing is stalled.

## Why this is filed now

The monitor that has been driving orchestrator sessions was never a product feature. It was a prompt an orchestrator scheduled for itself, in-session:

- The string `SELF-MONITOR TICK` appears nowhere in this repository. It existed only in a running session's memory.
- The scheduling primitive available to an agent is session-scoped: nothing is written to disk, the schedule dies when the session exits, and it expires after seven days regardless.
- So every rotation silently dropped the monitor. A successor seat starts with no tick at all and does not know it is missing one, because there is nothing on disk to be absent.

This was observed directly: after a rotation the outgoing session kept ticking for eight hours alongside the incoming one — two orchestrators acting on the same repository — while the incoming seat had no schedule whatsoever.

## What already exists, and must be reconciled rather than duplicated

`src/lib/monitor/` is in the default branch, from #741. It carries `run.ts`, `classify.ts`, `journalStore.ts`, `viewerApi.ts`, `cards.ts`, `evidence.ts`, `requests.ts`, `redact.ts` and their tests.

The first question for this work is therefore **not** "how do we build a monitor" but "what does the existing one already do, why is nothing driving it, and what is missing between it and the requirements above". A second parallel mechanism would be the worst outcome.

## Research to do first

1. Reconstruct how the session-scheduled monitor actually behaved across recent orchestrator sessions: what a tick did, what it caught, and what it missed. Ground every claim in transcript evidence.
[...]

PR #1571, opened 2026-09-08
fix(monitor): admit seat ticks only at an idle turn boundary

Latand:
Refs #1569 and #1245. Implements the automatic-tick slice of the approved idle-turn-delivery design at `2d3a6f4b110a9109846e0cc0852a9fc82e80a465`, including the operator's subsequent settlement/countdown directive.

While a seat works, the existing controller collects and prepares actionable obligations in its durable accounting state. It delivers no routine tick during model work, tool calls or waits. On authoritative turn settlement, retained work is refreshed and drained immediately. With an empty queue, the full configured default interval starts at that idle boundary. A genuine event during idle bypasses the remaining interval; busy activity cancels eligibility. Prose saying done cannot release the fence. Existing empty-board/proposal suppression remains after expiry.

The existing controller owns observation, notifications and its nearest deadline. Generation/cursor-bound timing survives restart conservatively. Superseded work is refreshed before admission; accepted unknown attempts retain their immutable original payload/key. Only a landed wake acknowledges its carried obligations. Final adapter admission fences an active/idle race, including previously queued ticks.

Earlier review findings:
- FIXED: authoritative Codex start refusal releases the pending reservation; uncertain handover keeps it. Corrected input can start after refusal, and a matching authoritative settlement releases a lost-ack fence without duplicating input.
- FIXED: settled and authoritatively dead/stalled seats can use existing owner recovery. Input still requires positive idle evidence. Active, retrying and unobservable turns remain protected.

Validation uses isolated homes, state and provider roots under Bun 1.4.0. Baseline causal fixtures sent ticks to busy seats and duplicated delayed turn starts. The corrected lane passed 797 tests across 16 specific files; a subsequent reclaimed-idle regression and fresh monitor/controller/source/signal run passed 299 tests. TypeScript, whitespace and privacy checks passed. Tests include full-interval settlement, immediate queue drain, tool waits, event-during-timer, dispatch races, restart/rotation, cancellation, original-key uncertainty, explicit refusal and authoritative recovery.

The four existing incident cards remain open for root-owned production verification. Stalled-seat and unmerged-PR retry cards require authoritative recovery and a successful admissible owed wake; the child-source card requires complete child-discovery evidence; the unresolved-wake card requires authoritative settlement under its original key and correct accounting. None closes from local tests alone.

Required root production probe after independent exact-head review and an approved release: observe a designated seat through model/tool/wait activity with zero automatic input; queue a genuine event while busy and verify exactly one refreshed drain after authoritative settlement; verify an empty queue waits the full interval; introduce an event during that timer; race a user turn against dispatch; exercise stalled-owner recovery and original-key uncertainty across restart/rotation in a scoped fixture. Observe actual production generation and receipts without interrupting an existing working host. No deployment or live lifecycle mutation is included here.

### DIFF
--- a/src/lib/accounts/migration/contracts.ts
+++ b/src/lib/accounts/migration/contracts.ts
@@ -313,7 +313,7 @@ export interface DurableQuotaObservation {
 export interface HeldDeliveryCommand {
   operationId: string;
   kind: "send" | "steer";
-  policy: "queue" | "steer-if-active" | "interrupt-active";
+  policy: "queue" | "idle-only" | "steer-if-active" | "interrupt-active";
   turnId?: string | null;
   /** Message authorship stamped at admission (#1117), persisted on the held
       record so a migration-held delivery replays with the same attribution.
--- a/src/lib/agent/registry.ts
+++ b/src/lib/agent/registry.ts
@@ -1646,7 +1646,7 @@ function canonicalHeldDeliveryCommand(
   const command: HeldDeliveryCommand = {
     operationId: value?.operationId || deliveryId,
     kind: value?.kind === "steer" ? "steer" : "send",
-    policy: value?.policy === "queue" || value?.policy === "steer-if-active"
+    policy: value?.policy === "queue" || value?.policy === "idle-only" || value?.policy === "steer-if-active"
       ? value.policy
       : "interrupt-active",
   };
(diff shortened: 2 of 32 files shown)


Latand:
Closing: 19 of 32 files conflict and two review findings were never fixed; #1569's missing piece (an idle re-check between the tick verdict and delivery) goes to a fresh small PR.

Issue #1577, opened 2026-09-08
fix(runtime): decode Codex stdout across UTF-8 chunk boundaries


Latand:
An accepted multilingual user message can reach Codex and receive a response while its delivery receipt becomes uncertain with "Codex queue entry id belongs to a different payload".

Read-only incident evidence binds the original client key and operation to a byte-identical canonical structured-user message. The started runtime echo is intact; the completed echo replaces one two-byte Cyrillic character with two replacement characters. The operation dedup hash matches the canonical marker, so this is a corrupted transport echo rather than a changed submitted payload.

Causal reproduction: split the completed userMessage JSONL frame between the two UTF-8 bytes of a Cyrillic character. The current stdout listener calls String(chunk) independently and produces the exact payload-conflict error. The Claude adapter already uses a streaming StringDecoder.

Scope: Codex stdout decoding and focused Unicode chunk-boundary tests, in a separate commit from #1569 scheduling work. Preserve payload comparison and immutable original keys. Do not accept genuinely mismatched payloads or resend uncertain operations.

The existing incident remains untouched. A separate supported canonical-evidence settlement path may be needed: current receipt lookups return terminal unverified records without revisiting canonical evidence. Recovery must verify the original key, conversation/generation, exact payload and canonical record, then settle without replay. Frontend recovery and stale-receipt behavior remain with their owners.

Issue #1576, opened 2026-09-08
fix(delivery): reconcile a proven original delivery after payload-confirmation uncertainty


Latand:
A submitted operator message appears in the canonical transcript and is acted on, while its original delivery operation remains uncertain with a payload-collision error. The next browser submission stays queued behind it.

Read-only evidence on production 84885e71:
- The canonical structured-user dedup equals SHA-256 of the original operation ID.
- Decoded canonical text exactly matches the original persisted delivery text (1,301 characters and identical SHA-256).
- The registry retains failed/unverified and the runtime journal retains uncertain, both naming `Codex queue entry id belongs to a different payload`; the receipt has no turn ID.
- Host is hosted/idle and deliverable. No resend, discard, browser mutation or host action was performed.

Frontend behavior is intentionally conservative: uncertain entries do not retire from text-only echoes and block serial dispatch. Investigate the structured host confirmation and original-operation settlement path; consume authoritative canonical identity plus full payload evidence without weakening actual collision checks. Publish an absorbing terminal receipt under the original operation/key so the existing frontend reconciliation advances the next queued payload in order.

Acceptance: isolated causal reproduction with a structured-user header carrying selected context, matching original dedup/payload, and delayed/failed confirmation. One recipient effect; original key/payload retained; terminal delivered evidence propagates through query/snapshot/restart; the next queued message becomes eligible. Wrong operation, altered payload, ambiguous or unreadable evidence must remain unresolved. No live resend or receipt fabrication.

Related: #1490, #1524, #1564. Frontend lane remains read-only on runtime/registry files; backend owner required for original-operation settlement. Private identifiers, transcript and payload remain outside GitHub.

Latand:
The live evidence now isolates a UTF-8 transport failure. For the same original operation and turn, the structured host records a correct `userMessage` started event and a completed event in which one two-byte Cyrillic character becomes two U+FFFD replacement characters. The canonical transcript and persisted original delivery text match exactly.

`src/lib/runtime/codexAppServerHost.ts:1208` passes `String(chunk)` from each stdout data event into the string frame buffer. A private differential replay splits the observed original payload inside that character: independently converting both buffers reproduces the recorded corrupt completed payload exactly; an incremental UTF-8 StringDecoder preserves the original text. Six assertions pass. No production code or live state was changed.

[...]

PR #1579, opened 2026-09-08
fix(delivery): reconcile canonical original receipts without replay

Latand:
Refs #1576 and #1577.

Codex stdout was decoded one Buffer at a time. Splitting a Cyrillic character between chunks changes the completed user echo into two replacement characters, so a message that reached the recipient becomes delivery-uncertain. Decode stdout incrementally, flush at EOF, and recover a damaged confirmation cache only from exact canonical evidence. Canonical response_item user records now participate in the existing bounded dedup reader, including selected-context comparison.

Add `POST /api/runtime/operations/{originalOperationId}` with `{"action":"reconcile-delivery"}`. The request supplies no delivery claim or replacement key. The runtime host checks the persisted operation/client/payload binding, one attributable native generation/turn, and the matching canonical record. A transaction fences receipt revision, host epoch and discard/retry arbitration before recording an absorbing delivered receipt. It never invokes send, retry, queue activation, or host control. Original-key queries recover a stale registry projection after a journal-first crash.

Validation:
- A split UTF-8 fixture reproduced the exact payload-conflict error before the decoder fix.
- Two-, three-, and four-byte splits, selected-context headers, duplicate calls, true collisions, poisoned caches, and incomplete EOF are covered.
- Twelve focused canonical recovery tests pass: HTTP/RPC, restart/snapshot/query, generation rotation, wrong key/payload/context, missing and ambiguous evidence, concurrent discard, stale uncertainty, and replacement-key refusal.
- Existing adapter, HTTP, receipt, journal, operation-query and structured-delivery suites passed in fresh isolated processes under Bun 1.4.0. TypeScript and privacy checks passed.
- A private replay of the captured operation, runtime echoes and exact canonical message settled the original operation/key with its original turn and digest, with zero pending effects and zero live writes.

Recovery requires this Viewer and runtime-host code to be deployed together. Evidence outside the bounded retained journal window stays unresolved. Root retains deployment and the production recovery call; no live receipt, host, queue, browser, or transcript was changed. Frontend PR #1575 and tick PR #1571 remain separate.

### DIFF
--- a/src/lib/runtime/canonicalDeliveryRecovery.test.ts
+++ b/src/lib/runtime/canonicalDeliveryRecovery.test.ts
@@ -19,7 +19,7 @@ function fixture() {
   const registry = new AgentRegistry(path.join(directory, "registry.json"));
   const conversation = registry.ensureConversation("codex", transcript, null);
   const filename = path.join(directory, "runtime.sqlite");
-  let journal = new RuntimeJournal(filename, { structuredHosts: true });
+  let journal = new RuntimeJournal(filename, { structuredHosts: true, maxEvents: 100_000 });
   const operationId = "original-operation";
   const clientKey = "original-client-key";
   const text = "[viewer context — selected reviewer]\nOriginal т instruction 🎯";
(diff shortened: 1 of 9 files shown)


Latand:
Review fixes at `0559846b827bc4a8816ecc9f4627ab0237c68d60` (base `01601310a5c887bf41bfd317caa7996b74ddde69`):

- FIXED F1: reject incomplete journal coverage at both the 256-item and 50,000-sequence bounds, and require the original admission revision.
- FIXED F2: fence session evidence and retention revisions inside settlement. Canonical file identity, size and timestamps must also remain unchanged during verification.
- FIXED F3: independently hash decoded canonical text, compare it alongside the content digest and selected context, and enforce text identity in cached confirmations.
- FIXED F4: recovery and damaged-cache repair use the complete canonical index and refuse incomplete framing. A collision before the final 16 MiB is covered.

The five original negative cases reproduced incorrect delivered settlement before the fix. All 18 recovery tests now pass, including a sixth regression for canonical changes during verification. The host suite passed 138 tests; the final text tightening passed nine focused host tests. Runtime journal: 80 passed. TypeScript, privacy and whitespace checks passed. CI and fresh independent review are required at this head.

The previously proposed live recovery is cancelled. No live reconciliation, resend, lifecycle mutation, merge or deployment occurred.

Latand:
Closing: the cause of #1576, a UTF-8 character split across stdout chunks, is fixed on main by #1636 (`StringDecoder` in `codexAppServerHost.ts`); the remaining recovery route served only operations stranded before that fix.

Issue #1583, opened 2026-09-08
bug(runtime): a live idle structured host regresses to pendingAction spawn / synchronizing after delivered messages, and every later send fails no-claim


Latand:
## Symptom (2026-09-08, a third project's manager seat)

The operator's screenshot at 11:47 UTC shows a Codex manager seat with a live badge whose operator message fails with `not delivered — no-claim`, while a worker in the same project reports that the Viewer's explicit `resume` on that manager answers `structured recovery ownership is unavailable`, so a prepared handoff cannot be delivered.

## Evidence (read-only, agent registry and lifecycle journal)

- The manager's host record: `status: idle`, `pendingAction: "spawn"`, `claimOwner: structured-host:<runtime host pid>`, `structuredHost.process` names a Codex app-server process that is still alive, `structuredHost.publishedAt: null`, `host: null`, `updatedAt` 11:31:10 UTC.
- Deliveries to that conversation at 11:23, 11:29 and 11:30 UTC were delivered (two of them on the second attempt). Every delivery from 11:44 UTC on failed with `error: "no-claim"` (one with "delivery failed and remains recoverable"), from both operator and agent origins.
- `conversation_deliverability` answers `transport: structured, hostStatus: idle, pendingAction: spawn, processRecorded: true, deliverable: false, condition: synchronizing, reason: "the durable host record is still publishing ownership"` (`src/lib/conversation/deliverability.ts:122`).
- The worker's `conversation_action resume` returned 503 `structured recovery ownership is unavailable` from `src/lib/runtime/structuredControls.ts:117` (the host process is alive, so `republishStructuredDeliveryHost` returned false).

So a host that delivered three messages regressed to a record that still carries the initial `pendingAction: "spawn"` and has never recorded ownership publication, and republishing from the MCP host process does not repair it. The resume half matches #1423 (structured actions from the MCP host process have no runtime-host socket).

## Not the same defect as the rejected-launch issue

This is host ownership publication in the runtime host and registry, not MCP receipt classification. Filed separately so the launch repair stays bounded.

## Ask

- Explain why a record with three delivered messages keeps `pendingAction: "spawn"` and `publishedAt: null`, and why the 11:31:10 write regressed deliverability.
- Make `synchronizing` self-heal or become a terminal condition with a supported repair, so a live idle host does not answer `no-claim` indefinitely.
- Keep the failed deliveries' payloads recoverable; nothing here licenses discarding them.

Related: #1423, #974.

[...]

PR #1584, opened 2026-09-08
fix(runtime): restore structured manager delivery after route recovery (#1583)

Latand:


Latand:
## Refreshed onto current `main` — ready for review at the new head

`main` moved to `ded209129719d76ba1435d0605a5cde3619886f8` (the CI/launch repair, which also
carries the workflow fix for the apt failure that had both required checks red repo-wide).
Branch protection is `strict: true`, so this PR had to be brought up to date before it could
merge. Integrated by an **ordinary merge**, no rebase, no reset, no force-push.

**Head: `21cde0edd7977bfafecfe91f0c5b7bdb26e55736`**

| Check | Result |
|---|---|
| `git merge-base --is-ancestor ded20912 HEAD` | yes — current `main` is an ancestor |
| `git merge-base --is-ancestor 3c477010 HEAD` | yes — the previously approved head is preserved |
| Six owned files vs the approved head | blob hashes identical, byte for byte |
| `git diff --stat origin/main HEAD` | 6 files, +557 / −10 — exactly the approved change |
| Conflicts | none; zero file overlap with the 32 files `main` gained since the merge base |

No source was added, removed or rewritten in the refresh. The `registry.ts` hunks here stay
disjoint from the board lane's, so the two remain free to integrate in either order.

### The pre-existing `registry.test.ts` baseline is unchanged at the new merge base

Counts alone cannot settle attribution, so the **failing test names** were compared as sets,
one file per invocation, in an isolated environment (private `HOME`, `XDG_CONFIG_HOME`,
`LLV_STATE_DIR`, provider homes, short `TMPDIR`) — never against live state.

| Tree | default backend | `LLV_AGENT_REGISTRY_SQLITE=sqlite` |
|---|---|---|
| old merge base `7e74cb7f` | 123 pass / 0 fail | 93 pass / **30 fail** |
| current merge base `ded20912` | 123 pass / 0 fail | 93 pass / **30 fail** |
| this head `21cde0ed` | 127 pass / 0 fail | 97 pass / **30 fail** |

The 30 failing names are the **same set** at all three points (`diff` empty in every pair), and
the test file is byte-identical at both bases. The 123 → 127 delta is this PR's four new cases,
green under both backends. The 30 are inherited; nothing here introduced or masked them.

### The repair predicate, exercised against the live stuck rows — read-only

The shipped `repairCompletedStructuredSpawnMarkers()` was run on a **copy** of the registry
mirror, with production's own liveness fence reading real process evidence. The live store and
its WAL were never opened; the only write was to the copy. An empty-registry control returned 0.

- 5460 entries, 6988 receipts. 360 rows pass the repair's opening gate; **7** also carry a
  recorded structured process — the shape this issue is about.
- The predicate repaired **7 of 7**. Every gate passed on every row: exactly one matching
  completed structured launch receipt, epoch agreement, complete and `alive` process *and*
  claim-owner identities, current unsuperseded generation, receipt conversation resolving to
  the row. The one row carrying an operation id took the single-receipt path; the rest took
  the legacy exact-evidence scan.
- Deliverability moved from `deliverable: false / synchronizing` to
  `deliverable: true / deliverable` for all seven, so the rows land in the state the issue asks
  for rather than merely losing a field.
- **0** rows of the protected classes moved (`dead`, `unhosted`, `handoff`, or no recorded
  process). Terminal external-worker rows are refused by the predicate's first clause.
- Cost: 1899 ms, once, at startup, on that production-sized copy.

Four of the five rows reported in yesterday's audit are still present and all four satisfy the
full predicate; the fifth no longer carries the marker. Three rows that did not exist at audit
time have since joined. The population is still growing, and the repair runs only at Viewer
startup, so a restart is what clears them.

### Gates at this head

`bunx tsc --noEmit` exit 0 · `registry.test.ts` 127/0 default and 97/30 sqlite (the inherited
set) · `structuredRecovery.test.ts` 18/0 · `structuredDeliveryController.test.ts` 2/0 — each
under both backends, one file per invocation · privacy publication gate PASS locally and again
in the pre-push hook · all PR check-runs green at this head.

The approval on record was bound to `3c477010`. This head is new, so it needs a fresh
independent review before merge; the diff against `main` is unchanged from what was approved.

PR #1588, opened 2026-09-09
fix(composer): unknown deliveries no longer mute a conversation, plus Next 16.3.3 security patch (#1587)

### LINKED ISSUE #1587: Composer: an unknown delivery holds the serial send fence forever, muting the conversation
## Summary

A composer submission whose request leaves the browser but never returns an
answer is parked as `state: "delivering"` + `deliveryUncertain: true` (#1538).
The serial dispatcher refuses to send while any such entry is present, and
nothing can ever take that fence back. **Every later message in that
conversation queues forever: the conversation becomes permanently mute.**

## Mechanism

1. The request dies before admission (wire death, or a structured response that
   carries no receipt). `markOutboxUnknown()` parks the entry as
   `delivering` + `deliveryUncertain`, with **no `operationId` and no
   `deliveryReceipt`** — the server never admitted an operation for it.
2. `nextDispatch` / `claimOutboxDispatch` (`src/components/conversation/outbox.ts`)
   refuse to dispatch while any entry is `delivering` and not `launchOwned`.
3. Nothing bounds that wait:
   - No receipt can ever arrive, because there is no operation to emit one.
   - The receipt-progression effect that normally advances a `delivering`
     bubble ("so it never sits `delivering` forever and blocks the serial
     dispatcher") only fires when a receipt matches.
   - `releaseReconciliationToRetry` marks the entry `failed`, but the local
     `unconfirmedReceipt` it publishes immediately re-projects it as
     `delivering` + uncertain — correctly, since the fate really is unknown —
     which puts the fence straight back.
   - A remount restarts a reconciliation only for records carrying the
     `reconciling` marker, which is written on one path only (the admission
     deadline). An entry stranded by wire death restores with no owner at all.
4. `persistedQueue` returns a `deliveryUncertain` entry unchanged, so the parked
   state survives every reload.

Result: one unlucky submission mutes the conversation for good.

## Observed

A conversation whose composer had one parked unknown entry and one later
submission sitting `queued` behind it. The queued message showed no
`dispatchedAt` and no `operationId`, and no send operation for it existed in the
runtime journal — it was never dispatched at all. The parked entry was over
14 hours old at that point, with no path to settlement.

## Expected

An entry whose fate is unknown must keep everything that is true about it — its
key, its payload, its unknown outcome, its ability to settle on a genuine late
receipt, and its refusal to be locally replayed or cancelled. It must **not**
keep this browser's in-flight fence, because the browser is no longer waiting on
a request for it and no evidence can arrive to release it.

Ordering is unaffected: each submission carries its own idempotency key and the
server orders by admission, so a later message cannot overtake an earlier one
that was genuinely admitted.

## Scope note

#1538's own tests encode "the serial slot stays taken", but the scenario they
cover is one where the original *does* settle. The never-admitted case makes
"until it settles" unbounded, which is what this issue fixes.

Latand:
Fixes #1587.

An image-bearing submission with an unknown outcome could keep every later message queued indefinitely, including after a reload. `nextDispatch` and `claimOutboxDispatch` now exclude `deliveryUncertain` entries from the local in-flight fence. The composer's existing busy/reconciliation gate still controls active waits.

The original entry retains its payload, key, unknown presentation, replay/cancel restrictions, and late-receipt settlement. This change authorizes a newer submission to proceed; it does not establish whether the original executed. The server continues to order admitted operations. An unknown request that is admitted later can still arrive later.

The release also pins Next.js to 16.3.3, resolving the two critical advisories that blocked CI: [Windows-hosted RCE](https://github.com/advisories/GHSA-p293-qw3h-jr36) and [AVIF image-optimization RCE](https://github.com/advisories/GHSA-2xp9-vwfh-vxw4). The still-required flight-manifest workaround is re-pinned and verified in both CJS and ESM after a clean frozen-lockfile install.

Validation at `5b9cafea0f9e444a9199fb1db4dd0e48c946494a`, based on `656e4a3b420dfddb2c8a7ea4641d33ab9cc8570b`:

- A restored-state regression through the structured send seam fails on the base and passes on the fix. Only the newer key is sent; the original image payload stays intact and a late authoritative receipt still settles it.
- Outbox: 82 passed; reconciliation expiry: 14 passed; stale-key: 2 passed; composer: 27 passed. Related queue, delivery, image, migration and echo checks passed.
- TypeScript, privacy checks, clean install and production build passed. Bun 1.4.0 serving checks loaded 22/22 modules with HTTP 200; runtime succession and both endpoint probes passed.
- `LogFeed.deliveryUncertainty` (30 failures) and `TmuxComposer.runtimeSnapshot` (7 failures) retain the same failing-test names as the base. These are existing failures, not passing suites.

All local tests and serving checks used isolated state and ephemeral ports. Hosted CI and independent review are required before merge and deployment.


Latand:
Deployed merged commit `bf60e71d6cdea09cd918a0825a65ab23ce6a531f`. Viewer and runtime-host release records agree; deployment completed successfully, HTTP returns 200, and release readiness is true.

Post-release browser verification used a restored uncertain outbox entry and intercepted send requests. The newer submission proceeds under its own key; the original stays unknown with its image bytes intact and is never replayed. The same baseline scenario remained queued before the fix.

The incident message was delivered under its original key before deployment, with one canonical text/image record. Its delivered receipt was verified again after deployment. The conversation was resumed through the supported API after its idle process exited during the release, and is now idle and deliverable.

The substantive independent review approved the exact PR head. A separate merged-artifact verification confirmed the approval, identical tree, and green head/main CI after the flow controller failed to parse an empty retry. The failed retry remains preserved as history; no approval metadata was fabricated.

Issue #1456, opened 2026-09-02
CI: 'Install media inspection tools' step hangs for 20+ minutes in the privacy workflows, blocking every PR merge


Latand:
## What happens

Three times on 2026-09-02 (runs on PRs #1454 and #1455, both privacy-tracker-audit and privacy-publication) the step **Install media inspection tools** stayed `in_progress` for 20–30 minutes with no output, while the same step on other runs the same day finished in well under a minute. The PR's merge state stays BLOCKED for as long as the step hangs, so every lane's merge waits on it. Cancelling and rerunning the run, or pushing a new head, made it pass immediately.

Workflows: .github/workflows/privacy-publication.yml .github/workflows/privacy-tracker-audit.yml 

## Mechanism (hypothesis)

The step installs system packages at run time (apt/pip fetch to an external mirror). A stalled mirror connection has no timeout, so the runner waits for the job's default six-hour limit. Nothing retries.

## Expected

- `timeout-minutes` on the install step (2–3 minutes) and one retry, so a stalled mirror fails fast and the run reruns itself instead of blocking merges.
- Or drop the run-time install: cache the tools between runs, or vendor the inspection into the Bun script the gate already runs.
- The orchestrator cannot distinguish a slow job from a hung one from the PR checks view; a hard step timeout makes the failure visible.

Issue #1433, opened 2026-09-01
Stage spawn parks on "account mutation is busy in this process" instead of waiting or retrying


Latand:
## Summary

Starting a pipeline stage sometimes fails at spawn with `account mutation is busy in this process; retry shortly`, and the pipeline parks in `needs_decision` until an operator or seat calls `retry-stage`. On 2026-09-01 one lane needed three attempts (failed at 20:51Z and 21:02Z, succeeded at 21:17Z); an earlier lane the same evening failed once the same way (18:50Z). Each failure costs a wake, a manual retry and up to 15 minutes of idle lane time.

## Mechanism (suspected)

The spawn path takes the in-process account mutation lock (`withAccountMutationLock`) to pick the account for the new conversation. When another long operation holds it in the same Viewer process (a limits refresh, an account switch, a migration step, or several stage spawns landing in the same second from parallel lanes), the spawn gives up immediately instead of waiting, and the stage attempt is recorded as a decision-needing failure rather than a transient retry. The `account-selection.lock.queue` directory was empty when inspected afterwards, so this is contention, not a stale lock.

## Acceptance

- A stage spawn that meets a busy account mutation lock waits a bounded time (for example up to 30 s with backoff) and only then fails; the wait is logged with the holder's operation name.
- A spawn that still fails for this reason is retried automatically by the pipeline engine a bounded number of times (for example 3 over 5 minutes) before parking, and the attempt record names the retries.
- A regression test drives two stage spawns into the lock at once and shows both complete without a parked attempt.
- The parked-attempt error, when it does happen, names the holding operation instead of the generic sentence.

Latand:
Covered by the unified automation model design (#1446, merged in #1449): docs/design/automation-v2/README.md §4.2 traces this incident to a mechanism, and implementation slice 3 (§6.3) ships it. Keep this issue open as the acceptance evidence for that slice.

Latand:
## 2026-09-09 recurrence, mechanism confirmed on the main revision the Viewer serves

- 06:21:00Z: `create_pipeline` with `autoStart: true`, one Codex stage on a project-bound account. The attempt started at 06:21:05Z and parked at once: `needs_decision`, cursor `spawning`, `launchId: null`, `conversationId: null`, error `account mutation is busy in this process; retry shortly`. No process was ever created.
- 06:25:31Z: the same account, model and prompt launched through `spawn_agent` (the direct path) and was running within seconds.
[...]

Issue #1589, opened 2026-09-09
fix(turn-state): a tool call cut off by a restart keeps a Codex turn busy after the re-hosted continuation completes, so the stage never settles


Latand:
## Observed (2026-09-09, production, serving the current main revision)

- A pipeline builder stage (Codex) was mid tool call when a deploy restarted the Viewer at 05:36:27Z. The last record the old host wrote is a `function_call` at 05:36:42Z whose output never arrived: that call id occurs exactly once in the whole transcript.
- Startup re-hosted the conversation and delivered the standing continuation (`Continue the interrupted turn from the transcript.`, `src/lib/runtime/startup.ts:137`) at 05:38:44Z. The agent finished at 05:40:18Z with `task_complete` and a final message carrying a valid fenced verdict (`status: pass`, empty findings).
- 48 minutes later the attempt was still `running`, the review stage had no attempts, `agent_activity` reported `turnState: busy`, `lifecycle: stalled`, `reason: host_alive_transcript_silent`, `evidenceSource: transcript`, while `conversation_deliverability` for the same conversation reported `hostStatus: idle`. The lane owner paused the pipeline and spawned a reviewer by hand.

## Mechanism (confirmed by replaying the projection over the real transcript tail)

`turnStateFromRecords` (`src/lib/accounts/migration/turnState.ts:125`) keeps one open-tool ledger across the whole tail window. `task_started` (line 138) reopens the turn but does not clear the ledger, and `task_complete` (lines 144-150) closes the turn only when the ledger is empty. A tool call the restart cut off has no output record and never will, so it stays in the ledger and every later `task_complete` in the file reads as premature. The projection answers `busy` with `source: "tool"` for the rest of the transcript's life.

Replaying the function over the last 128 KiB of the transcript (80 records, `task_started` through `task_complete`) returns `{ state: "busy", source: "tool" }` with exactly one open call id: the cut-off call.

Everything downstream reads that one function: the scanner's activity projection (`src/lib/scanner/activity.ts:185`, `jsonl_turn_open`), liveness and `agent_activity` (`src/lib/lifecycle/liveness.ts:361`), and the pipeline engine's settlement gate (`src/lib/pipelines/durableEvidence.ts:157`; `src/lib/pipelines/engine.ts:2317` returns early while the scan projects an open turn, and `engine.ts:2430` returns while `durable.turn === "busy"`). The stage cannot settle, the next stage never activates, and the board shows a running agent over a finished transcript.

[...]

PR #1591, opened 2026-09-09
fix(pipelines): defer transient account contention before launch allocation

Latand:
## Summary

- Classify pre-allocation `account mutation is busy` failures as transient pipeline contention.
- Reuse the existing durable controller wait for 1/2/4/8-second backoff within the unchanged 30-second budget.
- Keep the same stage attempt through recovery, surface the current retry time, and clear it after a claimed host is returned.
- Preserve reserved-launch and unknown-receipt handling when a busy-looking error arrives after allocation.
- Keep controller-unavailable rounds from retaining an obsolete busy retry time, and remove the retry hint from terminal exhaustion details.

Refs #1433

Issue #1433 remains open for the deferred acceptance: the parked error should name the holding operation, and a regression should drive two stage spawns into the lock concurrently and show both complete. This slice covers the engine-owned bounded wait; the remaining lock and caller work stays discoverable there.

## Causal evidence

- The real-lock child fixture holds an async account mutation and calls the same synchronous registry reservation used by pipeline spawns: it gets `AccountMutationBusyError` in under 100 ms with zero receipts. The queued async control completes with one receipt for the same client attempt.
- Before the engine change, the regression went red with `needs_decision`, `launchId: null`, and no scheduled retry.
- After the change, contention stays pending on the existing controller wait, resumes automatically after the scheduled delay, and reaches one host claim with attempt `n=1`.
- A mixed busy/controller wait test verifies the board detail carries the current retry timestamp for busy contention, clears it during publication rebinding, and stays silent for controller-only failure.
- The duplicate-claim guard went red with a deliberate two-line claims fixture (received 2, expected 1), then passed with the restored one-line fixture. A two-process restarted-controller race produced one fresh claim; the waiting process observed the running record. A post-reservation busy-looking failure retained its launch identity and stayed on receipt recovery.

## Verification

- Base: `bf60e71d6cdea09cd918a0825a65ab23ce6a531f`
- HEAD: `0e2699be04d25d9f490dca5ad072869abc9a688c`
- `bun test src/lib/pipelines/engine.test.ts src/lib/accounts/accountMutation.test.ts` — 244 passed, 0 failed, 2,379 assertions.
- `bun test src/components/pipelines/PipelineHub.dom.test.tsx src/components/pipelines/PipelineStrip.render.test.tsx src/components/pipelines/pipelineModel.test.ts` — 139 passed, 0 failed, 385 assertions.
- `bun x tsc --noEmit --incremental false` — passed.
- `bun scripts/privacy-publication-gate.ts --base bf60e71d6cdea09cd918a0825a65ab23ce6a531f --check-commits` — passed.
- Actual Chromium synthetic board fixture — deferred/queued, bounded terminal failure, and materialized running-after-claim states were rendered and inspected; no browser errors.

No merge, deployment, live worker action, or runtime state mutation was performed.

---

## CI repair: privacy jobs blocked repo-wide (#1605)

`4cf0c9bb` adds one commit outside the pipeline change, touching only `.github/workflows/privacy-publication.yml` and `.github/workflows/privacy-tracker-audit.yml`. From 17:28Z on 2026-09-09 both privacy jobs died in their shared `Install media inspection tools` step on every branch, before the gate they exist to run. Google's Chrome index served a `Packages.gz` whose SHA256 does not match the value its own `InRelease` declares, and `apt-get update` fails as a whole when a single source fails. The runner image configures that source; neither job installs anything from it.

The update now reads the Ubuntu archives the four packages actually come from. On the 24.04 image those live in `sources.list.d/ubuntu.sources`, on 22.04 in `sources.list`, and the step covers both.

- `ffmpeg`, `tesseract-ocr`, `tesseract-ocr-eng`, `tesseract-ocr-ukr`, the 4-minute timeout from #1456 and its retry loop are unchanged, and both gates still run with OCR in `eng+ukr`.
- No apt error is ignored, nothing is `continue-on-error`, no checksum or signature check is relaxed, and no credential is involved. A package missing from the Ubuntu archives still fails the step (verified: exit 100, `Unable to locate package`).
- Restricting to `sources.list` alone — the first option sketched in #1605 — installs nothing on 24.04: the empty update prunes every list (15 → 0) and the install then finds no package. Verified before choosing the deb822-aware form.

### Evidence

Reproduced against the live upstream index in an `ubuntu:24.04` container configured with the same Chrome source, running the step scripts extracted from the workflow files:

| Subject | Result |
|---|---|
| Step as on `c5773de8` | exit 1, same `Hash Sum mismatch` on `chrome-stable/deb`, 6 Google source lines |
| Step as changed | exit 0, 0 apt errors, 0 Google lines, all four Ubuntu suites read, 289 packages configured |
| Tools after the changed step | `ffmpeg 6.1.1`, `tesseract 5.3.4`, languages `eng osd ukr` |
| Chrome source during the passing run | still configured — the restriction is what passes, not its removal |

Real CI at this head, dispatched at this ref so the changed file is the one that runs:

| Run | Result |
|---|---|
| [Privacy publication gate 34386486115](https://github.com/Latand/live-log-viewer-next/actions/runs/34386486115) | success — install crossed reading only `azure.archive.ubuntu.com`, gate behavior tests ran, `PRIVACY GATE: PASS` |
| [Privacy tracker audit 34386489295](https://github.com/Latand/live-log-viewer-next/actions/runs/34386489295) | success — install crossed, `#1591` audited, `PRIVACY GATE: PASS` |

Locally on this head: `privacy-publication-gate --base c686b6bc --check-commits` PASS (again in the `pre-push` hook), `bun test scripts/privacy-publication-gate.test.ts` 148 passed / 0 failed.

### Merge note

The two `pull_request_target` checks on this PR keep failing, and will until the fix is on `main`. That event runs the workflow file from the base branch, so this PR's own privacy checks execute `main`'s copy of the step: the failing run at this head still contacts `dl.google.com` and carries no `Dir::Etc` option. The dispatched runs above are the same jobs at the same commit with the changed file, and they pass.

`origin/main` at `c686b6bc` is merged in by ordinary merge. No deployment, restart, or live-state change was made.


Latand:
## Updated for review against current main

`main` moved while this PR was approved: [#1592](https://github.com/Latand/live-log-viewer-next/pull/1592) (the #1589 turn-state repair) merged as `6cc462e9`. This branch is now up to date with it and carries the §4.4 amendment that #1592's body deferred here.

**Head:** `82990b20ad2c46109bd00f38d9ecd88437a5cb5a` (was `0e2699be04d25d9f490dca5ad072869abc9a688c`, the approved head, which is an ancestor — ordinary push, no rebase and no force).

### What changed since the approved head

| Commit | What |
| --- | --- |
| `c7e3dff0` | Ordinary merge commit of `main` into this branch. History preserved. |
| `82990b20` | `docs/design/pipeline-account-contention.md` §4.4 — `user_message` now documented as clearing the tool ledger. |

**The merge auto-resolved with no conflicts.** The only file both sides touched is `src/lib/pipelines/engine.test.ts`, where each appends independent cases; the result is their union, with nothing removed or altered on either side (`git diff` against each parent adds only the other parent's lines and deletes none). `src/lib/pipelines/engine.ts` is byte-identical to the approved head, and `src/lib/accounts/migration/turnState.ts` is byte-identical to `main`.

### §4.4 amendment

§4.4 carved `user_message` out of the boundary reset on the premise that "a steer mid-turn must not drop the turn's own tools". The code that shipped in `turnState.ts` clears at that boundary with `task_started` and `turn_started`, because #1592's corpus measurement overturned the premise: a `user_message` arrives 55 times in 5 files while a named tool is outstanding, and none of the 506 calls outstanding at those steers was ever answered later in the same file. The section now states the shipped rule and records the measurement. Documentation only — it describes no behaviour the merged code does not already have.

### Gates at this head

Each test file in its own invocation, under an isolated `HOME` / `XDG_*` / `LLV_STATE_DIR` / `TMPDIR` and isolated provider homes:

| Suite | Result |
| --- | --- |
| `pipelines/engine.test.ts` | 238 pass / 0 fail |
| `accounts/accountMutation.test.ts` | 8 pass / 0 fail |
| `accounts/migration/turnState.test.ts` | 35 pass / 0 fail |
| `pipelines/durableEvidence.test.ts` | 24 pass / 0 fail |
| `scanner/activity.test.ts` | 24 pass / 0 fail |
| `lifecycle/liveness.test.ts` | 38 pass / 0 fail |

367 pass / 0 fail across the six files the merge touches. `bunx tsc --noEmit --pretty false --incremental false` exit 0. `privacy-publication-gate.ts --require-known-values --check-commits --base 6cc462e9` PASS. CI green on all 7 checks at this head.

**One CI note:** the first `Platform tests` run failed on `Windows (native)` with `the CLI names a missing Bun prerequisite before starting the Viewer` timing out at 10016 ms against a 10000 ms limit. That test lives in `bin/server-runtime.test.ts`, and the whole `bin/` tree is the same git object in this merge, in `main`, and in the approved head — all three identical, and the other two passed this job. The re-run passed. Runner flake on an unchanged file.

### Left deliberately out of scope

The code comment in `turnState.ts` describes the design doc as "shipping on a separate branch". Once this PR merges that parenthetical is stale. It sits in a file this PR does not own, and the scope here is the merge plus the §4.4 correction, so it is flagged rather than edited.

Please review the full PR against current `main`.

Issue #1598, opened 2026-09-09
test(spawn): failedSpawnDelivery.test.ts is fully red on main — 7 issue-653 regressions assert nothing


Latand:
## Symptom

`src/lib/agent/failedSpawnDelivery.test.ts` is **0 pass / 7 fail** on `main`. Every failure is the same throw:

```
error: title is required for every new spawn
    at beginSpawnRequestInFile (src/lib/agent/registry.ts)
    at ... makeFailedSpawnWithHeldDelivery (src/lib/agent/failedSpawnDelivery.test.ts:46)
```

## Why it is red

The file calls `registry.beginSpawnRequest` directly with a launch profile carrying no semantic title. Spawn admission now rejects that (`SPAWN_TITLE_REQUIRED_ERROR`), and this file was never adapted to the fixture helpers that exist precisely for tests predating the contract — `beginLegacySpawnFixture` / `withLegacySpawnFixtureTitles` in `src/lib/agent/registryTestFixtures.ts`, which cross the real admission boundary with a semantic title.

So the seven issue-653 regressions this file is supposed to hold — a failed spawn's held delivery terminalizing to `failed` and no longer being owed, `failStructuredSpawn` terminalizing the initial held delivery in the same transaction, the failed launch never projecting a delivering initial message — are currently asserting nothing.

## Baseline

Confirmed pre-existing, not from any branch in flight: run in a throwaway worktree detached at `6cc462e9485164f2fe8f7dd090b2ccfe291448a8`, the file is **0 pass / 7 fail** with identical errors.

CI does not run the suite, so a file in this state stays red on `main` indefinitely.

## Fix

Route this file's spawn creation through `beginLegacySpawnFixture` (or adapt its registry with `withLegacySpawnFixtureTitles`), then confirm the seven assertions still hold — they have not been exercised since the title contract landed, so at least one may have rotted behind the throw.

PR #1599, opened 2026-09-09
fix(accounts): stop dead pinned receipts from blocking account removal

### LINKED ISSUE #1595: bug(accounts): a dead account-pinned spawn receipt blocks removal of every account on that engine
## Observed

Removing a managed Codex account that failed its device sign-in does nothing. The row's `Remove` → `Confirm` bar closes and the account is still listed. `DELETE /api/accounts/codex` answers `409 account_removal_blocked` with `blockers: ["live_sessions"]`, and the panel reports "still has an active session or sign-in" — for an account with no sessions, no conversations, and no credentials (`authPresent: false`, `auth.state: "signed_out"`, `loginPending: false`).

## Mechanism

`accountHasLiveSessions` in `src/lib/agent/accountLiveness.ts`:

```ts
for (const receipt of Object.values(file.receipts)) {
  if (receipt.engine !== engine || !owned(receipt.accountId) && receipt.accountPin !== true) continue;
  if (receipt.accountPin === true && OPEN_RECEIPT_STATES.has(receipt.state)) return true;
  if (receiptIsLive(file, receipt, probe)) return true;
}
```

The `accountPin` branch was added in #1401. It has two independent defects:

1. **It never checks ownership.** The `continue` guard admits any pinned receipt of the engine, whatever account it is pinned to, and the branch that follows returns `true` without ever consulting `owned()`.
2. **It never probes liveness.** Membership in `OPEN_RECEIPT_STATES` alone is treated as a live session, with no `receiptIsLive` call and no `UNPROVEN_LAUNCH_GRACE_MS` bound.

Together, one launch receipt pinned to account B and stuck in `starting` — no pane, no admission owner, no verified host, created weeks ago — permanently blocks removal of account A, and of every other account on that engine.

## Evidence

Read-only replay of the pure blocker functions over this machine's registry snapshot, for a signed-out account A that owns zero conversations:

- `accountHasLiveSessions("codex", A)` → `true`; `liveAccountConversationIds("codex", A)` → `[]`.
- 13 codex receipts carry `accountPin: true` in an open state. **Every one** is pinned to an account other than A, **every one** has `receiptIsLive() === false`, and **none** carries a `queuedPinnedSpawn`.
- Oldest blocking receipt: 26 days old, `state: "starting"`, `pane: null`, `verifiedHost: null`, `admissionOwner: null` — far past the 5-minute unproven-launch grace.

## Relation to prior work

This is the class of defect #643 fixed ("managed account removal is permanently blocked by dead historical conversations and stale starting entries"), reintroduced through a new path. #643 made every liveness claim probe-backed; the pin branch added afterwards restored a probe-free claim beside it.

## Proposed fix

Scope the pin branch to the case #1401 meant to protect: a receipt carrying a durable `queuedPinnedSpawn`, which is queued for account capacity and legitimately has no probe-able process. Every other pinned receipt falls through to `receiptIsLive`, exactly as before #1401.

Not changed by this: a genuine queued pin blocking every account on its engine remains the deliberate contract asserted by the test "an aged queued pin blocks every account until its durable receipt settles".

Latand:
Closes #1595.

## What was broken

Removing a managed Codex account whose device sign-in had failed did nothing. `Remove` → `Confirm` closed the bar and the row stayed. `DELETE /api/accounts/codex` answered `409 account_removal_blocked` with `blockers: ["live_sessions"]`, and the panel reported "still has an active session or sign-in" — for an account with `authPresent: false`, `auth.state: "signed_out"`, `loginPending: false`, zero conversations and zero sessions.

The stale login record was not the cause, and neither was the UI: the confirmation path fires the request and the refusal notice does render. The backend refusal was real, and false.

## Mechanism

`accountHasLiveSessions` in `src/lib/agent/accountLiveness.ts` treated an account pin as a liveness claim. The branch added beside the probe in #1401 had two independent defects:

1. **It never consulted `owned()`.** The `continue` guard admits any pinned receipt of the engine whatever account it is pinned to, and the branch that follows returns `true` without checking ownership — so a receipt pinned to account B answered for account A.
2. **It never called `receiptIsLive`.** Membership in `OPEN_RECEIPT_STATES` alone counted as a live session, with no `UNPROVEN_LAUNCH_GRACE_MS` bound — so a receipt merely *in* an open state counted for ever.

Together, launch receipts abandoned in `starting` — no pane, no admission owner, no verified host — blocked removal permanently, on their own account and on every other account of that engine.

## The fix

A pin only records which account a launch would have used. It legitimately outranks liveness in exactly one case: a spawn still queued for account capacity, which has nothing to probe yet and will actuate later on the account it names. The branch is now scoped to a receipt carrying that durable `queuedPinnedSpawn`; every other pinned receipt falls through to `receiptIsLive`, exactly as before #1401.

This deliberately preserves the #1401 contract asserted by "an aged queued pin blocks every account until its durable receipt settles" — that test is untouched and still passes. No force path, no generic deletion, no change to what a genuinely live session blocks.

## Verification

**Isolated red/green** — `src/lib/accounts/removal.test.ts`, new case "dead pinned receipts block neither their own account nor any other (issue #1595)": RED at `["live_sessions"]` before the change with the file's other 25 tests passing, GREEN after.

**Replayed read-only over this machine's live registry** (a snapshot parsed in memory; no account, receipt or transcript was mutated, and nothing was deleted):

| account | before | after |
| --- | --- | --- |
| A — signed out, failed login, 0 conversations | `live_sessions` | removable |
| B — signed in, 1 live conversation | blocked | blocked |
| three further accounts with live conversations | blocked | blocked |

The 13 codex receipts doing the blocking were all pinned to accounts other than A, all `receiptIsLive() === false`, none carrying a `queuedPinnedSpawn`, the oldest 26 days old.

**Test files run individually** (all in this worktree, after `bun install`): `removal.test.ts` 26 pass, `accountLiveness.test.ts` 1 pass, `rateLimit.test.ts` 22 pass, `SwitchCard.status.render.test.tsx` 4 pass, `api/accounts/codex/route.test.ts` 10 pass, `api/accounts/claude/route.test.ts` 16 pass — 0 fail.

`bunx tsc --noEmit` exit 0. Privacy publication gate PASS against the merge base with `--check-commits`.

**One pre-existing red file, not from this branch:** `src/lib/agent/failedSpawnDelivery.test.ts` is 0 pass / 7 fail, every failure `title is required for every new spawn` thrown at spawn admission in `registry.ts`. Baselined in a throwaway worktree at the merge base `6cc462e9`: identical 0 pass / 7 fail there. Filed separately as #1598.

### DIFF
--- a/src/lib/accounts/removal.test.ts
+++ b/src/lib/accounts/removal.test.ts
@@ -302,6 +302,20 @@ test("an aged queued pin blocks every account until its durable receipt settles"
   expect(accountRemovalBlockers("claude", "other", DAYS_LATER)).toEqual(["live_sessions"]);
 });
 
+/* Production shape of issue #1595: launch receipts pinned to an account, left
+   in `starting` by pipelines that ended days ago — no pane, no admission owner,
+   no verified host — and carrying no queued pin to actuate later. A pin records
+   which account a launch would have used; it is not evidence that anything is
+   running, so it may not outlive the launch it belonged to. */
+test("dead pinned receipts block neither their own account nor any other (issue #1595)", () => {
+  const store = registry();
+  const own = beginLegacySpawnFixture(store, { engine: "codex", cwd: "/repo", accountId: "other", accountPin: true });
+  if (own.kind !== "created") throw new Error("expected a launch receipt");
+
+  expect(accountRemovalBlockers("codex", "work", DAYS_LATER)).toEqual([]);
+  expect(accountRemovalBlockers("codex", "other", DAYS_LATER)).toEqual([]);
+});
+
 test("dead history plus stale starting entries and receipts no longer block removal (issue #643)", () => {
   const store = registry();
   // Production shape: ~dozens of historical conversations whose latest generation
--- a/src/lib/accounts/removal.test.ts
+++ b/src/lib/accounts/removal.test.ts
@@ -316,6 +316,20 @@ test("dead pinned receipts block neither their own account nor any other (issue 
   expect(accountRemovalBlockers("codex", "other", DAYS_LATER)).toEqual([]);
 });
 
+/* The opposite guard of the case above. Dropping the probe-free pin claim may
+   not cost a pinned launch its blocker while that launch is still real: inside
+   the unproven-launch grace the fall-through to `receiptIsLive` must keep
+   blocking the account the pin names — and only that account. Turning the
+   fall-through into a `continue` would make removal race a live launch. */
+test("a pinned receipt still inside its launch blocks the account it names, and only that one", () => {
+  const store = registry();
+  const begun = beginLegacySpawnFixture(store, { engine: "codex", cwd: "/repo", accountId: "work", accountPin: true });
+  if (begun.kind !== "created") throw new Error("expected a launch receipt");
+
+  expect(accountRemovalBlockers("codex", "work")).toEqual(["live_sessions"]);
+  expect(accountRemovalBlockers("codex", "other")).toEqual([]);
+});
+
 test("dead history plus stale starting entries and receipts no longer block removal (issue #643)", () => {
   const store = registry();
   // Production shape: ~dozens of historical conversations whose latest generation
--- a/src/lib/agent/accountLiveness.ts
+++ b/src/lib/agent/accountLiveness.ts
@@ -234,8 +234,16 @@ export function accountHasLiveSessions(
     if (entryIsLive(entry, probe)) return true;
   }
   for (const receipt of Object.values(file.receipts)) {
-    if (receipt.engine !== engine || !owned(receipt.accountId) && receipt.accountPin !== true) continue;
-    if (receipt.accountPin === true && OPEN_RECEIPT_STATES.has(receipt.state)) return true;
+    /* A pin that is still queued for account capacity is durable in-flight work
+       with nothing to probe yet: it will actuate on the account it names, so it
+       outranks liveness until its receipt settles. A pin WITHOUT a queued spawn
+       records only which account a launch would have used — it is no evidence
+       that anything runs, and treating it as such let receipts abandoned in an
+       open state weeks ago block removal for ever, on their own account and on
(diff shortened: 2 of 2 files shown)


Latand:
Review round 1 answered at `ea333cac9e9fe36fc03f70b67fca2c64b5049854`.

**Coverage note (removal.test.ts) — FIXED.** Added "a pinned receipt still inside its launch blocks the account it names, and only that one". It asserts both halves of the guard the diff redefines: inside the unproven-launch grace the fall-through to `receiptIsLive` still blocks the pinned account, while every other account of the engine stays removable. Verified it can fail: mutating the fall-through into `if (receipt.accountPin === true) continue;` turns that case red (26 pass / 1 fail) and nothing else in the file notices — which is the edit the note asked to fence.

**Queued pin blocking every account (accountLiveness.ts) — filed, not changed here.** Confirmed the reasoning: `queuedPinnedSpawnForReceipt` requires `queued.accountId === receipt.accountId`, so a queued pin can never actuate elsewhere, and a `retry-at` admission re-publishes the queue with an advanced `retryAt` rather than failing it, so the block can hold for a long time. Changing it flips the second assertion in "an aged queued pin blocks every account until its durable receipt settles", so it wants its own review: #1600.

**Receipt rot never terminalizing — filed.** Confirmed at the skip: a receipt whose `transport` is neither `structured` nor `tmux` with no effect and no runtime operation matches `continue` and is never failed, so the sweep's deadline and `UNPROVEN_LAUNCH_GRACE_MS` disagree despite the comment claiming they mirror each other. #1601.

**Environmental note** — `failedSpawnDelivery.test.ts` was already filed as #1598 before this round, with the merge-base baseline attached.

Gates at the new head: `tsc --noEmit` exit 0; `removal.test.ts` 27 pass, `accountLiveness.test.ts` 1, `accounts/{codex,claude}/route.test.ts` 26, `spawn/route.test.ts` 57 — 0 fail, run one file per invocation; privacy gate PASS from the merge base.

PR #1610, opened 2026-09-09
fix(mcp): recover an original-key send whose text admission trimmed

### LINKED ISSUE #1609: fix(mcp): original-key send recovery falsely reports a payload contradiction when admission trimmed the text
## What happens

An MCP `send_message` whose `text` ends in whitespace (a trailing newline is the common case) is admitted and delivered exactly once. Repeating the same call under the same `clientRequestId` with the same arguments — the documented original-key recovery — answers:

```
outcome:  unknown
evidence: delivery-record
reason:   the delivery payload contradicts the bound request
ids:      {}
```

with no operation, conversation or delivery id. The caller is told its payload contradicts its own bound request, so an already-delivered send is hidden behind a conflict that does not exist. Under `unknown` the contract permits only further lookups under the same key, and every one of them returns the same answer, so the send stays permanently unrecoverable while having actually arrived.

Reproduced on the serving revision and on `main` (001776e8 and 9f398dde behave identically).

## Mechanism

Admission normalizes the text; recovery compares the un-normalized argument against it.

1. `src/lib/mcp/bindings.ts` `sendMessage` posts the caller's raw `args.text` to the conversation-host route.
2. `src/app/api/conversation-host/handlers.ts:445` hands `payloadText.trim()` to structured admission, so the reservation text and its `contentDigest` hold the **trimmed** text.
3. `src/lib/mcp/bindings.ts:4002` `recoverSend` passes the **raw** `args.text` to `resolveOriginalSend`.
4. `src/lib/runtime/sendSettlement.ts:717-723` compares that raw text and its digest against the stored trimmed text/digest and returns `contradictory`.

The rest of the lookup — target resolution, key match, owner/reservation consistency — is correct. Only the payload comparison is asymmetric.

The repository already treats the trimmed form as the canonical one on the read side: `src/components/feed/deliveredOccurrences.ts:61-68` digests a transcript row's trimmed text as "what every admission surface holds" and keeps the verbatim text as a second candidate. The second form is genuinely reachable in durable records, because the legacy (non-structured) admission path stores `payloadText` untrimmed (`handlers.ts:461`, reserved at `src/lib/delivery.ts:741-761`).

## Fix

Compare the recovery argument against the forms an admission surface can have stored, rather than against the raw argument alone. The first-layer binding and conflict rules stay exactly as they are, and a payload that differs by anything other than surrounding whitespace must still resolve `contradictory`.

## Acceptance

- A send whose original text carries leading/trailing whitespace is admitted and delivered once; original-key recovery with identical arguments reports the actual operation with its real ids — no second POST, no new key, no replay, no fabricated delivered time or acknowledgement.
- Recovery under the same key with genuinely changed text still answers `unknown` / "the delivery payload contradicts the bound request" with no ids.
- A record admitted through the untrimmed legacy path still recovers under its original arguments.
- Settlement deadlines unchanged; no production send, replay or registry mutation in the tests.

## Not a duplicate

- **#1576** has a different cause. There the canonical execution's payload matches the stored delivery byte for byte, and the operation is held uncertain by the structured host's own `Codex queue entry id belongs to a different payload` confirmation. Here neither the journal nor the registry records any collision — the Viewer's recovery read invents the mismatch from its own comparison. Neither fix advances the other.
- **#1490** established the original-key recovery contract that this defect breaks.
- **#1582** covers pre-dispatch validation refusals classified unknown, and successor recovery, for `spawn_agent`.
- **#262** is dead-host/Stop retry idempotency.

Scope: the normalization/comparison and its MCP recovery tests only. The terminal-projection and retention gap found during the same investigation is separate work and is not addressed here.

Latand:
Fixes #1609.

## The defect

An MCP `send_message` whose text ends in whitespace is admitted and delivered exactly once. Repeating the call under the same `clientRequestId` with the same arguments — the original-key recovery the contract directs callers to — answers `unknown` / "the delivery payload contradicts the bound request" with no ids. Under `unknown` the only permitted next action is another lookup under the same key, so nothing the caller can do ever clears it: an already-delivered send stays permanently unrecoverable, and its caller is told its own payload conflicts with its own request.

Reproduced on the serving revision and on `main` at 9f398dde, identically.

## Mechanism

Admission normalizes; recovery compared the un-normalized argument against the normalized record.

- `handlers.ts:445` hands `payloadText.trim()` to structured admission, so the reservation and its `contentDigest` hold the trimmed text. After settlement the text is blanked and that digest is the only content evidence left.
- `bindings.ts:4002` passes the raw `args.text` to `resolveOriginalSend`.
- `sendSettlement.ts` measured that raw text and its digest against the stored trimmed ones and returned `contradictory`.

## The change

`admittedMessageTextForms` names the forms a durable record can be holding — the trimmed text, plus the verbatim text that the legacy path on the same route reserves (`delivery.ts`) — and `lookupOriginalSend` matches its argument against those instead of against itself. Surrounding whitespace is the whole of the normalization, so any other difference is still a real payload conflict and still contradicts.

Nothing else moves: the binding, target, key, owner and reservation rules of the first layer are untouched, the MCP argument-digest conflict check is untouched, deadlines are untouched, and the lookup stays read-only. The seat monitor, the other caller of `resolveOriginalSend`, passes no text and is unaffected.

## Verification

| check | result |
|---|---|
| `bun test src/lib/mcp/originalKeySendRecovery.test.ts` | 6 pass |
| `bun test src/lib/runtime/admittedMessageText.test.ts` | 3 pass |
| `bun test src/lib/runtime/sendSettlement.test.ts` | 25 pass |
| `bun test src/lib/mcp/bindings.test.ts` | 60 pass |
| `bun test src/lib/monitor/seatTickController.test.ts` | 113 pass |
| `bun test src/app/api/tmux/route.test.ts src/app/api/conversation-host/route.test.ts` | 32 pass |
| `bun test src/lib/delivery.test.ts src/components/feed/deliveredOccurrences.test.ts` | 42 pass |
| `privacy-publication-gate --base 9f398dde` | PASS |

**Red without the fix.** With `src/lib/runtime/sendSettlement.ts` restored to its `HEAD` content and the new tests unchanged, the two whitespace-recovery tests fail with the production string "the delivery payload contradicts the bound request" and the other four — the changed-payload negative controls, the verbatim/legacy record, and the read-only assertion — pass on both sides.

The new test is an integration one on purpose: the real `conversationHostPOST` performs the admission trim, a real `AgentRegistry` reserves and settles, and the real MCP `send_message.recover` reads it back. Only the structured host boundary is a fixture, which is the seam the route's own suite already uses and is downstream of every transform under test. It covers a trailing newline, leading-and-trailing whitespace, an untouched text, four changed payloads that must still contradict and disclose no operation id, a verbatim legacy record, and that recovery admits nothing and leaves the durable records byte-identical.

## Scope

Normalization and its recovery tests only. The terminal-projection and retention gap found in the same investigation is separate work and is not touched here; neither are `registry.ts` nor board-owned files.

### DIFF
--- a/src/lib/mcp/originalKeySendRecovery.test.ts
+++ b/src/lib/mcp/originalKeySendRecovery.test.ts
@@ -1,0 +1,255 @@
+import fs from "node:fs";
+import os from "node:os";
+import path from "node:path";
+
+import { afterAll, expect, test } from "bun:test";
+
+import { NextRequest } from "next/server";
+
+/* Type-only, so the module graph still loads in the order the state root needs. */
+import type { McpRequestBinding } from "./server";
+
+/*
+ * Issue #1609: a delivered send stayed unrecoverable because the two halves of
+ * one request disagreed about its text. The send route TRIMS before it reserves,
+ * and original-key recovery compared the caller's untrimmed argument against the
+ * trimmed record — so a report ending in a newline came back
+ * "the delivery payload contradicts the bound request", with no ids, forever.
+ *
+ * Both halves here are the production ones: the real `conversationHostPOST`
+ * performs the admission normalization, a real `AgentRegistry` holds the
+ * reservation and settles it, and the real MCP `send_message.recover` reads it
+ * back. Only the structured host boundary is a fixture — that is the seam the
+ * route's own suite uses, and it is downstream of every transform under test.
+ *
+ * The suite owns a throwaway HOME/state root: importing these modules drags in
+ * everything that resolves state from the environment, and no test may read or
+ * migrate the operator's live state (AGENTS.md).
+ */
+const root = fs.mkdtempSync(path.join(os.tmpdir(), "llv-original-key-send-recovery-"));
+const previous = {
+  home: process.env.HOME,
+  xdg: process.env.XDG_CONFIG_HOME,
+  state: process.env.LLV_STATE_DIR,
+  codex: process.env.LLV_CODEX_HOME,
+};
+process.env.HOME = root;
+process.env.XDG_CONFIG_HOME = path.join(root, "config");
+process.env.LLV_STATE_DIR = path.join(root, "state");
+process.env.LLV_CODEX_HOME = path.join(root, "codex");
+
+const { setConversationHostDependenciesForTests } = await import("@/app/api/conversation-host/dependencies");
+
+afterAll(() => {
+  setConversationHostDependenciesForTests(null);
+  for (const [key, value] of [
+    ["HOME", previous.home],
+    ["XDG_CONFIG_HOME", previous.xdg],
+    ["LLV_STATE_DIR", previous.state],
+    ["LLV_CODEX_HOME", previous.codex],
+  ] as const) {
+    if (value === undefined) delete process.env[key];
+    else process.env[key] = value;
+  }
+  fs.rmSync(root, { recursive: true, force: true });
+});
+
+const { conversationHostPOST } = await import("@/app/api/conversation-host/handlers");
+const { AgentRegistry } = await import("@/lib/agent/registry");
+const { emptyLaunchProfile } = await import("@/lib/accounts/migration/contracts");
+const { sendDownstreamKey, viewerMcpRecoverableTools } = await import("./bindings");
+
+const transcriptPath = path.join(root, "recipient.jsonl");
+fs.writeFileSync(transcriptPath, "{}\n");
+const registry = new AgentRegistry(path.join(root, "agent-registry.json"), undefined, undefined, { sqliteMode: "off" });
+registry.reconcileConversations([{
+  engine: "codex",
+  path: transcriptPath,
+  accountId: "recovery-fixture-account",
+  launchProfile: emptyLaunchProfile({ cwd: root }),
+  turn: { state: "idle", source: "assistant", terminalAt: null },
+  observedAt: "2026-09-09T08:00:00.000Z",
+}]);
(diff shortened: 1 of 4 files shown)


Latand:
CI is green at this PR's own commit `58fe3077e21b89595f38a1805119054db92ab4fa`: platform tests (Linux, Windows), darwin-identity, audit, bun-runtime, privacy-publication and privacy-tracker-audit all pass.

The first `Linux (unchanged)` attempt failed on `bin/server-runtime.test.ts` → "the CLI rejects a ready runtime socket owned by another process", which touches nothing in this diff. It passes locally (26/26 in that file), `main`'s own platform-tests runs are green, and a plain re-run of the job at this same commit passed. Filed as #1611 with the mechanism — the CLI's socket-ownership refusal racing its singleton-fence refusal.

Latand:
Refreshed onto main with merge commit `6103d036`, pushed as a fast-forward. None of this PR's four files changed on main since its base, so its diff is identical to the approved head `83592613` (successor review lane 1dda8672: APPROVE, no findings).

Checked at `6103d036`:

- `originalKeySendRecovery`, `admittedMessageText`, `sendSettlement` and `mcp/bindings` tests: 95 pass, 0 fail
- `bunx tsc --noEmit --incremental false`: exit 0
- privacy gate with commit checking from main: PASS

The head moved, so an exact-head review of `6103d036` is needed before merge. It should confirm that the merge commit brings in only main's changes.

Latand:

Issue #1607, opened 2026-09-09
test(runtime): route-recovered settle clears pendingAction, but two structured-spawn integration cases still expect "spawn"


Latand:
## What happens

`src/lib/runtime/structuredSpawn.integration.test.ts` is **red on `main`**: 102 pass / 2 fail.

```
bun test src/lib/runtime/structuredSpawn.integration.test.ts \
         src/lib/runtime/structuredSpawn.terminalize.test.ts
# main: 102 pass / 2 fail
```

Failing cases:

- `issue 533: matching runtime and transcript evidence preserves a claimed path-pending structured host` (`:452`)
- `issue 533: an initial-message status timeout releases admission ownership and late success settles once` (`:1950`)

Both fail on the same field:

```
-   "pendingAction": "spawn",
+   "pendingAction": null,
```

## Mechanism

`AgentRegistry.recoverRouteSpawn` now composes a `settledCandidate` that keeps only a deliberate `handoff` marker and nulls every other `pendingAction` before `settleSpawnInFile(..., "route-recovered")`:

```ts
const settledCandidate = {
  ...candidate,
  pendingAction: storedEvidence?.pendingAction === "handoff" ? "handoff" as const : null,
};
```

That is the intended behaviour of the change — a route-recovered launch is terminal, so the completed host must not look as though it is still registering. The two integration cases above were written against the previous behaviour and still assert `pendingAction: "spawn"` on the settled entry, so they now contradict the rule they should be pinning.

## Bisect

| Revision | Result |
| --- | --- |
| `656e4a3b` (before the change) | 104 pass / 0 fail |
| `001776e8` (`main`) | 102 pass / 2 fail |

Green immediately before, red immediately after; nothing else in between touches these files.

## Why it reached `main`

No CI job runs this file. `platform-tests.yml` and `bun-runtime.yml` each run a named list of test paths, and `src/lib/runtime/structuredSpawn.integration.test.ts` is in neither, so the suite went red without any check going red.

## Fix

Update the two expectations to the new terminal contract (`pendingAction: null` after a route-recovered settle), and keep one case that pins the `handoff` marker surviving — that is the only branch the new code deliberately preserves.

Related: the same "red on main, run by no job" class is tracked separately for the board/scheme files and for #1597.

PR #1613, opened 2026-09-09
fix(runtime): repair delivery status after a lost terminal acknowledgement

### LINKED ISSUE #1612: fix(runtime): a lost terminal-transition acknowledgement leaves the delivery record uncertain, and compaction deletes the only proof
A message the recipient actually received stays `delivery-uncertain` in the durable delivery record, for as long as the Viewer keeps running. The runtime journal already says `delivered`; only the acknowledgement of that terminal transition is lost. Two real operator messages were answered by their recipients while their reservations still read uncertain.

Reproduced on both the serving revision and current `main`, in an isolated fixture built from the real `RuntimeJournal`, `AgentRegistry`, `bindStructuredDeliveryQueue` and the repository's fake engine host. Nothing about the algorithm is rewritten; only the RPC acknowledgement of the terminal transition is dropped, exactly once.

## Mechanism

1. The queue hands the message to the engine and writes the terminal transition: `src/lib/runtime/structuredDeliveryQueue.ts:939`.
2. The controller's transition port awaits the runtime host first and updates the registry only afterwards: `src/lib/runtime/structuredDeliveryController.ts:492-511`. A transport failure after the journal has committed skips the registry write entirely.
3. `transitionUnlessSettled` catches that failure, re-reads the durable receipt, sees a terminal status and returns `false`: `src/lib/runtime/structuredDeliveryQueue.ts:1104-1120`. It holds the proof in its hand and does nothing with it.
4. The completing transition clears the outbox row in the same transaction, so the effect never appears in another drain batch: `src/runtime-host/journal.ts:678`. Further drains cannot revisit it.
5. The only repair is `reconcileTerminalDeliveries`, and it runs at controller bind only: `src/lib/runtime/structuredDeliveryController.ts:1132`.
6. Journal compaction deletes the operation row on the ordinary event-count cadence, with no regard for whether that receipt was ever projected: `src/runtime-host/journal.ts:1361`.

## Isolated reproduction (real call sites, isolated state directory)

| case | registry after 3 drains | runtime receipt | recipient writes | after controller rebind |
|---|---|---|---|---|
| acknowledgement intact | `delivered` | `delivered` | 1 | `delivered` |
| acknowledgement lost | `delivery-uncertain` | `delivered` | 1 | `delivered` |
| acknowledgement lost, then compaction | `delivery-uncertain` | receipt row deleted | 1 | `delivery-uncertain` |

Once the receipt is compacted, the original identity also stops replaying: a re-drive under the original operation id and original idempotency key is admitted as a **new** queued operation (`replayed: false`) rather than replaying the terminal receipt. The durable evidence that makes an original-key re-drive safe is exactly what compaction removed.

## Repair

Own the terminal projection end to end, minimally and boundedly:

- Project the durable terminal receipt when the transition acknowledgement is lost, in the same drain pass, from the journal's own answer — not from what the caller hoped it wrote.
- Keep a bounded ongoing reconciliation for failures that arise after startup, so the repair no longer depends on a Viewer restart.
- Retain the terminal proof until the projection has been acknowledged, under a bounded cap, so compaction cannot delete evidence that still has a consumer. A runtime-protocol addition here must be backwards compatible in both skew directions, because Viewer and runtime host are released separately.

## Acceptance

- Lost terminal acknowledgement: exactly one recipient write, and the original reservation settles `delivered` without a restart and without polling a mutating receipt endpoint.
- Original operation id, idempotency key, content digest and the server's own delivery time are preserved; terminal precedence and the retry/discard fences are unchanged.
- Compaction before projection retains the proof; compaction after projection acknowledgement compacts normally.
- Retention pressure is bounded and proven bounded.
- Repeated drains and repeated reconciliation are idempotent: no replay, no second recipient write, no fabricated receipt.
- A receipt naming a different conversation, an unreadable journal, or an ambiguous fate leaves the record unknown. Absence after compaction is never read as non-execution.
- Backwards compatibility proven in both directions: a new Viewer against a runtime host without the addition behaves exactly as today; a host with it retains nothing extra for a Viewer that never asks.

## Deduplication

- #1576 is a different cause — canonical execution with an unresolved payload confirmation and a queue-entry payload collision. Here the journal already reads `delivered` and no collision is involved. Deliberately kept separate.
- #1490 (closed) supplies the original-key recovery contract this repair must not weaken; it does not project a terminal receipt into the delivery record.
- #1584 repaired stale route-recovered spawn markers; it does not touch this path.
- #1366 (closed) is recipient-side duplication from retries after an uncertain outcome, not a lost acknowledgement of a terminal transition.

Latand:
Closes #1612.

A message the recipient had already received stayed `delivery-uncertain` in the durable delivery record until the Viewer was restarted, and journal compaction could delete the only proof before that happened. Two real operator messages were answered by their recipients while their reservations still read uncertain.

## What was wrong

The controller's transition port writes the runtime journal and the delivery record in that order, so a transport failure between them commits the outcome and drops the projection. The completing transition clears the outbox row in the same transaction, so no later drain pass can rediscover the operation, and the only repair — the startup sweep — ran at controller bind alone. Compaction then deleted the receipt on the ordinary event-count cadence, with no regard for whether it had ever been projected.

## What this changes

- **Drain-time repair.** The queue hands an operation whose terminal acknowledgement was lost back to the controller, which re-reads the journal's own answer — never what the failed call meant to write — and settles the reservation from it. One recipient write, original operation id / client message id / content digest untouched, no replay.
- **Ongoing recovery, bounded.** An operation whose receipt could not be read at that moment stays owed and is retried at the head of every later drain pass, under a cap that releases the oldest. The drain is the clock, so the repair needs no timer of its own and stops as soon as the journal can answer.
- **Retention until acknowledgement.** A terminal transition can ask the journal to retain its receipt until the projection is acknowledged, inside the same write transaction that commits it — a marker written afterwards would be lost by exactly the event it exists to survive. Compaction honours that up to a cap applied oldest-first, reporting anything it releases; the acknowledgement clears it and the next compaction takes the receipt as it always did.
- One shared reading of a receipt now serves both the startup sweep and the drain-time repair, so "what does this receipt settle, and against which reservation" has a single answer.

Nothing here issues a send, raises a deadline, or writes a receipt. An unreadable journal, a receipt naming another conversation, and an absent receipt after compaction all leave the record unknown.

## Backwards compatibility (Viewer and runtime host are released separately)

Both protocol additions degrade in each skew direction, and both directions are tested:

- **New Viewer, old runtime host:** the retention flag is an unknown parameter the old transition handler ignores, and the acknowledgement method answers "unsupported", which the Viewer swallows. The drain-time repair needs no protocol addition at all, so it still works; retention behaves exactly as it does today.
- **Old Viewer, new runtime host:** nothing ever asks for retention, so nothing is retained.
- A journal file written before the new column migrates on open, with every existing row reading "owed nothing".

## Verification

Named tests only, all under an isolated `HOME` / `XDG_CONFIG_HOME` / `LLV_STATE_DIR` / `TMPDIR`, run on the pinned Bun 1.4.0.

- `src/lib/runtime/lostTerminalAcknowledgement.test.ts` — 10 pass. Lost acknowledgement, acknowledgement intact, repeated drains + rebind, compaction before projection, compaction after acknowledgement, unreadable journal, mismatched target, legacy host, refused acknowledgement, operation with no reservation.
- `src/runtime-host/terminalProjectionRetention.test.ts` — 11 pass. Retention, no retention when unasked, release, receipt left untouched, idempotent/invalid/oversize acknowledgement, retention pressure, repeated compaction, legacy-file migration, and the same obligation over the socket RPC.
- Red paths proven for each mechanism separately: removing the drain-time repair fails 5 cases; making compaction ignore the obligation fails 5; making the controller stop asking for retention fails 2.
- Adjacent suites by path, all green: `structuredDeliveryQueue`, `structuredDeliveryController`, `structuredDeliveryRebind`, `structuredDelivery.integration`, `structuredDeliveryLegacyVerdict.integration`, `structuredCompactDelivery`, `client`, `http`, `sendSettlement`, `journal`, `journalCompact`, `journalRetention`, `journalVacuum`, `journalSnapshotBounds`, `receiptSweep`, `socket`.
- `startup.test.ts` (1 fail) and `structuredSpawn.integration.test.ts` (2 fails, #1607) fail identically at the merge base — verified in a merge-base worktree, one file per invocation.
- `bunx tsc --noEmit --incremental false` → exit 0. `privacy-publication-gate --base <merge-base> --check-commits` → PASS.

## Scope

Delivery controller, delivery queue, runtime journal, its RPC surface, and the two new test files. `registry.ts`, `sendSettlement.ts` and `admittedMessageText.ts` are untouched — they belong to other lanes. #1576 (payload-confirmation collision) and #1609 (recovery whitespace normalization) are separate causes and are deliberately not addressed here.

Runtime-protocol note for release: this adds one optional transition parameter and one new socket method. It needs no live host change to be safe — an unpatched host simply retains nothing — but the full repair only holds once the runtime host is running this revision.

### DIFF
--- a/src/lib/runtime/client.ts
+++ b/src/lib/runtime/client.ts
@@ -101,6 +101,12 @@ export interface RuntimeHostClient {
     details?: { turnId?: string | null; queuePosition?: number | null; reason?: string | null },
     options?: RuntimeTransitionOptions,
   ): Promise<RuntimeOperationResult>;
+  /** Releases the retention a terminal transition took out under
+      `awaitProjection` (#1612), once its outcome is in the durable delivery
+      record. Optional because a runtime host from before it answers
+      "unsupported": the receipt then expires on the ordinary compaction
+      cadence, which is exactly the behaviour that predates this method. */
+  acknowledgeTerminalProjection?(operationIds: readonly string[]): Promise<number>;
   requestViewerDeployment(request: ViewerDeploymentRequest): Promise<ViewerDeploymentReceipt>;
   cancelViewerDeployment?(deploymentId: string): Promise<ViewerDeploymentStatus | null>;
   readViewerDeployment(deploymentId: string): Promise<ViewerDeploymentStatus | null>;
@@ -158,7 +164,11 @@ export class UnixRuntimeHostClient implements RuntimeHostClient {
       status,
       ...(details ? { details } : {}),
       ...(options.fromStatuses ? { fromStatuses: [...options.fromStatuses] } : {}),
+      ...(options.awaitProjection ? { awaitProjection: true } : {}),
     }) as Promise<RuntimeOperationResult>;
+  }
+  acknowledgeTerminalProjection(operationIds: readonly string[]): Promise<number> {
+    return this.call("operation-projection-ack", { operationIds: [...operationIds] }) as Promise<number>;
   }
   requestViewerDeployment(request: ViewerDeploymentRequest): Promise<ViewerDeploymentReceipt> { return this.call("viewer-deployment-request", request as unknown as Record<string, unknown>, this.deploymentTimeoutMs) as Promise<ViewerDeploymentReceipt>; }
   cancelViewerDeployment(deploymentId: string): Promise<ViewerDeploymentStatus | null> { return this.call("viewer-deployment-cancel", { deploymentId }) as Promise<ViewerDeploymentStatus | null>; }
--- a/src/lib/runtime/contracts.ts
+++ b/src/lib/runtime/contracts.ts
@@ -200,6 +200,14 @@ export type RuntimeReceipt = RuntimeOperationReceipt;
 export interface RuntimeTransitionOptions {
   /** Compare-and-set fence evaluated inside the journal write transaction. */
   fromStatuses?: readonly RuntimeReceiptStatus[];
+  /** Marks a terminal transition as owed a durable projection, inside the same
+      write transaction that commits it (#1612). The caller is saying: the
+      answer to this call is what carries the outcome into the delivery record,
+      so the receipt has to outlive compaction until the projection is
+      acknowledged — the acknowledgement is the one thing that can be lost
+      while the outcome itself is already committed. A runtime host from before
+      this option ignores it and retains nothing extra. */
+  awaitProjection?: boolean;
 }
 
 export function runtimePresentationReceipt(receipt: RuntimeOperationReceipt): RuntimeOperationReceipt {
@@ -725,7 +733,7 @@ export interface RuntimeReplay {
 
 export interface RuntimeSocketRequest {
   id: string;
-  method: "runtime-host-health" | "snapshot" | "events" | "wait" | "append" | "operation" | "command" | "operation-status" | "operation-delivery-action" | "operation-retry" | "effect-batch" | "operation-transition" | "producer-cursor" | "viewer-deployment-request" | "viewer-deployment-read" | "viewer-deployment-cancel" | "mcp-health-probe-admission";
+  method: "runtime-host-health" | "snapshot" | "events" | "wait" | "append" | "operation" | "command" | "operation-status" | "operation-delivery-action" | "operation-retry" | "effect-batch" | "operation-transition" | "operation-projection-ack" | "producer-cursor" | "viewer-deployment-request" | "viewer-deployment-read" | "viewer-deployment-cancel" | "mcp-health-probe-admission";
   params?: Record<string, unknown>;
 }
 
--- a/src/lib/runtime/lostTerminalAcknowledgement.test.ts
+++ b/src/lib/runtime/lostTerminalAcknowledgement.test.ts
@@ -1,0 +1,524 @@
+import fs from "node:fs";
+import os from "node:os";
+import path from "node:path";
+import { afterAll, expect, test } from "bun:test";
+
+/* Isolated state only: this suite drives the process-scoped delivery
+   controller, a real runtime journal and a real agent registry, none of which
+   may touch the operator's live state. */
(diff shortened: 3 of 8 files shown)


Latand:
Refreshed onto main with merge commit `26171797`, pushed as a fast-forward. Main's native queue work (#1636, #1661, #1665) conflicted in four places, resolved as follows:

- `contracts.ts`: the runtime-host method union carries both `operation-projection-ack` and the native-queue methods.
- `structuredDeliveryQueue.ts`: a drain pass starts native queue reconciliation, then retries the owed terminal projections.
- `journal.ts` transition: the `projection_pending` obligation is written in the same UPDATE, and the native `acknowledged`, `observed-queued` and `proven` phases still complete the outbox row.
- `journal.ts` compaction: rows held in `native_queue_operation_holds` survive together with pending effects and the capped projection obligations, and the released-receipt count leaves held rows out.

Checked at `26171797` under the pinned Bun 1.4.0, one file per run: this PR's `lostTerminalAcknowledgement` and `terminalProjectionRetention` tests, plus main's `journal`, `journalCompact`, `journalRetention`, `nativeQueueCompaction` (unit and integration), `nativeQueueJournal`, `structuredDeliveryQueue`, `structuredDeliveryController` and `structuredDelivery.integration` tests. 236 pass, 0 fail. `tsc` exits 0 and the privacy gate passes. Local Bun 1.3.3 segfaults while loading `nativeQueueJournal.test.ts`, and unmodified main crashes the same way, so that crash predates this merge.

This PR still has no independent code review; the earlier pipeline pass was a recovery audit. Merging needs a fresh review of `26171797` that covers the conflict resolution. Merge order: this head still conflicts with #1667 in `structuredDeliveryController.ts` and `structuredDeliveryQueue.ts`, so whichever of the two lands second needs another refresh.

Latand:

Issue #1371, opened 2026-08-31
pipelines: a usage-limit-killed stage attempt shows running/waiting forever, and a pinned account blocks failover to an allowed account with capacity


Latand:
Operator report (2026-08-31): lanes launched on an account that then exhausted its weekly quota showed "stages running" and the stage conversations showed "waiting for your answer" indefinitely; nothing moved to the other allowed account. Verified live today across three lanes.

## Mechanism (verified)

1. A stage attempt's turn ends with a terminal provider message: "You've hit your usage limit. … try again at <date>" (`event_msg task_complete` with `codex_error_info: usage_limit`, preceded by a `token_count` whose `rate_limits` carries the reached limit and `credits.balance: "0"`). The engine does NOT park the attempt: the attempt keeps showing `running` on the board for the next hour-plus, and the conversation renders as an idle turn ("waiting for your answer") although there is nothing to answer. The truthful state — "rate limited until <timestamp>" — exists in the transcript's own rate_limits/task_complete records and is surfaced nowhere.
2. A stage with a pinned `account` can never fail over: the pin is immutable after start (stage surgery is refused), so when the pinned account's weekly window is exhausted the lane is dead until the reset — today that meant "wait a week" — while the project's other allowed account sat at 6% of its weekly window. The only remedy was to close each lane and recreate it without the pin (three lanes today).
3. Only on `close` does the truth appear: the close report terminalizes the attempt "from the attempt transcript ending on a terminal provider message with no fenced verdict: You've hit your usage limit…".

## Wanted

1. A stage attempt whose transcript ends on a terminal provider usage-limit message parks promptly with that reason ("rate limited until <timestamp>, account <label>"), instead of showing running/waiting; the board card and pipeline stateDetail carry it.
2. On park-for-limit, if the stage's account is NOT pinned, the engine respawns the attempt on another account allowed by the project binding that reports capacity; if it IS pinned (or no allowed account has capacity), the lane parks with the earliest reset timestamp named.
3. The account picker/board surfaces "rate limited until <ts>" for an exhausted account so the operator sees why work moved or parked.
4. Tests: an attempt transcript ending in the usage-limit terminal message parks with the limit reason; an unpinned stage re-spawns on the other allowed account; a pinned stage parks naming the reset time.

## Related

#1279 (per-stage account selection), the account binding fence (allowed-set failover), #1325/#1326 (the same family: evidence in the transcript ignored by attempt-state recovery).

Latand:
[...]

Issue #1182, opened 2026-08-25
bug(orchestrator dock): stuck at "Opening the conversation…" after a viewer restart while the seat is live on the board; only a rotate clears it


Latand:
## Observed (2026-08-26 ~00:00, prod at `704ff463`)

After a viewer deploy/restart, opening a project's orchestrator dock showed only:

> Opening the conversation… — The orchestrator is designated; its transcript hasn't reached this view yet. [Open it on the board]

for minutes, while the same orchestrator conversation was visibly live on the board (card with PID, model badge, "Replaced an earlier orchestrator", fresh assistant turns, 24 % context). The incumbent header showed the engine/model/account, "opening" badge, and a Rotate button. Pressing Rotate produced a new seat and the dock opened normally. So the seat record was fine, the conversation was fine, but the dock never bound them.

## Mechanism (from the code, to be confirmed)

`src/components/orchestrator/seatState.ts:306`: liveness is `resolving` whenever `!file || surface === null || surface === "unresolved"`. `file` is the seat's transcript looked up in the board's file catalog by the path recorded in the seat; `surface` comes from the runtime plane. After a restart the structured host is re-hosted (the conversation continues under a successor generation/path), so either the seat's recorded path no longer matches a catalog entry (the catalog carries the successor path) or the runtime surface stays `unresolved` for the old host. Nothing retries or re-resolves through the durable `conversationId`, so the dock waits forever; a rotate writes a fresh seat and path, which is why it "fixes" it.

## Expected

1. The dock resolves the seat's conversation by durable `conversationId` (through the registry's current generation), never only by the recorded path; a re-hosted or migrated seat binds to its successor transcript on the next poll.
2. `resolving` is bounded: if the seat is `live` per `/api/orchestrator/seat/status` but the dock cannot bind within ~10 s, the panel shows the real reason (recorded path missing from the catalog / runtime surface unresolved) and offers "Re-bind" (re-resolve) instead of only "Open it on the board" and Rotate.
3. Rotate is never required to recover from a restart.

## Acceptance

- Deploy/restart the viewer with a live orchestrator seat: the dock shows the conversation within one poll after the viewer is back, no rotate.
- Test: seat recorded with a path that the catalog only knows under a successor generation → dock binds via conversationId.

Issue #231, opened 2026-07-14
Turn duration: 'worked for N' from prompt to last message — live while running, feed separator when idle (design pre-approved)


Latand:
Owner ask (2026-07-14): "хочу знати скільки агент працював від моменту промпта до того, як повністю зупинився (останнє повідомлення)".

## Design (Fable, authoritative)

**What is measured:** duration of the LAST completed turn = timestamp of the last assistant/tool output in the transcript minus the timestamp of the user prompt (or relayed message) that started that turn. While the agent is still working, the value ticks live (elapsed since the prompt). Multi-turn conversations show the current/most-recent turn only.

**Where it renders (two spots, same derivation):**
1. **In the feed** — a muted caption separator after the final message once the agent goes idle: «Працював 12 хв 30 с» (mirrors the codex TUI's "Worked for …" line; en: "Worked for 12m 30s"). It appears only when the turn is finished (agent idle/exited), not after every message.
2. **On the window/card meta row** — while RUNNING, the existing "N с тому" recency slot gains the live elapsed: «працює 4 хв»; when idle, it shows «останній прогін: 12 хв» in the Hint tooltip of the recency chip (keep the visible row uncluttered — tooltip only for idle duration; the feed separator is the primary idle surface).

**Data:** transcripts carry per-message timestamps for both engines (claude jsonl entries; codex rollout events). Derive turn boundaries in the scanner/describe layer and expose `lastTurn: {startedAt, endedAt|null}` in the card payload. No engine/runtime changes. Formatting: uk/en humanized (с/хв/год), tabular-nums.

Scope: src/lib/scanner derivation + window/card header UI + feed separator. Do NOT touch src/lib/{flows,agent,runtime}. Tokens, en+uk, mobile intact.

Latand:
Shipped in PR #234: lastTurn derivation in scanner, live «працює N» while running, «Працював N хв M с» feed separator when idle, Hint idle duration.

PR #1618, opened 2026-09-09
fix(accounts,orchestrator): restore manual account choice and a truthful, answerable seat

### LINKED ISSUE #1617: Manual account choice is refused at the launch seam, and the orchestrator dock loses its composer — a seat outside its project's pool cannot be rotated or answered
## What the operator hit

The orchestrator seat for a project was running on a Codex account that project's binding does not list. The operator opened the rotate draft, pressed Confirm, and the dock answered

> **The last designation failed**
> codex account `<account-b>` is not allowed on project `<project>` (allowed codex accounts: `<account-a>`, `<account-c>`, `<account-d>`)

over a still-live incumbent — and the panel offered **no composer at all**: feed, control strip, green `live` badge, nothing to type into.

Two separate defects, one screen.

## 1. A manually chosen account is refused at the launch seam

`resolveHealthySpawnAccount` — the direct-launch seam that `/api/spawn`, the board's launch draft, the orchestrator's create and rotate drafts, and `spawn_agent` all reach — checked a NAMED account against the project's pool and threw `ProjectAccountRefusedError` when the pool did not contain it.

#1279's own rule says this is wrong. `explicitAccountChoice` in `src/lib/accounts/projectBindings.ts` already states that the binding constrains what the Viewer picks **by itself**, and that a choice which names an account is a capability the record never vetoes. The two switch seams (`structured-reconfigure`, `conversation-switch`) honour that. The launch seam never reached it.

**Why it is unrecoverable rather than merely annoying:** the rotate draft prefills the **incumbent's** account. A seat already running outside its project's pool therefore cannot be rotated at all — the default offer is the one the fence refuses, and choosing any other unbound account is refused on the same grounds.

Evidence, from the live seat read (`GET /api/orchestrator/seat`, `.../seat/status`):

| field | value |
| --- | --- |
| active seat epoch | 156 |
| active seat account | `<account-b>` — **outside the project's pool** |
| host state / lifecycle | `alive` / `waiting` |
| pending seat epoch | 157 |
| pending intent error | ``codex account <account-b> is not allowed on project <project> (allowed codex accounts: …)`` |

The seat was already running on the very account the next designation was refused for.

### What must not change

Only autonomous selection and fallback stay inside the pool. A pick nobody named still draws from the pool only, still reports an exhausted pool, and still refuses on a binding record this process cannot read. A named account is still never capacity-checked and never substituted, and authentication is still decided by the engine's own health pass.

There is a related hole on the same line: for Claude, the named account was filtered out of the health pass's candidate set, so the pass reported "the request named nobody I have" and the launch silently landed on the automatic account instead.

## 2. The hoisted composer is deleted by the host's own stale render snapshot

Independent of the account refusal, and the reason the seat looked like a status display.

`VoiceComposerHost` (the Viewer-level owner of every conversation card's composer, #691) retains a card's props past the card's unmount and prunes them once it renders no composer for that conversation. The prune read `conversationIds` from the enclosing **render**:

```ts
const conversationIds = new Set(getVoiceComposerCardIds());   // render
useEffect(() => {
  for (const cardId of getVoiceComposerCardPropsIds()) {
    if (!conversationIds.has(cardId)) forgetVoiceComposerCardProps(cardId);
  }
});
```

A card publishes its **place** from a ref, during the commit that notifies the host, and its **props** from its own effect, which runs before the host's. The host's first prune therefore runs with a set taken before the card existed, finds the props that card has just published, and deletes them — together with the per-place map, which nothing republishes until the card's own props change again.

Reproduced in the live browser (read-only CDP against the running Viewer, dock open on the project):

```
{"conv":true,
 "intentErr":"The last designation failed\ncodex account <account-b> is not allowed on project <project> …",
 "slots":1,       // the dock published its composer place
 "textareas":[]}  // and no composer rendered into it, anywhere on the page
```

and deterministically, mounting the real `Viewer` in happy-dom: `places: 1`, `inDock: false`, `parked: false`, with the slots registry showing `node: true, props: false`.

**Why it shipped invisibly:** `OrchestratorPanel.dom.test.tsx` mounts the *panel*, and a tree with no `VoiceComposerHost` in it keeps the inline card-scoped composer by design (`TmuxComposer`'s dispatcher). So its assertion — "an active seat mounts the REAL conversation column, feed and composer" — passes on a code path production never takes.

## 3. "live" over an idle agent

Smaller, and the operator named it: the badge could not distinguish an idle process from an actively working turn, because the catalog cannot see the difference. The status read already carries `liveness.lifecycle`, and the shared lifecycle vocabulary defines `waiting` as "a host is alive and idle". A green `live` over an agent that is only awaiting input reads as "it is working".

## Proposed fix

1. `selectProjectAccount` gains `requestedChoice: "explicit" | "fenced"` (default `fenced`, i.e. every existing caller unchanged). The direct-launch seam passes `explicit`; the binding no longer vetoes it, a damaged record no longer vetoes it, and the named account goes through the health pass rather than around it. The crossing is appended to the override journal the project view already renders beside the pool, under a new `launch` gesture.
2. The host's prune re-reads the registry at effect time instead of using the render snapshot.
3. `seatLivenessOf` reports `waiting` for a hosted seat whose lifecycle says so — applied only to `live`, and only to an affirmed reading of the seat's own conversation.

Account labels and the project key above are placeholders.

Latand:
Closes #1617.

The operator's screenshot: a live orchestrator seat, `The last designation failed — codex account <account-b> is not allowed on project <project>`, and no composer anywhere on the panel. Three defects behind it, fixed here.

## 1. An explicitly named account is a control the binding may not veto

`resolveHealthySpawnAccount` — the direct-launch seam behind `/api/spawn`, the board's launch draft, the orchestrator's create and rotate drafts and `spawn_agent` — refused a NAMED account the project's pool did not contain.

#1279's own `explicitAccountChoice` already says the binding constrains what the Viewer picks **by itself** and never vetoes a choice that names an account. The two switch seams honoured that; the launch seam never reached it. What that produced: the rotate draft prefills the **incumbent's** account, so a seat already running outside its pool could not be rotated at all.

`selectProjectAccount` now takes `requestedChoice`. `fenced` is the default and every pre-existing caller — a pipeline stage's recorded account, a round's own — where nobody worked a control. `explicit` is the direct-launch seam.

Unchanged, and asserted beside every relaxation:

- the pick nobody named still draws from the pool only, still reports an exhausted pool, and still refuses on an unreadable record;
- a named account is still never capacity-checked and never substituted;
- authentication still decides — and a named Claude account now goes **through** the health pass instead of being filtered out of its candidate set and silently replaced by the automatic account;
- a damaged record no longer vetoes a named choice, and with the pool invisible there is no automatic answer left to fall back to either.

The binding stops being a veto, so it becomes a record: an out-of-pool launch is appended to the override journal the project view already renders beside the pool, under a new `launch` gesture, once the receipt exists and only for the account the request actually named.

## 2. The hoisted composer, deleted by the host's own stale render snapshot

`VoiceComposerHost` prunes retained card props using `conversationIds` read from the enclosing **render**. A card publishes its place from a ref during the commit that notifies the host, and its props from its own effect, which runs first — so the host's first prune carries a set taken before the card existed, finds the props just published, and deletes them along with the per-place map nothing republishes until the card's props change again. Re-read at effect time, the prune sees what its render's commit produced.

## 3. `waiting` instead of `live` over an idle agent

A badge that says `live` over an agent which is only awaiting input reads as "it is working". Two evidence sources can say which it is, on very different clocks, and the order matters:

- **`lastTurn`, the transcript's own turn boundary** (`endedAt: null` while the turn runs), which arrives on the file poll. This leads, with no affirmation gate — it is that poll's own answer about that transcript, not a claim held over.
- **`liveness.lifecycle`**, the shared vocabulary's word, which rides the incumbent poll — a cadence set by context-window wear, one minute, and refreshed by nothing when the operator sends a message. It is the FALLBACK, for a tail that carries no boundary at all (#231), and only an affirmed reading may downgrade there (`hostLive`, the rule from #1182).

Leading with `lifecycle` was wrong in both directions: `waiting` held over an agent that started working a message ago, and `live` held over one that finished — the original complaint, time-boxed to a minute rather than fixed. `seatState.test.ts` asserts both crossings against a stale reading.

Applied to `live` only: `stalled`, `resumable` and `dead` are stronger statements about the HOST, and an idle turn does not soften any of them. The phone carries the same word from the same state machine, and a `waiting` seat whose conversation is on the device keeps speaking that conversation's own phrase — the duration included.

## Evidence

**Live state, before any change** (`GET /api/orchestrator/seat`, `.../seat/status`, `conversation_deliverability`): active seat epoch 156 on the out-of-pool account, `hostState: alive`, `lifecycle: waiting`, `deliverable: true`; pending seat epoch 157 carrying the refusal on its intent. The seat was already running on the very account the next designation was refused for, and the conversation it refused to leave was answerable — the panel simply offered nothing to answer it with.

**Live browser** (read-only CDP against the running Viewer, dock open): `slots: 1` and `textareas: []` — the dock published its composer place and no composer rendered into it anywhere on the page.

**Red → green**, each verified by restoring the pre-change source under the new tests:

| test | red before |
| --- | --- |
| `rotationAccountChoice.test.ts` — the operator rotates onto an unbound authenticated account | 409, the production shape |
| `managerProjectBinding.test.ts` — explicit choice outside the pool launches; automatic still refuses it; damaged record splits the two; Claude's named account reaches the health pass | 4 of 5 red |
| `route.binding.test.ts` — the crossing is recorded, and an in-pool launch records nothing | red |
| `seatComposerHoist.dom.test.tsx` — a seated hosted orchestrator has a composer in the dock, refused designation included | both red |
| `seatState.test.ts` — idle turn reads `waiting`, running reads `live`, only an affirmed reading may downgrade | red |

`rotationAccountChoice.test.ts` drives the real `POST /api/orchestrator/rotate` route over a loopback listener, through the real seat command and the real account decision reading a real binding record; only `/api/spawn` stands in, resolving the account with the same call the route makes and then reporting an accepted launch instead of starting a process. `seatComposerHoist.dom.test.tsx` mounts the real `Viewer`, so the hoisted composer path — the only one production takes — is the one under test; the existing panel suite asserts a composer in a tree with no host in it, which is why this shipped invisibly.

Two guard tests pass on both sides by design and are marked as such: the autonomous pick behind the same rotation still draws from the pool, and a refused rotation leaves the incumbent seated with the reason on the record and no successor.

## Checks run

- `bun scripts/privacy-publication-gate.ts --base d14f767` — PASS.
- Suites, by path: `accounts/{managerProjectBinding,projectSelection,accountOverrides,projectAccountsView,reseatBinding,reseatCommandBinding}`, `pipelines/stageAccountBinding`, `api/spawn/{route,route.binding}`, `agent/spawnCommand`, `agent/registry.migrationBinding`, `orchestrator/{rotationAccountChoice,rotationAuthority,seatCommand}`, `components/orchestrator/*`, `components/voice/*`, `components/TmuxComposer.dom`, `components/mobile/orchestratorRowState`, `lib/i18n` — all green.
- Sweeping `src/components/orchestrator/` reports 3 failures from `mock.module` leaking between `dockOpenState.dom.test.tsx` and `OrchestratorPanel.dom.test.tsx`. Those 3 fail identically at `d14f767` with this branch's new files removed, and `OrchestratorPanel.dom.test.tsx` passes on its own here. Pre-existing, not introduced.

## Ownership note

`src/components/voice/VoiceComposerHost.tsx` is Viewer-level composer machinery rather than orchestrator code. It is touched because it is the mechanism behind the missing composer this lane owns; the change is four lines and adds no behaviour beyond releasing only what has really gone.

## Deliberately not changed

`resolveProjectSpawn`'s `requestedId` — a pipeline stage's `account:`, a round's recorded account — keeps today's refusal. Those are forwarded ids on paths this directive did not name, and a pin there has its own history (#1371).

### DIFF
--- a/src/app/api/spawn/route.binding.test.ts
+++ b/src/app/api/spawn/route.binding.test.ts
@@ -71,6 +71,9 @@ async function spawn(
   /** The account the REQUEST names, when it names one — the board's launch
       draft, the orchestrator's rotate draft, `spawn_agent`. */
   requestedAccountId?: string,
+  /** A registry shared across calls, so a second request with the same
+      `clientAttemptId` is a REPLAY rather than a fresh launch. */
+  sharedStore?: InstanceType<typeof AgentRegistry>,
 ): Promise<{
   status: number;
   error: string;
@@ -82,7 +85,7 @@ async function spawn(
   projects: (string | null | undefined)[];
   receipts: number;
 }> {
-  const store = new AgentRegistry(path.join(SANDBOX, `${clientAttemptId}.json`));
+  const store = sharedStore ?? new AgentRegistry(path.join(SANDBOX, `${clientAttemptId}.json`));
   let accountResolutions = 0;
   /* Recorded inside the stub, because the route's own answer cannot show what
      it asked for — and what it asked for IS the fence at this seam. */
@@ -359,4 +362,37 @@ test("a launch onto an account the pool already contains records nothing", async
   expect(attempt.status).not.toBe(409);
   const { accountProjectOverrides } = await import("@/lib/accounts/accountOverrides");
   expect(accountProjectOverrides({ project })).toEqual([]);
+});
+
+test("a second request under the same attempt id appends no second crossing", async () => {
+  /* An idempotent retry — a lost response resent under the same
+     `clientAttemptId` — is the same launch arriving twice, not a second choice.
+     The journal is capped at 200 and drops its oldest entries, so a duplicate
+     evicts a crossing this record exists to keep.
+
+     WHAT THIS FIXTURE REACHES, stated because it is less than the guard covers:
+     the first launch here reserves its receipt (which is where the crossing is
+     recorded) and then dies further in on a stub the harness does not carry, so
+     the retry lands on the terminal-pinned-failure branch and answers 409. That
+     branch returns before the recording either way. The `begun.kind ===
+     "created"` gate in `executeSpawnRequest` is what covers the REPLAY branch,
+     which needs a first launch that settles — nothing in this file can produce
+     one. So this asserts the property, not the branch. */
+  const cwd = fs.mkdtempSync(path.join(SANDBOX, "named-replayed-"));
+  const project = projectForCwd(cwd)!;
+  fs.writeFileSync(RECORD, JSON.stringify({
+    schemaVersion: 1,
+    bindings: [{ engine: "claude", accountId: "acct-reserved", project, createdAt: "2026-09-10T00:00:00.000Z" }],
+  }), "utf8");
+  const store = new AgentRegistry(path.join(SANDBOX, "replayed-registry.json"));
+
+  const first = await spawn(cwd, "binding_named_replay_20260910", undefined, "acct-chosen-by-hand", store);
+  const second = await spawn(cwd, "binding_named_replay_20260910", undefined, "acct-chosen-by-hand", store);
+
+  /* One receipt across both requests: the second was the same launch, not a
+     new one, whichever branch answered it. */
+  expect(first.receipts).toBe(1);
+  expect(second.receipts).toBe(1);
+  const { accountProjectOverrides } = await import("@/lib/accounts/accountOverrides");
+  expect(accountProjectOverrides({ project })).toHaveLength(1);
 });
--- a/src/app/api/spawn/route.binding.test.ts
+++ b/src/app/api/spawn/route.binding.test.ts
@@ -74,6 +74,12 @@ async function spawn(
   /** A registry shared across calls, so a second request with the same
       `clientAttemptId` is a REPLAY rather than a fresh launch. */
   sharedStore?: InstanceType<typeof AgentRegistry>,
+  /** Report the named account as rate-limited until this instant, which is how
+      the route reaches its QUEUED answer: a 202 carrying a real
+      `SpawnResponse`, with no process started anywhere. It is the only success
+      shape this suite can produce — every other one ends in the tmux launcher,
+      which reads the process-global registry rather than the injected one. */
+  admissionRetryAt?: string,
 ): Promise<{
   status: number;
   error: string;
@@ -84,6 +90,8 @@ async function spawn(
       route contributes to the fence. */
   projects: (string | null | undefined)[];
   receipts: number;
+  /** The out-of-pool notice on the route's OWN answer, when it carried one. */
+  accountOverride: { recorded?: boolean; recordFailure?: string; accountId?: string; reason?: string } | null;
 }> {
   const store = sharedStore ?? new AgentRegistry(path.join(SANDBOX, `${clientAttemptId}.json`));
   let accountResolutions = 0;
@@ -106,6 +114,9 @@ async function spawn(
        nothing downstream of it — the attribution below included — can be
        observed at all. */
     resolveSpawnAccount: (_engine: unknown, accountId: string) => accountContext(accountId),
+    /* The queued answer stores the launch's images before it parks the pin;
+       this suite sends none. */
+    storeImages: () => [],
     resolveHealthySpawnAccount: async (
       _engine: unknown,
       requested: unknown,
@@ -117,7 +128,10 @@ async function spawn(
       /* The seam resolves an explicitly named account TO ITSELF, pool or no
          pool — that is the rule this stub stands in for, and it is what the
          route's attribution has to see to record the crossing. */
-      return accountContext(typeof requested === "string" && requested ? requested : "acct-default");
+      const context = accountContext(typeof requested === "string" && requested ? requested : "acct-default");
+      return admissionRetryAt
+        ? { ...context, requestedAdmission: { kind: "retry-at", retryAt: admissionRetryAt } }
+        : context;
     },
     defer: (work: () => unknown) => { void work(); },
   } as unknown as SpawnRouteDependencies;
@@ -134,13 +148,17 @@ async function spawn(
       ...(requestedAccountId ? { accountId: requestedAccountId } : {}),
     }),
   }), dependencies);
-  const payload = await response.json() as { error?: string };
+  const payload = await response.json() as {
+    error?: string;
+    accountOverride?: { recorded?: boolean; recordFailure?: string; accountId?: string; reason?: string };
+  };
   return {
     status: response.status,
     error: payload.error ?? "",
     accountResolutions,
     projects,
     receipts: Object.keys(store.snapshot().receipts).length,
+    accountOverride: payload.accountOverride ?? null,
   };
 }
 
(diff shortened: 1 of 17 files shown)


Latand:
Follow-up on the pre-existing red noted in the PR body: filed as #1619 — `dockOpenState.dom.test.tsx` and `OrchestratorPanel.dom.test.tsx` poison each other through the shared `@/hooks/runtimeBus` mock, which is why `bun test src/components/orchestrator/` is red on `main`. Verified at `d14f767` with this branch's new files removed: the same three failures. The new `seatComposerHoist.dom.test.tsx` mounts the real Viewer for the same reason `dockOpenState` does, so it joins that pair rather than creating a new problem — it passes on its own, and each file in this PR was run by path.

https://github.com/Latand/live-log-viewer-next/issues/1619

Latand:
Round 1 addressed at `7e7257cb7e55c98d593e298499b2cfdcdbbc761c` — all 8 checks pass.

**1 FIXED (the real one).** `seatLivenessOf` now leads with the catalog's `lastTurn` boundary (`endedAt: null` = running), which arrives on the file poll, and falls back to `liveness.lifecycle` only for a tail that carries no boundary (#231) — where the `hostLive` affirmation still gates the downgrade. You were right that it was wrong in both directions, and the new test asserts both: a turn opened after the last incumbent read shows `live`, a turn ended after it shows `waiting`.

**2 FIXED.** `waiting` joins `live` in the conversation-phrase branch of `seatCardView`. It is a hosted seat, the conversation vocabulary already has its own `waiting`, and excluding it was exactly the "one seat reading two ways" the invariant forbids. Covered.

**3 FIXED.** Gated on `begun.kind === "created"`. The test is a guard rather than a red-proven one and says so in its own comment: the first launch in that fixture dies further in on a stub the harness does not carry, so the retry lands on the terminal-pinned-failure branch (which returns before the recording anyway) and the `replay` branch is not reachable there. The property is asserted; the branch is not.

**4 FIXED.** The Claude candidate set is the named account alone when `recordUnreadable`. You are right that `allowed === null` reads identically for "unbound" and "unreadable", which is what let the degrade through — the Codex twin was there and the Claude one was not.

**5 FIXED — dropped.** `outsidePool` is gone from `ProjectAccountSelection`. `attributeNamedAccountChoice` is one classifier for every explicit seam; a second copy on the return value is only a second place for them to disagree.

**6 FIXED — reframed.** The test now names what it drives and what it does not, and points at `route.binding.test.ts` for the call site.

Findings 1, 2 and 4 were red-verified against the previous commit (3, 1 and 1 failures respectively) before the fix landed.

Latand:
Round 2 addressed at `83a1384af6a5b4f7e13215e18150b12dc497483e` — all 8 checks pass. Four fixed, one rejected with reasoning.

**1 FIXED.** Only `resolveHealthySpawnAccount`'s named branch left the automatic list; the doc now says that and names `resolveProjectSpawn`'s four callers explicitly as still fenced. You're right that this file's whole purpose is to be the reviewed artifact — a line implying a pipeline stage's named account crosses the pool is the defect it exists to prevent.

**2 FIXED.** `SpawnResponse` carries `accountOverride`, and every answer past the reservation carries it. Two tests: the notice on a recorded crossing, and `recorded: false` with its `recordFailure` when a directory sits where the journal's file belongs. Both drive the route's QUEUED 202 — the only success shape that suite can reach without entering the tmux launcher, which reads the process-global registry rather than the injected one.

**3 FIXED.** `recordUnreadable` is the error itself now instead of a boolean, and the empty-candidate-set branch throws it. Test asserts the Claude answer names `account-project-bindings.json` and does *not* say "Re-login", and that Codex still names the account.

**4 REJECTED** — the codebase already decided this, and it decided the other way. `accountOverrides.ts`: "A person **or an agent** naming an account is exercising a control, and a control is a capability"; `AccountChoiceActor` has an `agent` kind with the calling conversation for exactly this; `rotationActor` bans nobody on the same reasoning. Adding a fence for agent callers now would be a new refusal nobody asked for, and the crossing is already attributed with `actor: agent` and the conversation id, which is what makes it reviewable. If the operator wants that fence, `authenticatedCaller` is in hand at the attribution site and it is one condition — but it should be their call stated as such, not mine inferred from "manually".

**5 FIXED.** PR body §3 rewritten to the shipped rule: `lastTurn` leads with no `hostLive` gate, `lifecycle` is the fallback for a tail with no boundary, and affirmation guards only that fallback.

Findings 2 and 3 were red-verified against `7e7257cb` (2 and 1 failures) before the fix.

Unrelated, found while sweeping every `SpawnResponse` consumer: `route.admission.test.ts` and `tasks/route.test.ts` are red on `main` and fail **in isolation**, so they are not the #1619 leakage. Baselined at `d14f767` with my files removed — identical failures. Filed as #1622; I have not diagnosed either.

https://github.com/Latand/live-log-viewer-next/issues/1622

PR #1626, opened 2026-09-10
fix(board): a catalog open keeps its promise until the conversation is on screen (#1625)

### LINKED ISSUE #1625: fix(board): catalog opens can leave completed pipeline agents outside the viewport
After #1620, the desktop catalog loads and retains pages correctly, but opening a completed pipeline conversation can leave its card outside the viewport. This prevents the operator from finding and reading the selected agent.

Observed on production `e552cd6e7cc708391289a3af9030f188709fb5e3`: opening two different completed pipeline agents changes the conversation hash and focused path correctly. The requested file is hydrated, admitted as an isolated ephemeral manual node, and present in both authored and projected layout indexes. The projected node remains outside the visible viewport. Reopening one target leaves camera `{x:-8,y:-6438.3056,z:0.58}` while its projected rectangle is `{x:68.9655,y:9336.8966,w:1034.4828,h:1172.4138}`; its task band is entirely above the viewport. The requested node has no rendered viewport element.

Reproduce with a completed pipeline conversation opened from the desktop catalog, many preceding task bands, and restored camera state. Capture camera and final target geometry across materialization/measurement. Determine whether the focus is consumed against provisional geometry or displaced by subsequent reflow; this is a hypothesis, not a confirmed cause.

Acceptance: an actual catalog row click brings the requested conversation into view and makes its content readable, including first open and repeated open of folded/completed pipeline history. Keep navigation one-shot; a later manual pan must remain respected. Preserve membership, task flags/history, retained catalog pages, valid camera restoration and the empty-board migration behavior. Add causal red/green coverage with the real composition and focused browser evidence; avoid a timeout increase or permanent camera follow.

Private screenshots and trace are retained by the owning repair session. No data deletion or agent interruption is part of this fix.

Latand:
Closes #1625.

## What the operator saw

Opening a historical agent from the desktop catalog changed the conversation hash, hydrated the transcript, admitted it as an isolated ephemeral manual node, and placed it in both the authored and the projected layout — and then left it outside the viewport, where it could not be read. Reopening one target left the camera at `{x:-8, y:-6438.3056, z:0.58}` while the requested node's projected rectangle was `{x:68.9655, y:9336.8966, w:1034.4828, h:1172.4138}`; a first open landed on the plain `{0, 0, 0.58}` framing with an unrelated band on screen.

## Cause, measured

The band projection is built for a camera it is built **before**. `SchemeBoard` has to hand `layoutTaskBands` a zoom and a viewport width, and because the layout is an input to `useSchemeCamera` rather than an output of it, those can only be copies of the camera's own values applied one commit late (initial `0.5` zoom, `1400px` width).

A catalog row click is precisely the case where that lag matters: the list unmounts and the board **mounts with the request already standing**. The requested node's first rectangle is therefore the provisional one, and on a deep stack of task bands it is thousands of world pixels from where the node settles. The aim was taken against it once and the request marked handled, so nothing remained to answer the relayout — and the saved-camera restore and the off-world re-fit then ran and framed something else.

Driving the reported board composition reproduces it deterministically: the requested node settles at world `y 9336.8966` while the camera stops **2601px short of it**, and a freshly loaded tab lands on `translate(0px, 0px) scale(0.58)` with **no node drawn at all** for the requested path. Both of those readings are taken from the running production build, not from a model of it.

## Fix

A focus request becomes an obligation rather than an event:

- the camera holds it until the node it names is actually **framed** — its head inside the viewport, with enough of it to read;
- it re-aims only while the node is not framed, and only when the node's rectangle or the camera has moved since the last aim, so it answers a settling board and never chases a settled one;
- it is **discharged** the moment the node is shown — a later reflow moves the node freely and the camera stays put;
- it is **dropped** by `queueCam` (every pan, wheel and pinch) and by `glideTo` (every explicit framing), so the operator always wins;
- neither the highlight expiring nor a scanner poll can re-arm it. Only a new focus value can, and upstream only a new nonce produces one — which is what keeps this from being the standing follow `focusRequestEdge` removed.

No timeout was changed, and no permanent camera follow was added. The saved-camera restore, the off-world recovery, task flags/history/associations and catalog page retention are untouched.

## Coverage

**Mounted board, the composition that failed** (`SchemeBoard.catalogFocus.dom.test.tsx`) — a completed pipeline folded into a band far down a stack of 15, the requested conversation admitted from the catalog:

| case | before | after |
| --- | --- | --- |
| board mounts with the request standing | fail | pass |
| repeat open after the operator panned away | fail | pass |
| pan respected through a later band reflow | fail | pass |
| no request → the saved camera is restored | pass | pass |

**Pure framing questions** (`useSchemeCamera.test.ts`): `centredCamera` frames what it was computed from; the reported rectangle under the reported camera is *not* framed; a pane taller than the viewport is judged by its head; an unmeasured viewport frames nothing, so a request stays owed.

**Rendered acceptance against the production build** (`scripts/capture-issue-1625-catalog-focus.ts`), in a real browser, on a synthetic home with its own `HOME`/XDG dirs/`TMPDIR`/state dir/provider homes. It seeds 260 conversations, 16 tasks and a finished two-stage pipeline, then pages the catalog to the archived row and clicks it — once from a board already on screen, once from a freshly loaded tab.

| reading | with the defect | with the fix |
| --- | --- | --- |
| open from a mounted board | shown, top 134.7 | shown, top 134.7 |
| open from a fresh tab (the catalog remount) | **no node at all**, camera `translate(0px, 0px)` | shown, top 134.7, own content rendered, native presentation |
| camera after a manual pan + 25s (2+ polls) | — | unchanged (`translate(-8px, -8704px)`) |
| red self-check: node pushed off screen | flagged | flagged |
| red self-check: node removed outright | flagged | flagged |

The board viewport in that run is `40..934`, so `top 134.7` is the requested conversation sitting near the top of the canvas with its title and body rendered inside it.

The suite is not run wholesale here: the broad pre-existing failures on this branch's merge base were re-checked file by file at `e552cd6e` and are base-identical (`SchemeBoard.lineage`, `SchemeBoard.pipelineRegions`, `SchemeBoard.pipelineComposition`, `SchemeBoard.stageOverlay`, `ProjectDashboard.selection`), unchanged by this diff.

### DIFF
--- a/scripts/capture-issue-1625-catalog-focus.ts
+++ b/scripts/capture-issue-1625-catalog-focus.ts
@@ -1,0 +1,561 @@
+/**
+ * Rendered acceptance for #1625, in a real browser against the production
+ * build:
+ *
+ *   bun run build && bun scripts/capture-issue-1625-catalog-focus.ts
+ *
+ * The question is the operator's own: after clicking a row in the desktop agent
+ * catalog, is the conversation they asked for ON SCREEN and readable? The
+ * earlier rendered acceptance (#1614) opened whichever row the list happened to
+ * draw first — a recent agent in the top band, under no layout pressure — and
+ * asked only whether a node for it existed anywhere in the DOM. Both of those
+ * pass while the reported defect is live, so this capture opens the shape that
+ * actually failed: a HISTORICAL conversation, outside the board's own window,
+ * belonging to a COMPLETED pipeline, whose task band sits far down a long stack
+ * of preceding bands.
+ *
+ * It measures, in the page, at the real production build:
+ *
+ *   - the first open: the requested node's rectangle against the board
+ *     viewport, and whether its own title is rendered inside it;
+ *   - the repeat open, after the operator has gone back to the list;
+ *   - that a manual pan afterwards is RESPECTED — the camera is read again
+ *     after more than one scanner poll, and must not have crept back.
+ *
+ * Then it proves the audit can go red: the same reading is taken against a page
+ * where the defect has been reintroduced by hand (the world pushed so the node
+ * leaves the viewport, and the node removed outright). A run where either of
+ * those still reads as "shown" exits non-zero.
+ *
+ * Everything is served against a purpose-built synthetic home under the temp
+ * root — its own HOME, XDG dirs, TMPDIR, viewer state dir and provider homes —
+ * so nothing here reads or writes the operator's live state, and no real path
+ * appears in any frame. Nothing is deployed.
+ *
+ * Frames and the measurement JSON land outside the repository, under
+ * <BOARD_CAPTURE_DIR>/<unique-run>/out.
+ */
+import { spawn, type ChildProcess } from "node:child_process";
+import fs from "node:fs";
+import path from "node:path";
+
+import { chromium, type Browser, type Page } from "playwright-core";
+
+import { createCaptureDirectory } from "./capture-directory";
+
+const repoRoot = path.resolve(import.meta.dir, "..");
+const BASE = createCaptureDirectory({
+  envName: "BOARD_CAPTURE_DIR",
+  prefix: "llv-issue-1625",
+  raw: process.env.BOARD_CAPTURE_DIR,
+  repoRoot,
+});
+const HOME = path.join(BASE, "home");
+const OUT_DIR = path.join(BASE, "out");
+const STATE_DIR = path.join(HOME, ".config", "agent-log-viewer", "state");
+
+/** Conversations on disk. The board draws a bounded window of the freshest. */
+const CONVERSATIONS = 260;
+/** Task bands. The requested conversation's band must have many above it. */
+const TASKS = 15;
+/** Members each staffed band holds, so the stack is genuinely tall. */
+const PER_BAND = 5;
+const PROJECT_NAME = "catalog";
+/** The row this capture opens, named so it can be found in the catalog. */
+const TARGET_TITLE = "Folded pipeline builder from the archive";
+
+const projectSlug = (cwd: string) => cwd.replace(/[^A-Za-z0-9]/g, "-");
+const line = (record: Record<string, unknown>) => JSON.stringify(record) + "\n";
+
+function conversationFile(index: number): { file: string; uuid: string } {
+  const cwd = path.join(HOME, "Projects", PROJECT_NAME);
+  const folder = path.join(HOME, ".claude/projects", projectSlug(cwd));
+  fs.mkdirSync(folder, { recursive: true });
+  const uuid = `${String(index + 1).padStart(8, "0")}-3333-4333-8333-333333333333`;
+  return { file: path.join(folder, `${uuid}.jsonl`), uuid };
+}
+
+interface Seeded { file: string; uuid: string; title: string }
+
+/**
+ * The corpus. The freshest {@link TASKS} × {@link PER_BAND} conversations are
+ * what the board's window draws, so their task bands hold real members and the
+ * stack is tall. The LAST conversation written is deliberately the oldest one
+ * on disk: it is outside that window, which is what makes opening it from the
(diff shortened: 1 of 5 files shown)


Latand:
Rendered acceptance, re-run against the production build at this PR's exact head `020dd988b56b5ca262f3fa5a2a14798df521af0f`:

```
bands: 21   corpus: 260 conversations, 16 tasks, board window 80 cards
board viewport: top 40, height 894

open from a mounted board   shown, top 134.7, own content rendered, native   camera translate(0px, -6713.28px) scale(0.58)
open from a fresh tab       shown, top 134.7, own content rendered, native   camera translate(-8px, -6713.28px) scale(0.58)
after a manual pan          translate(-8px, -8704px) scale(0.58)
after 25s (2+ scanner polls) translate(-8px, -8704px) scale(0.58)   unchanged
red self-check: node pushed off screen   flagged
red self-check: node removed outright    flagged
failures: []
```

The same script run against a build carrying the old one-shot aim reports the reported failure verbatim — the fresh-tab open leaves the camera at `translate(0px, 0px) scale(0.58)` with **no node drawn at all** for the requested path:

```
#1625 acceptance FAILED:
  the repeat catalog open left the requested conversation off screen (top 0)
  the repeat open put the node on screen without its content
```

Reproduce with `bun run build && bun scripts/capture-issue-1625-catalog-focus.ts`. Everything it touches lives under its own synthetic `HOME`/XDG dirs/`TMPDIR`/state dir/provider homes; nothing is deployed and no live state is read or written.

Latand:
Correction pushed for the confirmed regression: `nodeIsFramed` was answering the completion question **before** the first move, so a request whose node was already somewhere in the viewport was answered without moving at all. On an overview-scale board that dropped the pre-existing `centerOn(node, 0.55)` guarantee and left the conversation a chip.

Two fixes, both to the same confusion between *located* and *readable*:

1. **The first aim of a request is unconditional.** The move is made, and only then does "already shown" complete the request. Bounded re-aim, discharge, and operator takeover through `queueCam`/`glideTo` are unchanged.
2. **`nodeIsFramed` no longer accepts what it should not** — a camera below the readable scale (`READABLE_Z = 0.55`, the same floor `centerOn` has always applied), or a node with a sliver inside an edge. It now requires the readable scale and a real part of the node's width.

Root's independent hook probe `camera-focus-probe/readable-zoom.test.tsx`, unmodified:

```
before  ROOT_FOCUS_ZOOM          {"z":0.55,"x":480,"y":39}
        ROOT_DELAYED_FOCUS_ZOOM  {"x":0,"y":0,"z":0.2}      1 pass / 1 fail
after   ROOT_FOCUS_ZOOM          {"z":0.55,"x":480,"y":39}
        ROOT_DELAYED_FOCUS_ZOOM  {"z":0.55,"x":480,"y":39}  2 pass / 0 fail
```

Both readings now match the `e552cd6e` baseline exactly, to the pixel.

Repo-side regression coverage is a real-hook test at both timings (`useSchemeCamera.focus.dom.test.tsx`: mount with the request standing, and a request arriving after the viewport is measured on a saved `0.2` board), plus unit tests for the two readings the predicate got wrong. Behaviour-only revert of this correction turns the new hook test red (`2 pass / 1 fail`) at the same case root's probe flags, so the coverage is causal rather than incidental.

Nothing else moved. The rendered acceptance re-ran at this head and is identical to the previous run:

```
open from a mounted board    shown, top 134.7, own content, native   translate(0px, -6713.28px) scale(0.58)
open from a fresh tab        shown, top 134.7, own content, native   translate(-8px, -6713.28px) scale(0.58)
after a manual pan           translate(-8px, -8704px) scale(0.58)
after 25s (2+ scanner polls) translate(-8px, -8704px) scale(0.58)    unchanged
red self-checks              both flagged
failures: []
```

The operator's captured board input also still frames the requested node at screen top `109.6`, world `y 9336.8966`, 16 bands.

Issue #1622, opened 2026-09-10
Two test files are red on main and fail in isolation: spawn route admission (#341 silent root) and the tasks route GET


Latand:
Two test files are red on `main` and each fails when run alone, so this is not the `mock.module` cross-file leakage in #1619 — these are genuinely broken.

At `d14f767` (current `main`), with no branch changes in the tree:

```
bun test src/app/api/spawn/route.admission.test.ts
(fail) an operator capability caller without src proceeds as a silent root (#341)
 7 pass  1 fail
```

```
bun test src/app/api/tasks/route.test.ts
(fail) GET derives pipelineIds including closed history and filters stale task ids
 0 pass  1 fail
```

The second file has exactly one test, so that suite is entirely red.

## Why it survives

CI does not run the suite, and the working convention is to run touched files by path — neither file is touched by most work, so nobody arrives at them. I hit both only because a change to `SpawnResponse` made me sweep every consumer of the spawn route, then baseline the result against `main` before blaming my own branch.

## What is worth deciding

Either the two tests describe behaviour that changed and should be updated with the change that caused it, or they describe behaviour that regressed. Both readings need somebody who owns those seams; I have not diagnosed either, only established that they are inherited rather than introduced.

Related: #1619 covers a different failure mode in the same blind spot — files that pass alone and poison each other in a directory sweep.

Issue #1614, opened 2026-09-09
Desktop board unusable at scale: every task draws a full-width band, hand-tool controls are dead, and the agent list has no browsable pagination


Latand:
## What the operator sees

At current `main` (`d14f7676`, "task-centered bands"), the desktop board is unusable on a
real board. The reported state, from a live snapshot:

- **~390 tasks, ~680 catalog entries, zero visible agents.** The camera sat at
  `y = -25239`, `zoom = 1.6`. Every recorded task draws a band, so several hundred
  bands with `0 working · 0 conversations` stack vertically and the bands that hold
  agents are scrolled far off screen.
- **Every band spans the whole canvas.** An empty band — a title, a status pill,
  a counts run and two controls — rules a full-width line across the board.
- **Band controls do not answer a pointer** in the hand tool, at either scale.
- **No desktop equivalent of the phone's browsable agent list.** The desktop list
  paginates behind a "load more" button and drops its accumulated pages on update.

A scanner duration of 18 s was observed in the same snapshot. That is a symptom
recorded alongside the report, not a proven cause of the board's state, and this
issue does not assume it is one.

## Mechanisms, read from the code

1. **Every task becomes a band.** `buildTaskBands` (`src/components/scheme/taskBands.ts`)
   opens step 1 with `for (const task of orderedTasks)` over every recorded task and
   calls `makeBand` unconditionally. There is no notion of a task that is not on the
   board, so a board's band count is its lifetime task count.
2. **Band width is the viewport, never the content.** `layoutTaskBands` computes
   `bandW = Math.max(BAND.nativeMinW * s, viewportWidth * s - gutter * 2)` once, before
   the band loop, and assigns that same width to every band's `rect` and `header`.
   Content never narrows it.
3. **Band chrome is disabled on the hand tool.** `SchemeBoard` passes
   `interactive={!mapMode && !handLike && !session}` to `TaskBandsLayer`, and the layer
   turns `interactive === false` into `disabled` plus no `pointer-events-auto` on the
   title, status pill, Details and «+ Agent». `handLike` is true for the hand tool, so
   in that tool none of those controls can be clicked at any zoom.
4. **The desktop list drops its pages.** `ConversationList` calls
   `useConversationCatalog` without `scopeKey`. The hook's effect cleanup runs
   `if (scopeKey === undefined) cache.current.delete(key)`, so each teardown discards
   every page loaded so far and the list restarts at page one. Its page size is the
[...]

PR #1628, opened 2026-09-10
fix(tasks): the per-project limit bounds board bands, not stored history

### LINKED ISSUE #1627: Task admission counts stored history, not board bands: creation refused at 300 rows while the board draws 5
## What happens

`create_task` (MCP) and `POST /api/tasks` refuse with

```
The project already has 300 tasks. Close or delete extra tasks.
```

on a project whose board draws **five** bands. The cap in
`src/lib/tasks/commands.ts` counts every stored row of the project:

```ts
const count = existing.filter((task) => task.project === project).length;
if (count >= TASKS_PER_PROJECT_LIMIT) { … }
```

Since #1614 the board no longer draws a band per stored row: an empty task
whose `board` flag is `hidden` stays in the task list and off the canvas, and
the one-time `emptyTaskBandsHidden` migration set that flag on every legacy
row. The stored list is therefore a **history**, and the board is a small
subset of it — but the admission limit is still counting the history.

Observed on a live board (read through `GET /api/tasks`):

| | count |
|---|---|
| stored task rows, all projects | 910 |
| stored rows in the largest project | 493 |
| rows with `board: "hidden"` | 902 |
| rows the board is asked to draw, all projects | 8 |
| …of them in the largest project | 5 |

So the surface that is actually bounded — the canvas — is at 5 of 300, while
task creation has been refused for the whole project since row 300. The
refusal also asks for the one remedy the operator does not want (deleting
history) to fix a problem history is not causing.

## Why the limit exists at all

To stop the board becoming an unusable vertical stack of bands (#1586, #1614).
That is a **display** bound, not a storage bound.

## Expected

1. The per-project limit counts only tasks that occupy the board, and keeps its
   current numeric value (300).
2. Stored history above the limit is not capped, stays listable, and is never
   auto-deleted as product behaviour.
3. Board membership is defined in one place and enforced consistently at every
   path that can put a band on the canvas — create, and the `board: "shown"`
   restore — with the count taken inside the serialized task-file transaction so
   two concurrent writers cannot both take the last slot.
4. A task that holds a durable agent association draws its band whatever the
   flag says (`update_task` documents this), so that case must be counted
   truthfully wherever the caller can answer it, and the guess must be stated
   out loud where it cannot.
5. Creating a task **off** the board stays possible when the board is full, so
   an agent, a receipt or a worker identity is never lost to a display bound.
   Mandatory launch membership (`ensureTaskMembership`) remains exempt.
6. The refusal text names the real remedy: hide a band, or create the task off
   the board.

## Not in scope

Changing the number 300, deleting rows as product behaviour, or touching the
`emptyTaskBandsHidden` migration. No new migration is needed: on the state
above the change takes the largest project from "refused" to 5 of 300 the
moment it ships.

Latand:
Closes #1627.

## The defect

`create_task` refused every new task on a project whose board draws **five**
bands:

```
The project already has 300 tasks. Close or delete extra tasks.
```

Admission counted stored rows. Since #1614 those rows are a **history**: a task
with nothing on the canvas keeps its place in the task list with
`board: "hidden"`, and the one-time `emptyTaskBandsHidden` migration set that
flag on every legacy row. Read off a live board through `GET /api/tasks`:

| | count |
|---|---|
| stored rows, all projects | 910 |
| stored rows in the largest project | 493 |
| rows flagged `hidden` | 902 |
| rows the board is asked to draw, all projects | 8 |
| …in the largest project | 5 |

So the surface that needed bounding sat at 5 of 300 while creation had been
refused since the project's 300th row — and the refusal asked for the one
remedy that destroys the history it was counting.

**Causal proof before any change**, two cases against unmodified `main`
(`red-probe-main.log`): with 493 hidden rows and zero bands, `createTask`
returns `ok: false`; hiding a band frees nothing.

## The fix

`countBoardTasks(tasks, project, hasMembers)` (in `boardVisibility`, beside the
rule it uses) asks the board's own predicate — `taskShowsOnBoard` — of one
project's rows. There is no second definition of membership: the number a
refusal quotes and the number of bands on screen are the same statement. The
membership answer stays the caller's, exactly as `taskShowsOnBoard` already
required; a caller with no scene to consult passes `() => false` and says out
loud that it is not guessing from assignment rows (the guess this module
already documents as wrong: 265 kept where 11 bands existed).

**Precision on that, since the commit message states it more strongly than it
holds.** Every production call site leaves `hasBoardMembers` unset, so the
effective server rule is `board !== "hidden"` — the tasks the board is *asked*
to draw. A hidden task that still holds a live agent draws a band anyway and is
not counted, so the on-screen total can exceed the cap by however many of those
exist, and restoring exactly such a task at the cap is refused although its band
is already on the canvas. Both are preferred to the alternatives the module
rejects: counting assignment rows would refuse the operator's own board (300 of
their rows carry an assignment to a conversation the board no longer draws), and
counting scan keys kept 265 where 11 bands existed. Those bands come from
mandatory launch membership, which is exempt by construction either way. The
predicate is a parameter precisely so a caller that *can* see the resolved bands
counts them truthfully, and a test pins that direction.

Both admissions answer to the bound:

- **create** — `createTask` counts bands, not rows, and only when the new task
  will occupy one;
- **restore** — `patchTask` applies the same count to `board: "shown"`, because
  a cap on one control that the other can walk around is not a cap. A task
  already occupying a band re-asserting `shown` is never refused, and hiding
  never is.

Each decides inside the serialized read-modify-write of the task file
(`mutateTasks` / `mutateTasksFile`), so two writers cannot take the same last
slot, and a create replayed under its own `clientRequestId` is answered rather
than refused by a board that filled behind it.

Nothing bounds the history. `create_task` and `POST /api/tasks` now take an
optional `board`, so a caller records work off the board when the canvas is
full — no durable identity is ever refused to keep a display small — and
mandatory launch membership (`ensureTaskMembership`) stays exempt as before.
The refusal names that remedy instead of deletion, and the task panel's
«show on board» reports a refusal instead of swallowing it.

`BOARD_TASKS_PER_PROJECT_LIMIT` keeps the number it replaces, 300.

## Migration

None, deliberately. The flag every historical row already carries is what the
count reads, so on the state above the largest project goes from *refused* to
*5 of 300* the moment this ships. Nothing is re-flagged, nothing is deleted,
no row leaves the task list, and no existing membership becomes a phantom band.

The figures in the table and the commit message are a snapshot taken before the
commit and are not reproducible after the fact: the board is live, and a
separately authorized one-off cleanup has since removed 106 completed,
association-free, pipeline-free rows through the API. The same reading now
gives 805 stored rows, 448 in the largest project, behind 7 bands — the
mechanism and the direction are unchanged, and the cleanup was not part of this
change (no deletion path is added here).

## Verification

Bun 1.4.0, run per file (this repo's CI runs a fixed platform list, not these):

| Suite | Result |
| --- | --- |
| `src/lib/tasks/boardTaskLimit.test.ts` (new, 13 cases) | 13 pass / 0 fail |
| `src/app/api/tasks/boardBandLimit.test.ts` (new, 5 cases, real route handlers on an isolated `LLV_STATE_DIR`) | 5 pass / 0 fail |
| `src/lib/mcp/taskBoardVisibility.integration.test.ts` (+1 case, in-memory MCP client) | 4 pass / 0 fail |
| `src/components/tasks/TaskPanel.boardPreference.dom.test.tsx` (+1 case) | 6 pass / 0 fail |
| `src/lib/tasks/tasks.test.ts` | 38 pass / 0 fail |
| `taskCreate` · `boardVisibility` · `membership` · `curator` · `inboxScanner` · `taskApi` · `schemaParity` | all pass |
| `bun scripts/privacy-publication-gate.ts --base <merge-base>` | PASS |

New behaviour is red-verified, not merely green:

- create counting rows again (the defect restored) → 3 of the new cases fail;
- the restore bound removed → the full-board restore case fails;
- the panel toggle dropping its result again → the refusal-toast case fails.

Scale matches what was observed: the history cases use 493 rows, and the cap
cases fill a board to exactly `BOARD_TASKS_PER_PROJECT_LIMIT` bands.

Pre-existing and untouched by this branch: `src/app/api/tasks/route.test.ts`
fails at the merge base too (`pipelines.json` ENOENT — #1622), and
`src/lib/tasks/revision.test.ts` passes alone but fails in any multi-file
invocation of that directory, because it asserts a `TASKS_FILE` resolved when
the store module was first imported.

No UI geometry changed: the only client edit reports an error that previously
could not occur.

## Follow-up filed, not fixed here

#1631 — history is unbounded by design now, and three consequences deserve
their own work rather than scope creep on a cap fix: the task panel renders
every row unvirtualized and is the only surface off-board rows are reachable
from; no UI affordance creates an off-board task, though the refusal names that
remedy; and the band-cap refusal is raw server English quoting an API field,
outside the en/uk catalogue (as its predecessor was). It also records that an
operator-driven retention sweep must skip any row with a non-failed assignment,
because `deleteTask` mints replacement placeholders for those.


Latand:
Adjacent finding filed while verifying: #1630 — `src/lib/tasks/revision.test.ts` is green alone and red in any multi-file invocation of that directory (it asserts a module-level `TASKS_FILE` resolved by whichever file imported the store first). Pre-existing at the merge base; untouched by this branch.

Latand:
Round response — all five notes are non-blocking and none required a source change; two were answered in the PR body, two are tracked in #1631, one is by design.

- **Unbounded history / unvirtualized panel** → filed as #1631 (history fold, virtualization, operator-initiated previewed retention). Not fixed here: pruning history in this PR would be the auto-deletion the directive forbids.
- **Untranslated refusal quoting an API field / no UI path to an off-board create** → also #1631. The string matches the convention of the refusal it replaces (server-side English, no catalogue plumbing for API errors), so changing it is a surface change, not a cap fix.
- **Default predicate leaves a staffed hidden band uncounted** → correct, and the commit message overstated it. The PR body now states the effective server rule (`board !== "hidden"`), what it can overshoot by, and why the two alternatives are worse. The predicate stays a parameter so a caller that can see the bands counts them.
- **Admission placeholders bypass the bound** → by design and unchanged from the old row cap: mandatory membership must never be refused, or an agent ends up outside every task.
- **Commit figures are a snapshot** → PR body now says so and gives the current reading (805 rows, 448 in the largest project, 7 bands), and names the separately authorized cleanup that moved them.

No commits added; head is still `75be3ba439802867a1f9fa4c37a42ea1c9977c82` with all checks green.

PR #1635, opened 2026-09-10
feat(runtime): bounded native Codex canonical history reader

Latand:
Native Codex delivery reconciliation needs bounded canonical history with preserved client, payload and turn identity. This adds an injected-RPC reader and original-key lookup for #1557 under #1629, based on `22ca45579f01bb0088b753cc2cf7b84c93422f58`.

Only three new files are included: the reader, its named tests and an integration note. Metadata binds thread/path identity; turns and items paginate with byte/page/deadline bounds. Summary/partial items require full item traversal. Unsupported pagination can request the established legacy fallback; incomplete history, missing records, identity mismatch and transport errors remain unknown.

Codex 0.154 serializes omitted text-span placeholders as null. The reader normalizes those equivalent forms while preserving text, ranges, explicit placeholder strings, original keys, turn fences and attachments. A credential-free native round-trip regression fails on the original reader and passes with the fix; changed placeholders and ranges remain unknown. Seam coverage also checks both omission/null directions and frozen-input preservation.

Validation: 17 focused tests and 159 assertions pass under Bun 1.4.0, including a credential-free real Codex 0.154.0 Responses fixture, second turns/items pages, full/summary equivalence, native shared-history cutoff, cold app-server replacement, and a negative control that removes the exact persisted native history row. Removing it leaves that original key unknown while another key remains found. Aggregate Unicode bytes, large tool items, wrong identities, malformed pages and exhausted budgets are covered. TypeScript and the full-diff privacy gate pass. ESLint's documented baseline plugin crash is not counted as green.

The compressed-rollout fixture retains Codex's native SQLite history projection. It proves pagination works in that state; rebuilding a missing projection from compressed files and compression-worker scheduling remain unverified. Existing CI does not invoke this new named test or install its optional CLI, so native fixture evidence is local.

**Integration remains owed:** after voice releases the host file, the integration worker must wire this into session materialization and original-key delivery lookup, preserve existing account/generation/lifecycle fences and deadlines, and prove Viewer host replacement without another send. No existing host, journal, HTTP, UI, voice or queue files changed. This PR does not complete or close #1557 or #1629. No merge or deployment is included.

PR #1634, opened 2026-09-10
fix(voice): configure native voice instructions and call context

Latand:
Configures the spoken prompt and native backing-model start/end instructions, enables transcript-tail flushing, surfaces the usage warning, and adds voice-context lookup plumbing. The intended outcome is reliable voice in the existing orchestrator conversation, with its role, tools and ongoing work preserved.

This component is incorporated into #1636 for parent #1629. **It is not independently approved for merge:** the exact-head review of `016271376270bb306839ed2ea6b38199ea2de61c` found two P1 context-binding defects and one P2 retention defect. The integration experience stage owns their repair.

Outstanding findings:

- A delayed or duplicate handoff can bind to a later utterance after the ambiguous queue is cleared.
- Conversation identity alone cannot establish which work may consume a selected-card reference.
- Ten-minute post-hangup retention can expose an earlier card to unrelated text work, expire ongoing work, and lose its binding on reconnect.

Evidence corrections from a stronger, successful local native session:

- A USER-delivered synthetic manager mandate reaches spoken startup context and remains in backing requests. The earlier developer-only failed-call fixture does not establish wholesale loss of the actual Viewer mandate.
- The native tool catalog is unchanged before, during and after voice in that fixture.
- Native records start/end instructions as canonical developer messages. The end instruction remains alongside earlier history; it does not literally remove the start text. Same-turn reconnect can retain the earlier start instruction until another mode transition.
- The native app's fallback spoken prompt also uses a general Codex identity. Its functional delegation, steering and USER/BACKEND framing must be preserved when adapting persona. A generic identity alone is insufficient causal proof of the reported behavior.

The original component passed 460 named local tests, TypeScript, privacy checks and applicable CI. Its notice/error images establish fixture presentation. Additional tests drove real Codex 0.154.0 through credential-free loopback Responses, WebSocket and MCP fixtures. Those results establish protocol/context behavior; real-provider audio, model quality and whole-call production behavior remain unverified.

No deployment is represented by this PR. Final readiness requires the remaining #1629 integration, repaired context behavior, and independent review of the complete integration head.

PR #1633, opened 2026-09-10
Add native Codex queue transport adapter

Latand:
Adds the native Codex queue transport adapter needed by #1629. It supports list/add/update/delete/reorder/start while keeping mutation acknowledgements separate from delivery and preserving uncertain outcomes after lost responses.

Closes #1632. Parent host/UI/journal integration remains owed, including default interrupt, capability negotiation, request-error classification, notification forwarding, operation recovery, and canonical delivery settlement. Existing host, HTTP, journal, composer, and voice files are untouched.

The adapter binds one thread to an injected RPC connection. It preserves native input/media and server/client identities, coalesces bounded pagination, and retains stale snapshots after incomplete or contradictory reads. Notifications invalidate state without scheduling sends. Native update remains available for edits. Disposal before invoking the RPC port returns a typed `not-submitted` outcome; failures after transport invocation preserve uncertainty.

Validation:

- Bun 1.4.0 with frozen lockfile dependencies and isolated home/config/cache/data/provider/state/temp roots.
- 54 focused tests / 164 assertions, including an opt-in Codex 0.154.0 fixture using a credential-free loopback synthetic backend. Real pagination and every mutation response pass. Dropping a successful add acknowledgement returns uncertainty without retrying.
- Four private negative controls detect missing completion, lost invalidation, lost uncertainty classification, and lost client-identity validation.
- Freshly generated experimental CLI schemas match all 13 audited queue schemas.
- TypeScript checking and the full-diff publication gate pass.
- ESLint is unavailable: its React plugin crashes with `contextOrFilename.getFilename is not a function` on both the new module and an unchanged baseline file.

The current hosted workflows run their existing named suites; the new focused test is verified locally. Parent integration must add its ongoing CI coverage. This PR is ready for the pipeline's independent review. No merge, deployment, or production runtime changes were performed.

PR #1636, opened 2026-09-10
Support native Codex queue, steering, and orchestrator Voice

### LINKED ISSUE #1629: Native Codex parity: reliable orchestrator voice, responsive sends and native queue support
The operator requests full native Codex support for message sending, queue management, and live voice in Viewer. Keep interrupt as the default send behavior, make interactions responsive, and port the relevant native ChatGPT/Codex application logic into the existing Viewer architecture. Voice currently appears to lose the orchestrator role, available tools, and useful context; restoring that behavior is the first priority.

This direction supersedes the earlier decision to defer native queue adoption and the historical voice rollout deferral. It does not authorize a new account, API-key fallback, hidden agent runtime, or discarding unresolved message history.

Verified baseline: Viewer 22ca45579f01bb0088b753cc2cf7b84c93422f58; Codex CLI 0.154.0; current Linux ChatGPT 26.903.61454 with bundled app-server 0.153.4; Claude Code 2.1.266. Recheck actual versions and running host generations before accepting behavior.

Evidence already available:

- The isolated Codex contract probe in docs/design/codex-api-update/probe.py passes 46 cases and 15 assertions on 0.154.0. Native queue adds with the same client message ID create distinct submissions; an idle add and a normal cold resume can dispatch automatically. Queue disappearance does not prove delivery.
- Viewer drops isBlocking from question projection. True and false requests produce identical projected attention. Any pending request currently makes the host report attention and prevents the active-only steer path.
- Voice records selected context, but production routing has no consumer of voiceSelectedContext. A late older reference can overwrite a newer reference from the same window.
- The native app subscribes to thread/realtime/transcript/delta and /done and uses flushTranscriptTailOnSessionEnd=true. Viewer currently uses older browser event names and omits that flag. Live wire behavior still needs causal verification; schema differences alone do not prove a failed call.
- The existing modality/coordinator persona fix is already in main. A new failure must be reproduced without assuming that historical fix never shipped.
- Native turn/start supports model, serviceTier and serviceTierForTurn. Native turns/items pagination works in the installed CLI.

Implement in this order:

1. Diagnose and fix live voice parity. Compare the installed native app's complete startup, persona/context, MCP/tool exposure, WebRTC/sideband ownership, delegation and handoff, worker responses, interruption, compaction, transcript-tail flush, stop and reconnect flow. Preserve an existing orchestrator's role, authority, tool access and active work. Correlate selected context with the actual utterance/handoff and reject reordered or replayed publications. Do not add repeated persona text, persona-driven self-relay, duplicate speech, lost terminal chunks or duplicate work. Propagate real provider failures and usage warnings.
2. Integrate native Codex send/queue behavior, including queue read/add/update/delete/reorder/start where supported, and native lifecycle events. Keep default interrupt. Provide explicit queue and matching-turn steer actions. Port native sequencing and responsiveness while retaining one dispatch owner and Viewer operation identity. Derive cancellation/edit semantics from canonical evidence; native queue IDs, client IDs, operation IDs and turn IDs remain separate. Avoid dual schedulers and blind retries. Preserve the Claude stream broker's supported behavior and expose capability differences clearly.
3. Incorporate the required pagination, nonblocking questions, frozen message profile and capability diagnostics from #1557, #1558, #1559 and #1561. Evaluate #1560 against the new native parity scope. Preserve existing ownership of unrelated delivery repairs #1576, #1609 and #1612; coordinate any shared files before editing them.
4. Update outdated protocol/voice documentation with actual versions, behavior, limitations and reproducible checks. Preserve historical incident evidence.

Acceptance:

- A real existing orchestrator keeps its role and advertised tools throughout a fresh voice call, reconnect and text continuation; the assistant can correctly perform a harmless Viewer read requested in speech. Use an isolated fixture/call and bounded synthetic audio, preserve all active operator calls and sessions, and distinguish source/fixture evidence from provider/browser evidence.
- Final spoken input followed immediately by hangup survives in canonical context. Barge-in, interrupted output, duplicate notifications, late handoff/worker responses, and compaction do not replay earlier speech or execute an instruction twice.
- Queue/steer/interrupt and nonblocking questions behave correctly with actual installed app-server protocol, concurrent arrivals, restart, uncertain acknowledgements, account changes, attachments, and explicit turn fences. Negative controls remove the canonical evidence and turn successful assertions red.
- Measure submit-to-visible-feedback and delivery boundary latency against the same populated-board fixture and current deadlines. Verify changed controls in a rendered browser; DOM assertions alone do not prove smooth behavior.
- Named isolated tests, TypeScript checking, privacy gate and exact-HEAD CI pass. Fresh independent review reports no unresolved findings on the complete diff. Publish a ready PR or clearly linked stacked PRs; no merge or deployment by implementation agents.

All worker activity is visible through Viewer pipelines, with at most three concurrent workers project-wide and one writer per file. Preserve previous worktrees, receipts, transcripts and artifacts. User data cleanup and live runtime manipulation are outside this issue.

Latand:
Codex messages can now use the native queue, explicit steering, or the existing default interrupt. Voice runs on the current orchestrator with its role, tools, and explicit conversation context preserved. Fixes #1629 and integrates the reviewed queue, history, and Voice work from #1633, #1634, and #1635.

- The composer reads Codex’s queue order and supports edit, remove, reorder, start, and send now. Alt+Enter or the send menu queues a draft. Enter keeps the default send behavior. Unsupported capabilities are refused before admission; Claude steering remains unavailable.
- Native queue and handoff controls retain immutable request identities and payload versions through lost acknowledgements and reloads. Unconfirmed operations remain recoverable. Queue admission and delivery are shown separately, and capacity or storage failures preserve the draft. Queued messages use the thread’s runtime profile at dispatch.
- The queue, call panel, and receipt controls share a bounded scroll area above the input. Input/Send remain reachable on compact cards and phones; the send and runtime menus open outside that scroll area.
- Voice uses native session instructions without replacing the orchestrator’s persistent role. Canonical voice transcript events, final transcript flushing, tool context, nonblocking questions, and UTF-8/history handling follow the native protocol. Explicit conversation/context targeting is supported. Automatic attribution of a selected card to a spoken utterance remains unavailable because the installed protocol does not provide the required correlation.
- Integration preserves the current-owner Interrupt/Stop controls and the transcript-tail fix that prevented sends to long Codex histories.

Validation: independent review of the native contracts and compact composer; 635 native tests plus 107 history/related tests, 250 focused UI tests, 37 assembled capture cases, and pointer/keyboard traversal of all 662 controls in the combined compact fixture. The final integration adds 139 focused tests and 27 browser control scenarios. Typecheck and privacy checks pass; exact-head CI and final integration approval are recorded before release.

Browser/runtime fixtures use synthetic data. Live provider Voice verification and production acceptance follow deployment. Large attachment durability beyond the native handoff store is tracked separately in #1647 / #1652.


Latand:
P2: consecutive refused updates can select an unaccepted queue payload.

At `29c9521b94676514dddd69e2b6e54f50b535950f`, `src/runtime-host/nativeQueueJournal.ts` restores a refused update with `entry.versions.at(-2).revision`. The versions array retains rejected attempts, so that is not necessarily the last accepted revision.

Isolated journal reproduction:

1. Add and acknowledge revision 1, text `accepted-original`.
2. Admit revision 2, text `refused-one`, then receive an authoritative refusal. Current revision correctly returns to 1.
3. Admit revision 3, text `refused-two`, then receive another refusal.
4. The queued record now selects revision **2**, text **refused-one**, although neither edit was accepted. The expected current revision is still **1**.

The retained immutable command already carries the validated `expectedRevision`; use the actual prior accepted revision for rollback and retain rejected versions only as history. Add the consecutive-refusal case and assert that presentation and later dispatch use the accepted payload. Keep canonical proof and concurrent-delivery handling intact.

This is a runtime integration finding for the current experience/repair owner; it does not require restarting the completed component workers.

Latand:
Three integration checks currently fail when the new queue panel payloads go through the actual command parser and an in-memory journal:

- Editing text on an image-bearing queued message submits no `images`. The parser defaults that to `[]`, and the admitted edit loses the attachment. Preserve existing attachments when editing text.
- The header Start control submits `action: "start", turnId: null` without `entryId` or `expectedRevision`. Parsing refuses it with `entryId is invalid`.
- A withdrawn row's Send now control submits `action: "send-now"`. The journal refuses it with `withdrawn input requires an explicit idle start`. Route this recovery through the supported idle-start action and retain the frozen payload.

These observations cover the in-progress UI on top of `a0812fa1`; they must be checked again against the final review head. The probe uses isolated local state and has no live runtime effects. Add control-to-parser/journal coverage so a mocked successful submit cannot hide these failures. The earlier consecutive-refused-edit finding also remains part of the review scope.

Latand:
A further Voice regression reproduces in the committed `a0812fa1` ledger:

1. Speak about card A and report handoff A.
2. Tool call A on a realtime native turn reads card A.
3. Speak about card B and report handoff B into the same native turn.
4. A new tool call with different call/item IDs still receives `joined` card A and handoff A.

`voiceUtteranceContext` returns the first record bound to the turn before checking later evidence. Multiple realtime handoffs sharing one turn is verified native behavior. The effectful MCP selected-context resolver accepts this `joined` result, so the old card can become the target of the new request.

A turn ID does not establish per-utterance ownership. Preserve the context of already accepted operations, but refuse unproven implicit targeting for new requests; keep explicit conversation/context arguments usable. The verified native protocol lacks the exact utterance-to-handoff and handoff-to-accepted-work edges needed to infer this safely from order or candidate count. Please include this case in the final integration repair and review.

Latand:
The real composer also permits a duplicate after an unknown native queue admission:

- Make the queue transport throw after possible server acceptance.
- Submit with Alt+Enter. The error path restores the draft and retains no outbox entry.
- Submit again. The composer emits a second write with a new idempotency key.

An isolated DOM reproduction using the current composer reports two distinct keys where the check requires one. `queueForCodex` clears the draft/attachments before admission and `useNativeQueue.submit` reports a transport loss as a generic failure. The error path treats it like a definite refusal.

Keep the original key, payload and recovery state through unknown admission and reload. A durable presentation record can support this while Codex remains the sole dispatch owner. Also preserve attachments on definite refusal; the current restoration restores only text. This needs a negative-path DOM regression in addition to the immediate-feedback test.

Latand:
Rechecked all four conductor reproductions at published HEAD `9307c626276d176ec4f09c732e19a292425d402f`, with relevant source hashes unchanged during the runs. The failures above remain: rejected-edit rollback, the three queue control/attachment defects, same-turn Voice misbinding, and a second operation key after unknown admission. The independent full integration review and repair remain required; green component checks do not resolve these cases.

PR #1652, opened 2026-09-10
Preserve attachment submissions and unblock new drafts after uncertain sends

### LINKED ISSUE #1647: Preserve large attachment payloads across composer remounts and reloads
A failed message with four images remains visible as text plus `4 images`, but its retry control disappears because the complete payload cannot be restored. The operator kept the page open; component remounts also need to preserve payloads.

The actual pending-delivery persistence API was exercised in isolated Chromium. A small image retained its bytes and complete envelope. Four images totaling 16 MiB of base64 exceeded synchronous session storage; the fallback retained only text/key metadata with `payloadComplete:false` and zero images. Reload kept that incomplete record. The in-memory owner had all four images before disposal.

Store complete submitted attachment payloads durably before clearing the draft or sending. Preserve immutable conversation/key/context/runtime bindings and unknown outcomes. Support the existing allowed payload sizes using suitable browser storage; storage failure must keep the full draft and prevent an unsupported admission. Restore actual image/file bytes across card remount and page reload, and provide truthful recovery controls for older metadata-only records.

Acceptance: causal real-browser large-payload checks, remount/reload retaining exact bytes and original keys, current browser storage failures preserving the draft, no silent text-only replay, no eviction of unresolved operations, and settlement/discard cleanup only after authoritative matching evidence or explicit operator action. Add a focused PR without overlapping the current native composer writer.

Latand:
Large image and file submissions now retain their complete bytes before the composer clears the submitted draft or lets a request onto the wire. Fixes #1647.

**Retention.** IndexedDB stores the authored submission and its exact wire envelope, keyed to the original conversation and request. Reload, card changes and a new browser context keep the same images, files, context, runtime settings and key. Storage or quota failure at either phase keeps the draft and attachment tray and sends nothing. Old metadata-only attempts stay visibly incomplete while a separately authored new message proceeds under its own key. The recovery row states the server's actual diagnostic, and its controls share the compact composer's space budget (reachable on a 500px card).

**Retry of an admitted message.** Once the runtime journal admitted a message, its next attempt comes from `POST /api/runtime/operations/<id>`, the same contract as the receipt Retry. Resending the sealed envelope under the original key was checked against the real send handler and journal: it answers 409 with the recorded failure, delivers nothing, and its refusal deletes the inbox files the admitted operation names. The saved-message row, the queue bubble and the receipt controls now all use the operation contract. The journal starts a leaf from the recorded request, so every original byte travels, and a repeated request converges on the same leaf. The payload store reads receipts in the journal's presentation identity: raw and presented leaf receipts both bind to the admitted operation, and only the current attempt's latest receipt settles and releases the copy. Foreign, stale and admitted-only receipts still retain it. A converged retry answer records the delivered leaf before any stream copy reaches the composer, so cleanup reads the latest recorded receipt whether or not the pass adds a revision: a lost 202 followed by a repeat that answers 200/delivered releases the copy at once. Previously it stayed retained under an "outcome is unknown" row across Re-check and reload.

**Native queue hand-offs.** Alt+Enter and the send menu's "Queue for Codex" kept the whole envelope in one sessionStorage slot, so four screenshots (16 MiB encoded) were refused before the wire although the queue route admits 24 MiB. A hand-off too large for the slot now writes its envelope to IndexedDB through the same payload store and only its identity to the slot: key, original binding, the command without its bytes, the stored copy's fingerprint and a digest of what was authored. Both commit before the request leaves, and the draft stays until they do. Pressing the same message again finds that operation by the authored digest and replays the stored envelope under its key and binding. The unresolved row's "Send it again" reads the envelope back and verifies it first; a copy that does not verify is not sent and its identity stays. The durable record is written in a shape an older build carries as opaque, entries this build cannot name survive every write, an unreadable slot still refuses, and small commands and panel controls keep the synchronous slot path. Native API semantics and the single dispatch owner are unchanged.

**A held queue acknowledgement no longer blocks Enter.** A large hand-off held the composer's shared submission guard until the journal answered its upload, so while a 16 MiB acknowledgement was on the wire, Send stayed enabled and Enter on a new small command sent nothing. Only the IndexedDB and slot retention hold the guard now. The upload and its answer run after the guard is released, and they settle that one operation under its original key, binding and payload. A late refusal gives the draft back whole, and only into an empty composer. It is never placed beside newer words or over a tray that is still reading. An unknown outcome keeps the operation recoverable from the panel. A second press during retention still mints nothing, and words typed while retention runs stay for the next message.

**Files ride a queue hand-off.** The queue route read images and dropped `files` without an error, so a hand-off with a staged document queued the text and lost the file. The route now takes files the way an ordinary send does. The same admission refuses a bad file with its reason, the bytes land in the viewer inbox under a batch derived from the original key, and their paths are folded into the queued words before the command is parsed. The version's digest covers the files, an edit that carries the text keeps them, and a replay of the key rebuilds the identical command. Files are written only after the command validates and are never written over an existing file: a path already holding the same bytes is reused and is not this request's to release, and one holding other bytes answers 409. A terminal refusal deletes only the files that request created, so a rejected replay or an idempotency conflict never removes a file an accepted message names. Staging, admission and that release take one turn per inbox batch. Without the turn, a request under the same key with the same bytes and different words could reuse a file while its creator waited on the journal, be admitted first, and then lose the file when the creator's refusal deleted it. Requests under different keys do not wait for each other, and the wait is bounded by the runtime client's timeout. A control that names files is refused. The composer hands the whole tray to Codex's queue, keeps the files in the retained envelope (in IndexedDB when large, with a count in the slot), replays them under the original key, and gives a refused draft back with its files ready. A message without files keeps the authored identity it was stored under.

**A refused hand-off stays recoverable.** A late refusal used to release the retained envelope, IndexedDB bytes included, before the composer decided whether to give the draft back. With a newer draft in place, the words, images and file were gone, and even a draft that did come back lost its large images and the file's bytes on reload. A refusal now marks the retained record as refused and keeps its bytes. The queue panel lists it with the runtime's reason. "Put back" restores the verified copy into a composer that holds nothing else, or only what a reload left of that same message, and never over a newer draft. "Discard" removes the copy and its bytes. Queuing the same message again is a new operation under the current binding, and once that operation is retained it replaces the refused copy. A refused copy is never replayed under its old key.

**Refusals and unknown outcomes.** A 400/403/413/415 answer that names no operation is recorded as a refusal before admission. After reload the row states the server's reason and offers the same sealed envelope again or a discard. A 409 or 5xx answer without an operation stays unknown. Unknown outcomes are never resent, and a retained copy is never offered the uncertain retry; its row offers Re-check until authoritative evidence arrives. Once an attempt was unknown, only a proven safe rejection re-opens its Retry.

**Suites that regressed against main.** A switch-held entry restored after reload was dispatched twice under its key; the claimed replay now counts as that hold cycle's one replay. The generic "Delivery couldn't be confirmed" guidance had vanished for text-only messages; it is now folded only into rows that state their own diagnostic. The image-bearing DOM suites had no IndexedDB, so a test helper supplies an in-memory IndexedDB and lock manager. Where the payload contract deliberately changed an expectation (operation retry for admitted images, Re-check for a retained copy's unknown fate, no send fence behind metadata-only records), the test now states that contract.

**Validation** (pinned Bun 1.4.0, isolated state, fake engine host, no provider or account):
- `composerPayloadRetry.integration.test.ts`: real send, retry and query handlers over a real runtime host journal and registry. An admitted send fenced by a dead host; the same-key resend returns 409 and delivers nothing; the operation retry delivers all four images and the file once; a repeat converges.
- `scripts/verify-composer-payload-runtime.mjs`: Chromium drives the assembled composer against those handlers. Four admitted messages, each with four 3 MiB images (16 MiB encoded) and a binary file, are fenced while the host is down, and a retry against the down host gets 503 with nothing delivered. Once the host is back, the receipt Retry, the saved-message Retry after reload (double-clicked) and the delivery notice Retry in a new browser context each deliver exactly once with byte-identical images and file. The saved-message Retry of the fourth message loses its 202 answer while the page's receipt stream lags behind the journal. The copy and its Retry stay. The repeat converges on the delivered leaf with 200, and that answer alone releases the copy while the stream still lags. After a reload there is one byte-identical delivery, one leaf, no resend and no row. Against the previous head this case fails with the copy retained. Each leaf names its own parent, its reservation settles delivered, and only its own copy is released. A repeat after delivery and a reload deliver nothing more. A request body cut short (400 invalid JSON) shows its reason after reload, and the same envelope is resent and delivered once. An ambiguous 409 offers neither resend nor discard.
- `scripts/verify-native-queue-payload.mjs`: Chromium drives the composer against the real queue route and journal. A 16 MiB four-image Alt+Enter hand-off holds about 500 characters of identity in the slot while in flight and is admitted once with byte-identical stored images. A send-menu hand-off whose answer is lost survives a reload and "Send it again" replays the same canonical envelope into the same single journal operation. Alt+Enter again after a lost answer keeps the key. A small text hand-off clears in under 100 ms on the synchronous path. An opaque record is carried through (the verifier now waits for the small hand-off's slot release before seeding it; seeding earlier let the page's mirror overwrite it, about one run in three); an unreadable slot, missing IndexedDB and a 1 MiB quota each refuse with the draft and all four tiles kept and nothing sent. After settlement no envelope is left in IndexedDB, and the journal holds one operation per key.
- `scripts/verify-native-codex-delivery.mjs` (new): Chromium drives the assembled composer through the real send and queue routes, runtime journal, structured delivery queue and native queue executor into the installed Codex CLI 0.154.0. The model provider is a credential-free local Responses server that records what Codex sends. Four distinct 3 MiB PNGs and a binary file are handed to Codex's queue as one envelope while a turn runs, and the queue answer is held on the wire. Enter on a small command reaches the wire within 25 ms while that answer is still held; against the previous composer it never leaves. Codex then dispatches the queued entry as one canonical input (four localImage items and one text), with one queue add and one journal operation under the original binding. Its four localImage paths, read through the Codex process's own mount namespace, and the four images the provider receives are byte-identical to what the browser attached; the inbox path the text names reads back byte-identical through the same namespace, and the provider's text carries it. Asking the original key again after delivery answers 202/delivered, adds nothing and leaves the file in place. A second image-and-file hand-off loses its answer; after a reload the unresolved row replays the same envelope, files and binding into the same single operation, the provider sees one turn, the file reads back through Codex, and a terminal replay leaves it in place. Against the previous queue route the same run delivers the words with no path and no file. The journal then refuses a four-image and binary-file hand-off (the verifier rewrites that one request's account on the wire), and the answer is held until a newer draft is typed. The refused copy stays in the slot and IndexedDB and survives a reload. Put back is refused over the newer draft. After that draft goes out with Enter, Put back restores all five attachments ready. Queued again, the message is one new operation that Codex dispatches as one canonical input, with images and file byte-identical through the Codex process's namespace, and the refused copy and its stored envelope are gone. A second refused copy that the operator discards leaves no slot record or stored envelope, and its draft stays. Against the previous composer the run stops at the refused copy, which no longer exists. The same kind of draft of four images and binary file sent with Enter, with the answer lost, reach the provider once. The inbox file named in the text reads back byte-identical through the Codex process, the reload sends and retries nothing, and asking the original key again after the terminal answer (200/delivered) leaves the accepted file in place with no second turn. The final verifier passed against Codex 0.154.0.
- Native queue DOM suite (28) adds the held acknowledgement cases: an ordinary command sends while it is held, and success, refusal and loss each settle only the original operation. The refusal cases cover a newer draft, an empty composer, a reload that loses the draft's bytes, Put back, Discard and queuing again; the four refused-copy cases fail against the previous composer. A double press and typing during retention are covered too. A document rides the hand-off beside its image and comes back ready after a refusal; a large hand-off with a document keeps it in the durable envelope and replays the same files under the same key after a lost answer.
- `nativeQueueHttp.files.test.ts` (15, new): the queue route over a real journal and a private inbox. Files become inbox paths in the queued text, including a files-only message. Replaying the original key is the same operation and leaves the accepted file untouched. A lost answer keeps the file and the replay finds the committed operation. A transport failure keeps it. Different bytes under the original key answer 409 without touching the accepted file. An idempotency conflict releases only the file it created. A stale binding is refused under its key with its files released. Invalid files and a control naming files are refused with nothing written. An edit keeps the paths its words name and can carry new files under its own key; a stale edit releases only its own. A replay after delivery answers delivered and keeps the file. Two concurrent requests under one key cannot delete the file the admitted one names, whether the creator or the later request is refused, and requests under different keys do not wait for each other. 11 of the first 12 fail against the route before files were admitted, and both concurrent-request cases fail against the previous head.
- The synthetic assembled, scenario, legacy, adapter and storage browser verifiers pass, with their retry assertions moved to the operation contract.
- TmuxComposer unit (31), native queue DOM (28), native queue panel (31), the retained queue admission module (13), inbox files (11), runtime attachments (8), runtime HTTP (34), native queue runtime (26) and the native queue host integration against Codex 0.154 (3) pass. Every TmuxComposer and outbox DOM suite passes, or fails exactly the tests that fail on main: runtime snapshot (7), live refresh focus (3), LogFeed delivery uncertainty (30), delivery wait (1).
- Typecheck and the privacy gate pass.

**Limits.** Original attachments lost before this change cannot be recreated from metadata. For a queue-fenced failure the settlement query answers `verify-first` while the journal answers plain `failed`; the answer a browser sees decides whether Retry is offered, and both server answers are unchanged here. A leaf that supersedes its original before this browser observed the original cannot be bound and stays unknown. The ordinary-retry verifiers still use a fake engine host. Native delivery is now qualified against the installed Codex 0.154 CLI with the host and history-reader repair from #1661, which main now carries. A real model provider is not exercised. A queued file reaches Codex by path in the message text, as on Enter, so an operator editing a queued message sees and can remove those path lines. Real Claude delivery, and real Codex delivery of an ordinary retry leaf, rely on the existing host tests. A queue replay's request body carries the stored envelope's key order, which the journal's canonical request hash treats as the same request.


Latand:

PR #1670, opened 2026-09-12
Group a task band's pipelines into enclosed sections and bound stage expansion (#1668)

### LINKED ISSUE #1668: board: a task band's pipelines are indistinguishable — detached heading pills over one undifferentiated stage grid, and unbounded stage expansion
## What the board looks like today

On the task-centered board (#1586), a task band that owns several pipelines renders as:

1. A stack of narrow, differently-coloured outlined pills at the top of the band — one per pipeline — each with a hard-truncated title and a `k/n · <lifecycle>` tail.
2. Below them, **one undifferentiated grid** of every stage card from **every** pipeline in that band, wrapped left-to-right across the band's full inner width.

The pills are detached from the cards they name: nothing ties pill colour to any card, and a pipeline's stages wrap across rows interleaved with another pipeline's stages. With four pipelines in one band, `stage 4/5` and `stage 1/2` sit side by side in the same row, and the only way to work out which stage belongs to which pipeline is to read every card's `k/n` and mentally re-sort them. Connectors then cross rows between cards that are visually unrelated.

The mechanism is in `layoutTaskBands` (`src/components/scheme/taskBands.ts`):

* every container heading is placed at `cursorY + headerH + index * BOARD_SURFACE.groupHeader` — all headings stacked before any card;
* every member of every container is then flowed through **one** wrap loop with no break at container boundaries;
* `GroupsLayer` hides the enclosing region for a band heading (`group.bandHeader ? "hidden" : ""`, `src/components/scheme/nodes.tsx`), so the heading is left as a floating pill with no region under it.

The comment above the heading placement already states the intended rule — "Each container gets a dedicated heading before its members. A heading never borrows the task title row or spans another container's cards" — and the code does not implement it.

## Expanding a settled stage

`stageSurface(true)` reserves `600 × 724`: a 104px status row plus a 620px card, **flush**, no gap. The disclosed `StageCompletedCard` then:

* repeats the stage heading the status row directly above it is already showing;
* repeats the state badge the row is already showing;
* pushes its "open conversation" control to the bottom of the 620px box with `mt-auto`, leaving a large blank area between the prompt block and the control;
* forces every connector attached to that stage into a long vertical sweep down the board.

## Other readability defects in the same surface

* `groupLabel` falls back to `spec.key` — a raw `group::pipeline::<uuid>` — when a container has neither a pipeline nor a flow record, putting an opaque identifier in the headline position.
* A pipeline whose recorded goal is empty gets an empty heading label.
* A future stage that has not started reads `pending` in the badge **and** `pending` as its entire explanation line, which says nothing about whether the stage is simply queued behind earlier stages or genuinely held because the pipeline needs a decision.

## What should change

* A task band is the semantic parent. Each pipeline/flow it owns is a visibly enclosed section **inside** it, its heading attached to the top of its own region, its stages wrapping **within** that region, with deliberate separation between sections.
* Headings keep their meaningful colour, attached to the region rather than floating as a detached pill, and accommodate a long title instead of hard-truncating it.
* Expansion is bounded, separated from the compact row it belongs to, and does not repeat what the row already says; the complete prompt and the full transcript stay reachable.
* A stage's explanation line states truthfully what the stage is waiting for. Nothing may present unstarted work as finished.
* The stage graph, navigation and every existing control keep working; technical identifiers stay in detail surfaces, never in a headline.

Latand:
Closes #1668.

## What changed

**A task band's pipelines are now enclosed sections, not a shared grid.** Every container heading used to be stacked at the top of the band body, after which every container's cards were flowed through *one* wrap loop. Four narrow coloured pills floated above one undifferentiated grid in which a five-stage pipeline and a two-stage one interleaved in the same rows — the only way to attribute a stage was to read every card's `k/n`. Each container now owns a section: its heading, its own rows, and a framed region enclosing both, with `BAND.sectionGap` between sections. Items belonging to no container keep an unframed section after them. Section items are measured against the *section's* usable width (`maxInnerW - 2 × sectionPad`), so a card scaled down on a phone-width board cannot overhang the frame that names it.

**The heading is attached to its region.** It keeps the container's colour — now tinting the territory it actually owns rather than floating as a detached tag — wraps its title to two lines instead of hard-truncating, and states progress in words (`stage 2 of 2`) next to the lifecycle. Every control the pill carried (disclosure, `data-pipeline-group-header`, progress, lifecycle) is unchanged. Outside a band the free-map halo keeps the counter-scaled pill it always had.

**Technical identifiers left the headline.** `groupLabel` returned `spec.key` — a raw `group::pipeline::<id>` — for a container with no record; it now returns an empty label and the heading names the kind ("Pipeline without a recorded goal"). The pipeline goal is cleaned to 120 characters rather than 60, because the heading wraps to two lines.

**Stage expansion is bounded and stops duplicating its own row.** `stageSurface(true)` reserved `600 × 724` and glued the card to the row with no gap. A settled stage's disclosure is now `600 × 460`: a deliberate `stageDetailsGap` below the row, no repeated heading, no repeated state badge, and only what the row cannot carry — engine, kind, model/effort, verdict, and the prompt, which scrolls inside the bounded card so the whole prompt stays reachable. The transcript link follows the prompt instead of sitting at the bottom of a mostly blank box. A planned stage keeps the full editor height it needs. Placement and rendering read one function (`stageDetailsCardHeight`), so the reserved rectangle and the drawn card cannot disagree.

**Unstarted stages explain themselves truthfully.** A stage with no attempt used to render `pending` in the badge and `pending` as its whole explanation. It now states what it is waiting for: queued behind a named stage, held for a decision, paused, planned, or never reached because the chain ended first. Every one of those says the stage has **not** started, because none of them has — nothing here presents unstarted work as finished, and nothing claims a stage is blocked when it is merely later in the chain. The row gives its title and its explanation two lines each, with the full strings still reachable as tooltips.

The stage graph, connectors, navigation, selection, disclosure state and every control behave exactly as before; this changes where surfaces are placed and what they say, not what they do.

## Tests

New focused coverage:

* `taskBands.test.ts` — two pipelines of one task get no shared rows; each region encloses exactly its own cards below its own heading; the sections do not overlap and sit `sectionGap` apart at equal width. Repeated across widths 375 / 680 / 830 / 1024 / 1440 / 1920 × overview / intermediate / near: no card of one section ever touches another section's frame.
* `GroupsLayer.render.test.tsx` — a band container draws a visible region with a full-width heading in its own colour, progress in words, a two-line clamp and no truncation, keeping every control; a container with no goal names its kind rather than its board key; the free-map halo is untouched.
* `stageCardAria.render.test.tsx` — a disclosed card repeats neither the row's heading (in *visible* text; the accessible name is deliberately retained) nor its state badge, scrolls its prompt, and drops `mt-auto`; the reserved footprint equals row + gap + card for both presentations, and a settled disclosure is bounded below a planned one.
* `pipelineModel.test.ts` — each not-started case yields its own sentence, the cursor stage still reports its live state, and a stage that did run still reads from its own record.

Verification:

| check | result |
| --- | --- |
| `taskBands.test.ts` | 40 pass / 0 fail |
| `GroupsLayer.render.test.tsx` | 13 pass / 0 fail |
| `pipelineModel.test.ts` | 124 pass / 0 fail |
| `stageCardAria.render.test.tsx` | 5 pass / 0 fail |
| `SchemeBoard.bands.dom.test.tsx` | 11 pass / 0 fail |
| `RoundDeck.dom.test.tsx` | 7 pass / 0 fail |
| `i18n.test.ts` | 20 pass / 0 fail |
| `privacy:check` | PASS |

`bun test src/components/scheme/ src/components/pipelines/` is **19 fail / 992 pass** on this branch and **23 fail / … ** at the merge base with the same command. Comparing the two failure sets name-by-name, this branch introduces **zero** new failures; the four that differ are cross-file ordering flakes that fail at the merge base and not here. Every one of the 19 is red on `main` before this branch exists.

## Browser evidence

Driven against a dev server on the real board through raw CDP on headless Chrome, at 1600 / 980 / 700 CSS px and on the phone surface. Rasters are not committed (the gate requires generator provenance for images), so the measured geometry and what the frames show:

**Desktop, one task owning five pipelines.** One outer task container carries the title, `assigned`, `0 working · 4 conversations · 4 planned`, `Details` and `+ Agent`. Inside it, five separately framed sections in five hues, each headed by its own bar (`Preserve complete attachment submissions through send and reload` · `stage 1 of 2` · `needs a decision ▾`) with that pipeline's two stage cards inside the frame beneath it. No card sits in a row with another pipeline's card. Pending cards read `Not started — the pipeline is waiting for your decision`.

**Narrow (980 and 700).** The band narrows to one card per row and each pipeline wraps *within its own frame* — stage 1 above stage 2, joined by the handoff connector, both inside the section. Headings wrap to two lines; the longest clamps with an ellipsis and keeps the full string as its tooltip. No overlap, no overflow past any frame.

**Expanded stage.** Measured at 58% board zoom: row `h≈60`, gap `7` (= 12 board px), card `h≈200` — the row states `Builder · finish · stage 1/2` / `✓ passed` / `Reproduced your P1 exactly, then fixed it.`, and the card below carries `Claude` `RUN` `pass` · `Opus 5 · high`, the stage prompt, and `Open conversation` directly under it. No repeated heading, no repeated `passed`, no blank lower area.

**Phone (430 × 932).** The board renders the mobile list surface, which carries no band or section nodes at all — 49 controls, none under 28 px tall, `scrollWidth - clientWidth = 0`. This change does not reach it.

## Not done here

No merge, no deploy. Nothing outside board task/pipeline presentation was touched; no task prose was written, no board state mutated.

### DIFF
--- a/src/components/pipelines/StageCompletedCard.tsx
+++ b/src/components/pipelines/StageCompletedCard.tsx
@@ -29,7 +29,16 @@ import {
  * difference from the eventual live conversation), the settled state/verdict, and
  * an affordance that opens the full transcript on demand.
  */
-export function StageCompletedCard({ slot, onOpen }: { slot: StageSlot; onOpen?: () => void }) {
+export function StageCompletedCard({ slot, onOpen, disclosed }: { slot: StageSlot; onOpen?: () => void;
+  /** Rendered underneath its own {@link StageStatusRow} as that row's
+      disclosure (#1668). The row directly above already names the stage and
+      states how it ended, so repeating both here read as two cards for one
+      stage and pushed the transcript control to the bottom of a mostly blank
+      box. In this mode the card drops the duplicated heading and state badge and
+      adds only what the row cannot carry: the runtime, the verdict, and the
+      prompt that was actually sent — which scrolls, so the whole prompt is
+      still reachable inside a bounded surface. */
+  disclosed?: boolean }) {
   const { t } = useLocale();
   const { pipeline, stage, attempt } = slot;
   const state = stageChipState(pipeline, stage);
@@ -58,7 +67,7 @@ export function StageCompletedCard({ slot, onOpen }: { slot: StageSlot; onOpen?:
       style={{ borderColor: tone.color }}
     >
       <span aria-hidden className="h-1 w-full shrink-0 opacity-60" style={{ backgroundColor: tint.color }} />
-      <header className="flex h-10 shrink-0 items-center gap-1.5 border-b border-border px-2.5" style={{ backgroundColor: tint.soft }}>
+      {disclosed ? null : <header className="flex h-10 shrink-0 items-center gap-1.5 border-b border-border px-2.5" style={{ backgroundColor: tint.soft }}>
         <span className="h-2 w-2 shrink-0 rounded-full" style={{ backgroundColor: tone.color }} title={t(`pipelineChipState.${state}`)} />
         <span className="shrink-0 rounded-full border border-border bg-card/70 px-1.5 py-0.5 text-caption font-bold capitalize text-muted">
           {stage.effectiveRole.engine}
@@ -69,14 +78,26 @@ export function StageCompletedCard({ slot, onOpen }: { slot: StageSlot; onOpen?:
         <span className="shrink-0 rounded-full border border-border bg-card/70 px-1.5 py-0.5 text-caption font-bold uppercase tracking-wide text-muted">
           {review ? `⟳ ${t("groupOverride.reviewKind")}` : t("groupOverride.runKind")}
         </span>
-      </header>
+      </header>}
 
       <div className="flex min-h-0 flex-1 flex-col overflow-hidden px-4 py-3">
         <div className="mb-3 flex items-center justify-between gap-2">
-          <span className="flex items-center gap-1.5">
-            <span className="rounded-full px-3 py-1 text-body font-bold" style={{ backgroundColor: tone.soft, color: tone.color }}>
-              {attempt ? attemptStateLabel(t, attempt.state) : t(`pipelineChipState.${state}`)}
-            </span>
+          <span className="flex min-w-0 items-center gap-1.5">
+            {disclosed ? (
+              /* What the row above does NOT say: which engine and kind ran it. */
+              <>
+                <span className="shrink-0 rounded-full border border-border bg-card/70 px-2 py-0.5 text-caption font-bold capitalize text-muted">
+                  {stage.effectiveRole.engine}
+                </span>
+                <span className="shrink-0 rounded-full border border-border bg-card/70 px-2 py-0.5 text-caption font-bold uppercase tracking-wide text-muted">
+                  {review ? `⟳ ${t("groupOverride.reviewKind")}` : t("groupOverride.runKind")}
+                </span>
+              </>
+            ) : (
+              <span className="rounded-full px-3 py-1 text-body font-bold" style={{ backgroundColor: tone.soft, color: tone.color }}>
+                {attempt ? attemptStateLabel(t, attempt.state) : t(`pipelineChipState.${state}`)}
+              </span>
+            )}
             {verdict ? (
               <span
                 className="rounded-full px-2 py-1 text-label font-bold"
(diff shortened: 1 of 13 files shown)


Latand:

Issue #1602, opened 2026-09-09
Retired seat-tick attempts have a count bound but no age bound, so the wake fence can return after twenty rotations


Latand:
Follow-up to #1594, which introduced the retired-attempt slot.

## What is missing

A retired attempt leaves the project's row only when its holder accounts for it — landed, dropped, withdrawn, or an affirmed absence. The attempts the #1594 incident produced are exactly the ones that never do: a delivery record that ended a send without proving arrival answers `uncertain` for ever, and a holder that cannot be read answers nothing at all. Nothing else ages such an entry out.

That leaves two open ends.

**The bug can recur, further away.** Retention is bounded at `SEAT_TICK_RETIRED_WAKE_LIMIT` (20). A row at the bound refuses further retirement and keeps the fence, which is the right default — an obligation is never discarded to make room — but it is also the pre-#1594 behaviour. A project that rotates its seat twenty times while leaving unresolvable attempts behind stops being woken again, silently, in exactly the way #1594 describes.

**Nothing says the project is approaching it.** The per-attempt board card added in #1594 says an individual attempt is unresolved. No surface says "this project has 18 of 20 retired attempts and will stop waking at 20", so the only warning is the outage itself.

**And every check pays for the whole list.** `reconcileRetiredWakes` asks `sources.wakeState` once per retired attempt, on the five-minute check, indefinitely — up to 20 round trips per project per check, each of which may settle an in-flight operation. Bounded, but permanently non-zero for a condition that is never going to change.

## Shape of a fix

`heldDeliveries` needed the same treatment in #906. Candidates, roughly in order of appetite:

1. An age bound beside the count bound: past some multiple of the wake interval an attempt whose holder has answered nothing conclusive for that long is moved to a terminal, evidence-only record — still readable, no longer polled — rather than kept in the reconcile loop for ever. The evidence must survive; what stops is the asking.
2. A back-off on the ask, so a holder that has answered `unknown` for a day is asked hourly rather than every check.
3. An operator-visible signal when a project's retained count crosses a fraction of the bound, so the return of the fence is announced before it happens rather than discovered as silence.

Whatever is chosen has to keep the #1594 invariants: nothing is dropped that could still be delivered, nothing is ever re-dispatched, and a retired attempt credits nothing whatever becomes of it.

## Not urgent

[...]

Issue #1594, opened 2026-09-09
A wake prepared for a seat that has since been replaced blocks the successor's tick for ever: 34 hours with no wake


Latand:
Refs #1465, #1245.

## What happened

One project's seat tick has delivered no wake since 2026-09-08T04:59Z — 34.5 hours by the time an audit found it on 2026-09-09T15:47Z. The tick itself is healthy: enabled at the default hour, monitor prompt current, the five-minute check running, and every check deciding `wake`. Every one of those wakes is suppressed before it is sent.

The project's accounting row holds an `outstandingWake` prepared 2026-09-08T05:14:24Z for seat epoch 140. The project's active seat is epoch 155, on a different conversation. The attempt carries `operationId: null` and `dispatch.state: "refused"`, and the layer holding it answers `uncertain` — it ended the send without proving arrival either way.

## Why it never ends

`reconcileOutstandingWake` classifies the attempt as `replaced`, and for `uncertain` settles it `row: "keep"` (`src/lib/monitor/seatTickController.ts:459-461`). `keep` is unconditional and has no bound: the holder's answer cannot change, because nothing will look at that send again. The one release path for a `refused` attempt with no operation requires the holder to answer `absent` (`:462-467`), and a holder that holds an unverified record answers `uncertain`, never `absent`.

With the attempt kept, every later check reaches `const withheld = outstanding !== null` and records `deferred-outstanding` without dispatching (`:743-744`, `:759`). No wake for the project, indefinitely. The same wedge is reachable from `too-late`, from `unknown`, and from a holder that cannot be read.

## The conflation

`state.outstandingWake` does two jobs at once:

1. the duplicate fence — do not put a second wake in flight while one may still reach the seat;
2. the ledger — nothing the attempt named is acknowledged until it lands.

Job 2 is right and must not change. Job 1 is only sound while the attempt can still reach the seat the tick is about to wake. Once the project's seat has been superseded by a different conversation at a higher epoch, the kept attempt is addressed to a conversation that is no longer the seat. Two wakes to two different seats are not a duplicate, and holding the successor hostage to its predecessor's unresolved obligation unsends nothing.

## What a fix must not do

- **Not drop the attempt.** It may have been delivered. It keeps its original key, is still asked of its holder at every check, and stays on the board while it is unresolved.
- **Not replay it.** Its payload was prepared for a seat that no longer holds the project; it must never re-enter transport, whatever the holder answers.
[...]

Issue #1465, opened 2026-09-02
Seat tick goes silent for hours: a seat whose lanes have all finished has openWork but no intervalAgenda, and a dropped wake is never re-raised


Latand:
## Symptom

A project's orchestrator seat received no wake for ~2.5 h (12:42Z → 15:16Z on 2026-09-02) with `enabled: true`, `wakeIntervalMinutes: 30`, and work open the whole time. The seat only ran again because a human wrote to it. An earlier gap the same day (09:53Z → 12:38Z) has the same shape. Six lanes reached terminal states inside the window and were harvested late.

## Journal evidence

The tick was checking the whole time — it is not paused, and it is not the seat-busy skip (one `skipped` line in 38). Every other line is `quiet — nothing owed`:

```
14:08:03 quiet    | nothing owed              | reasons []
14:13:03 wake     | the wake interval elapsed while work is open | reasons ['interval'] | items 1
14:18:04 dropped  | the layer holding the wake settled it without delivering it
14:18:04 quiet    | nothing owed              | reasons []
...  (every 5 min, unchanged, through 16:38Z)
```

State row: `lastWakeAt: 12:42:57Z`, `lastCheckAt: 16:38:08Z`, `outstandingWake: null`, `quietSince: 12:47:58Z`.

Board at the time of writing: **0 non-terminal pipelines**, tasks `{inbox: 66, assigned: 26, blocked: 13, done: 118}`.

## Root cause — two defects

**1. A raised wake was dropped and never re-raised.** At 14:13 the tick raised a valid interval wake; at 14:18 the delivery layer "settled it without delivering it". `lastWakeAt` correctly stayed at 12:42, but nothing re-raised it, and the tick then went quiet for the rest of the day. The seat had been seeing `Viewer control did not reconnect after 2 attempts` from its own viewer calls around 11:12Z, 11:29Z and 12:41Z, so the control channel was flaky in exactly this period. A wake the tick knows was not delivered should be re-raised on the next check, not silently abandoned.

**2. `intervalAgenda` ignores every kind of open work except an open lane.** In `src/lib/monitor/seatTick.ts`:

```js
const intervalAgenda = input.pipelines.some(isOpenLane) || input.signals.length > 0;
if (openWork && intervalAgenda && candidates.length === 0) { … interval wake … }
```

[...]

PR #1673, opened 2026-09-13
fix(monitor): a wake the settlement ended unrecorded no longer fences the seat past the journal's own verdict (#1672)

### LINKED ISSUE #1672: fix(monitor): a wake the settlement ended unrecorded fences the manager for ever, and no surface says why
## What happened

The manager seat of one project received no scheduler wake for two days. The tick was healthy: enabled, checking every five minutes, deciding `wake` every time, and journaling `uncertain` + `deferred-outstanding` for every check. Nothing on the board said why beyond "a wake is unresolved under its original key", and no Viewer surface exposed the row, the delivery record or the journal answer behind it, so the diagnosis had to be assembled by hand from the accounting database, the registry file and the runtime journal.

Refs #1465 (the receipt boundary), #1594 (retirement on supersession), #1602 (retired attempts have no age bound), #1612 (lost terminal acknowledgements).

## Mechanism, as reproduced

1. A wake is prepared and sent. The delivery layer reserves the tick's key and claims the attempt, then the runtime-host RPC times out. The layer answers the tick a bare 503 with no operation handle (every structured refusal reaches `deliverConversationMessage`'s caller in that shape; the `transportUncertain` flag is dropped there). The tick records the attempt with `operationId: null` and `dispatch.state: "refused"`.
2. Ten minutes later the tick's own settlement deadline ends the send. The runtime journal holds no record of the operation, so the record is settled `failed` / `unverified` with the unrecorded reason.
3. Every later check reads the record, classifies `unverified` as `uncertain`, keeps the attempt (same seat, so nothing retires it), and defers the project's wake behind it. Nothing ever asks the runtime journal about the operation the record names, because the tick only ever asked it under the handle it never got.
4. The board card says the holder "answered uncertain" and to check the record under the client message id. It does not name the operation, the record's reason, or what would end the fence.

Two classes of attempt are wrongly fenced by step 3, because the journal is never consulted once the record is terminal:

- a send whose journal receipt is `delivered` while the record stayed unverified (the #1612 shape: the terminal acknowledgement was lost);
- a send the host admitted and rejected outright, or that an operator later discarded, whose journal receipt is `rejected` / `failed` while the record was written without a disposition or ended unverified.

And one class is honestly ambiguous and must stay fenced: the journal holds no record (never admitted, or compacted), answers `uncertain`, or cannot be reached. Absence and silence are not proof of non-execution.

## Repair (this PR)

- `wakeStateFromRecord` asks the runtime journal about the operation the **record** names when the record ended without proof. Only the journal's own terminal verdicts change the answer: `delivered` lands the wake and `rejected` / a genuine `failed` release it. `uncertain`, open statuses, no record and an unreachable host leave the record's `uncertain` standing.
- A release is honoured only if the delivery record accepts the journal's verdict: a reservation ended `unverified` absorbs its key for ever, and only one settled `lost` re-arms it on the same-key re-raise. The tick writes the verdict onto the record through the registry's existing outcome API, the way the settlement and the controller's startup reconcile do, and re-reads it; a record that cannot be re-armed (reservation compacted to its owner row) keeps the attempt fenced and says so.
- Every journal line and board card for a kept attempt now carries the evidence: the operation the record holds, the record's state and reason, the journal's answer, and the exits that would end it.
- `GET /api/monitor/seat-tick?project=…&limit=…` reads one project's tick whole and read-only: settings, row, every held attempt with the holder's current answer and its exits, and the project's newest checks. It never ends a send or writes a record.

## What is deliberately not done

- No age bound and no "missing operationStatus means not delivered" rule. The live attempt stays fenced under its key with the row untouched; its permitted exits are supersession of the seat by a different conversation at a higher epoch (which retired three earlier attempts on the same project), or journal evidence that cannot exist any more because the journal has been compacted past the operation.
- The delivery layer's dropping of `transportUncertain` before the tick sees it, and the runtime-host RPC timeouts that started this, are outside the monitor's file ownership and are named for the runtime owner.

Latand:
Closes #1672. Refs #1465, #1594, #1602, #1612.

## What was wrong

A scheduler wake for the manager seat was reserved by the delivery layer and then lost to a runtime-host RPC timeout. The tick's own settlement deadline ended it ten minutes later as `failed` / `unverified` because the runtime journal held no record of the operation, and from then on every five-minute check classified the record as `uncertain`, kept the attempt under its key, and deferred the project's wake behind it. Two days, no wake, and the only board text was that a wake was "unresolved under its original key".

The tick had recorded the send with no operation handle (every structured refusal reaches it as a bare 503), so it never asked the runtime journal about the operation the **record** names. That left two classes of proven outcomes fenced as ambiguous:

- a journal `delivered` the record never learned of (the #1612 shape);
- a journal `rejected` / genuine `failed` — a host that admitted and rejected the send, or an operator discard — beside a record written without a disposition or ended unverified.

## What changes

**`src/lib/monitor/seatTickSources.ts`**
- `wakeStateFromRecord` now asks the runtime journal about the record's operation whenever the record ended without proof, and only the journal's own terminal verdicts change the answer (`journalWakeState`, mirroring `journalVerdict`): `delivered` lands, `rejected` and a genuine `failed` release. `uncertain`, open statuses, no record and an unreachable host leave `uncertain` standing. A `failed` carrying the settlement's own unverified reason stays uncertain (old-host fallback).
- The journal's verdict is written onto the delivery record through the registry's existing outcome API (`settleRecordFromJournal`), so the record and the journal stop disagreeing. The release does not depend on that write taking: a record that cannot take it (reservation compacted to its owner row, or absorbing) keeps its own answer, the evidence says so, and the release rests on the journal's proof.
- **A released attempt is re-raised as a new message** (`releasedWake` on the project row, folded into the next key as `:after-<digest>`). The old key stays bound in the runtime journal to the operation it refused, and the record re-arms that same operation, so the same key could only replay the refusal every check. The replacement carries the same obligations under a key both layers admit; the marker is cleared by the landing that ends it, so an unlanded replacement still replays under its own key.
- The answer now carries its evidence (`SeatTickWakeObservation`): the operation the record holds, the record's state and reason, the journal's answer, and what was written. `wakeRecordPorts({ end: false })` is the inert variant for diagnostics.

**`src/lib/monitor/seatTickController.ts`**
- Journal lines and the unresolved-wake card carry that evidence, and a kept same-seat attempt's card lists what would end it (`seatTickAttemptExits`), including that a discard cannot reach an operation the journal has no record of.

**`src/lib/monitor/seatTickDiagnostics.ts`, `GET /api/monitor/seat-tick?project=…&limit=…`** (new, inert)
- One project's tick whole: settings, row, every held attempt with the holder's current answer and its exits, the project's newest checks. Wake text and the monitor prompt are reported as lengths only. Never ends a send, never writes a record, and peeks the row (`peekSeatTickState`) so a name nobody has ticked mints no accounting row.

## What deliberately does not change

- No age bound and no "no operation status means not delivered" rule. Absence, silence and timeouts are not proof of non-execution.
- The live attempt on the manager's project stays fenced with its row untouched: its record is unverified/unrecorded and the runtime journal has been compacted past its operation, so no journal evidence can arrive. Its permitted exits are supersession of the seat by a different conversation at a higher epoch (#1594, which retired three earlier attempts on the same project) or a delivery-record verdict, both of which this PR reports and neither of which it performs.

## Boundary

Touched: `src/lib/monitor/*`, the two monitor test files, and the new route under `src/app/api/monitor/seat-tick/`. Not touched, and named for the runtime owner: `src/lib/delivery.ts` maps every structured refusal to a bare failure and drops `transportUncertain` before the tick sees it (`delivery.ts:745`), and the runtime-host RPC timeouts that reserved-then-lost the send are the same class that blocked a pipeline stage spawn in the source diagnosis. The tick calls only existing registry and settlement APIs (`recordDeliveryOutcome`, `recordDeliveryOutcomeForOperation`, `journalVerdict`, `sendReceiptFor`, `resolveOriginalSend`, `resolveSendReceipt`), the same ones the settlement and the controller's startup reconcile use.

## Evidence

Isolated, production-shaped: the real delivery record, the real settlement and an in-memory runtime journal side by side, a fresh controller per check (a restart between checks changes nothing).

| scenario | before | after |
|---|---|---|
| RPC timeout after the reservation, journal never holds the op (the live row) | `uncertain`, deferred every check | same fence; journal line and card name the operation, the record's reason and the journal's silence; row byte-identical across checks and two days |
| journal later answers `uncertain` (queue fenced a late admission) or `failed` with the settlement's unverified reason | fenced | fenced |
| journal later answers `rejected`, or `failed` by operator discard | fenced for ever | record settled `lost` on the journal's word, attempt released, same check re-raises the wake under a new key, which lands |
| the re-raise through the real `deliverConversationMessage` → `enqueueStructuredMessage` path over a real `RuntimeJournal` | the same key replays the stored `rejected` (asserted), a changed payload under it conflicts (asserted) | the replacement is admitted as a second operation with exactly one pending `runtime.send` effect; the drain's `delivered` lands it once and clears the marker; no further send |
| journal answers `delivered` (lost acknowledgement) | fenced for ever | landed on the raising plan, record settled delivered, nothing re-sent |
| host admitted and rejected the send outright | fenced from the same check on | released in the same check, next check wakes |
| reservation compacted to its owner row, journal `rejected` | fenced | released on the journal's proof; the owner row keeps its answer and the evidence says so |
| a replacement rejected in its turn | — | released again under a third key, distinct from both before it |
| seat superseded | retired, successor woken (#1594) | unchanged |

Red at the merge base for the released/landed rows (`uncertain` and `deferred-outstanding` where `dropped`/`landed` were expected; card without an operation id); green at this head.

- `bunx tsc --noEmit`: exit 0
- `bun test src/lib/monitor/seatTickController.test.ts src/lib/monitor/seatTickSources.test.ts src/app/api/monitor/seat-tick/route.test.ts`: plus `seatTickState.test.ts`: 245 pass, 0 fail (203 at the merge base for the two existing monitor files); the route test also passes with `LLV_RUNTIME_HOST_SOCKET` set, because it clears the variable before the route is imported
- `bun scripts/privacy-publication-gate.ts --base <merge-base>`: PASS
- The new route run read-only against a copy of the live state reproduces the diagnosis for the manager's project: outstanding attempt with `operationId: null`, `dispatch: refused`, record `failed` unverified with the unrecorded reason, and the same three retired attempts.


Latand:

Issue #1709, opened 2026-09-15
A completed account switch cancels and empties the messages sent for it, including ones held while it waited


Latand:
## What happens

When an account switch completes, the messages the operator sent for that conversation which have not been delivered yet are lost. `commitSuccessor` commits the switch and, in the same registry transaction, calls `terminalizeCancelledMigrationDeliveries(file, conversation, COMMITTED_MIGRATION_DELIVERY_REASON)`. That affects every pending delivery of the conversation:
- it becomes `failed` with "delivery cancelled because its owning account migration committed; send again to authorize a fresh delivery action";
- its `text` is emptied, so the operator cannot even copy it to send it again;
- its delivery operation owner is settled as `lost`, or `unverified` for an uncertain one.

This includes a message the operator sent while the switch was waiting for the running turn. The picker in the kanban board (#1695) tells them "Messages sent meanwhile are held for the switch", but the message is then dropped when the switch completes.

## Evidence (isolated test, #1708 head 04d53556 and main fdc693e1)

Setup, using the account-switch fixture of `src/lib/runtime/structuredAccountSwitch.test.ts`:
1. A structured conversation on `account-a` is running a turn.
2. A `reconfigure` to `account-b` is claimed and waits for the turn (`waiting-turn`).
3. The turn ends, the queue applies the switch again, and it commits on `account-b`.

Both trees give the same result:

| Delivery | Before the switch commits | After it commits |
| --- | --- | --- |
| Sent while the switch waited, never attempted | `held`, attempts 0 | `failed`, `text: ""` |
| Sent before the switch, never attempted | `assigned` to the source generation, attempts 0 | `failed`, `text: ""` |
| Held again after an earlier attempt | `held`, attempts 1 | `failed`, `text: ""` |

Repro for the first row, built on that fixture's helpers (`claudeTranscript`, `recordStructuredHost`, `successorProvider`). Run it by path with an isolated `HOME`, `XDG_CONFIG_HOME` and `TMPDIR`:

```ts
test("a message held while the switch waited is delivered on the successor after the switch commits", async () => {
  const fixture = await switchWaitingForTurn(); // account-a running a turn, reconfigure to account-b pending
  const held = fixture.registry.holdDelivery(fixture.id, "sent while the switch waits", "held-during-switch");
  expect({ state: held.state, attempts: held.attempts }).toEqual({ state: "held", attempts: 0 });

[...]

Issue #1560, opened 2026-09-08
Codex API: Fenced idle history context without a new response


Latand:
Research and exact file manifests: https://github.com/Latand/live-log-viewer-next/pull/1562 (docs/design/codex-api-update/README.md and INTEGRATION.md).

The operator authorized integration. This PR repairs research and probes; product integration remains outstanding. Begin after the deployment release gate and explicit release of the shared host/queue files. The deployment owner retains startup, succession, registry and release behavior. Each slice below needs a separate implementation PR and fresh review at its exact HEAD.

Source baseline: Viewer `88e5a9508be2802266734056d6436f3168201cad`; installed Codex `0.153.4`. Recheck the target base and these file manifests before implementation. Serialize shared-file ownership in the order **pagination → questions → profiles → idle context → diagnostics**. Diagnostics has no semantic dependency on idle context, but shares the host file. Idle context depends on the pagination evidence reader. No runtime file changes belong to the research PR.

Every slice preserves the original operation key, immutable content/attachment/profile snapshot, explicit null/string turn fence, account authority, writer generation and terminal delivery evidence. Timeout or partial history leaves the outcome unknown. Retain unresolved records and blobs; never use a replacement key, queue disappearance, auth refresh or elapsed time as proof of delivery. Native queue dispatch stays disabled in Viewer. No API-key, account or model fallback.

Run only named tests, with private HOME, XDG_CONFIG_HOME, XDG_CACHE_HOME, XDG_DATA_HOME, LLV_STATE_DIR, CODEX_HOME, CLAUDE_CONFIG_DIR, GEMINI_CLI_HOME and a short TMPDIR. Allowlist environment variables; do not copy operator credentials. Use the repository's Bun version and dependency lockfile. Read the installed Next guide before editing a route or component. Add TypeScript checking for each product PR and rendered/browser verification for changed affordances. Product tests listed below are future acceptance work; this research PR has exercised the separate Python probe matrix only.

[...]

PR #1711, opened 2026-09-15
Kanban board, K6c: a completed account switch carries the messages sent for it, in order

Latand:
Part of #1695 and #1709. #1709 stays open after this merges: the review's P1, P2 and P3 are tracked there.

A conversation switched to another account now keeps the messages sent for it. A message the switch provably owns and that never left is carried to the successor with its identity and order, and delivered once by the ordinary migration tick. Nothing is replayed that may already have been delivered. A send made after the switch is ordered behind what it carried on the paths the tests cover; the independent review found two paths where that ordering does not hold (see Review), tracked in #1709. A pipeline stage whose conversation is switched continues on the message held for the switch and settles on the successor's verdict.

## Why
Since PR #759, a committing switch failed every pending delivery of its conversation and emptied its text (#1709):
- a message sent while the switch waited for the turn was never delivered;
- a message sent before the switch and still queued was dropped;
- the operator could not even copy the text to send it again.

A stage switched to another account had nothing to start its successor's turn, so it parked.

## Server
**At commit** (`settleDeliveriesAtCommit`, in the commit's transaction; `commitDeliveryDecision` decides per delivery).

**Carried** when all of these hold:
- unattempted: `attempts` is 0, with no materialized artifact and no retry attempt;
- exactly this conversation's identity;
- owned by this switch:
  - a held delivery has its fence on the committing migration, or K6b's legacy proof;
  - an assigned delivery is assigned to the predecessor generation;
- its owner row exists and has no outcome.

It moves to the successor as `assigned` with text, images, command (operation id, kind, policy, turn id, origin), client message id, digests, attempts and admission order unchanged.

**Excluded**: the switch owns it, but it is not delivered automatically. It ends `failed` **with its payload kept** and a specific reason:

| Case | Disposition, receipt |
| --- | --- |
| attempted (attempts > 0, an artifact, a retry attempt, or uncertain) | `unverified`, "verify-first" |
| bound to the previous turn (a string `turnId` under a policy other than `interrupt-active`) | `lost`, "safe" |
| injected context (#1560) | `lost`, "safe" |
| `ephemeral-*` attachments only the client holds | `lost`, "safe" |

A `turnId` under `interrupt-active` only names what to interrupt, so that message is carried.

**Left exactly as it is**: nothing proves the switch owns it. That covers:
- an alias identity;
- a fence naming another operation;
- a legacy held record without proof;
- an older generation;
- a missing or settled owner.

These are release limitations (below).

**Order.**
- `HeldDelivery.admissionSeq` is additive and optional. It is assigned in the reservation's own serializable transaction, as the conversation's highest sequence plus one, and replays keep it.
- `compareDeliveryAdmission`: sequenced records by sequence. Records without one order first, and among themselves by `createdAt`, then id, as before.
- The code this builds on keeps the field on records it rewrites, because its normalizer spreads the stored record.

**Ordering behind carried messages (not complete; see Review).**
- `beginDeliveryAttempt` refuses a claim while an earlier-ordered delivery of the same conversation is `assigned` to the same generation.
- A claim alone does not decide journal order. So the structured send (claim through the journal's answer), the migration drain (claim through delivery), the legacy send (the whole send) and the HTTP retry (claim through re-admission) all run inside `withConversationActuation`, one section per conversation, per process, re-entrant.
- A later send waits there until the earlier command is admitted, on the paths the tests cover.
- It does not hold everywhere. A migration tick requested from inside a section inherits the section's context, so its drain can claim past another send's section (review P1). A later send is refused while an earlier row waits in a `failed-recoverable` switch, which the tick does not drain, so it can stay `held` (review P2).
- Refused claims keep today's handling: the sends requeue and answer `held`, the drain skips, and the retry answers 503 retryable.

**Unchanged recovery** (reported exactly as it behaves, with tests):
- a `lost` record resent under the same key and payload re-arms the same record and operation;
- an `unverified` record stays failed;
- a new key is a new message, ordered after the rest;
- nothing re-arms an excluded record automatically, and the composer outbox never replays a held send locally.

## Kanban board
While a recorded switch waits, the picker says:
- "Messages sent now are held and delivered after the switch, in the order they were sent. Cancel delivers them on the current account instead."
- "Not carried: a message bound to the current turn, injected context, or attachments only the sending browser holds. Each ends failed with its reason, keeps the text the Viewer holds for it, and its receipt says whether sending it again is safe."

Both are in English and Ukrainian.

## Release limitations
- **Ordering is not guaranteed everywhere yet** (review P1): a tick requested inside a send's actuation section can let a drain claim ahead of that send. Open in #1709.
- **A later send can stay held in a `failed-recoverable` switch** (review P2): the tick does not drain that phase, and the claim gate refuses the later send behind an earlier row still `assigned` there. On main such sends were delivered. Open in #1709.
- **Deliveries a committed switch cannot attribute** (see above) stay held or assigned:
  - they are not delivered, failed, deleted or cleaned up silently;
  - they still count toward the conversation's limit of 100 pending deliveries;
  - they keep a rolled-back block from clearing (K6b);
  - they do not hold back new sends;
  - recovery is an explicit send.
- **Ordering holds within one Viewer process.** During a deploy's overlap, two Viewer processes can each actuate the same conversation.
- **A crash between a claim and its journal admission** leaves an uncertain delivery that gates nothing, so a later send can reach the journal first. The uncertain one is settled by reconciliation and never replayed.
- **The runtime host's startup continuation** for an interrupted Codex turn has no reservation and is not ordered by this.
- **An excluded attempted delivery** keeps its failed record even if its earlier journal command is later confirmed. That is the existing cancellation contract, and its receipt says "verify-first".

## Evidence
All tests run by path with isolated HOME, XDG, state and TMPDIR, one DOM file per process.
- **`structuredSwitchCommitMessages.test.ts` (8).** The first cases were red on main f1396da9.
  - Messages sent before and while the switch waited are carried with every identity field and their sequence. Two ordinary migration ticks deliver them to the successor once each, in order, and each owner settles once.
  - Excluded messages end failed with their payload kept, and their receipts say "verify-first" or "safe". The interrupt-bound message is carried.
  - Foreign fence, unproven legacy hold, older generation and settled owner stay byte-equal through commit and tick.
  - Explicit recovery: a `lost` record is re-armed under its key, an `unverified` one is not, and a new key is new.
  - Same-millisecond admissions, with the clock frozen and descending UUIDs, keep their order; legacy records order first, by time.
  - The claim gate: refused behind an earlier assigned delivery of the same generation, allowed once that one began its attempt, and never gated by held or older-generation rows.
  - A send after the commit, over a real runtime journal, is held behind the carried message and reaches the journal after it.
  - **Paused actuator:** the drain claims the carried message and pauses before its journal admission. A structured send made meanwhile stays out of the journal. The journal's admission order is then [carried, later].
- **`deliveryActuation.test.ts` (3):** serial per conversation, independent across conversations, re-entrant, released on throw.
- **`delivery.test.ts` (41):** a legacy send waits for the conversation's section, and is held behind an earlier waiting admission without actuating.
- **`engine.test.ts` (289), the integration:**
  - an isolated registry switches a stage's conversation through `applyStructuredReconfigure` while its turn runs;
  - the operator's message held for the switch is carried at commit and delivered by `reconcileMigrations`, and that delivery starts the successor's turn;
  - the stage passes on the successor's verdict, with the same conversation id, no fail edge, no recovery stage and one spawn.
- **`structuredDelivery.integration.test.ts` (38):** with the production held-message port, a real journal and fake engine hosts, the published Codex and Claude successors each receive the held message once under its original operation, and the source receives nothing.
- **Pins that asserted commit-time cancellation now assert carrying** (`registry.reseat` 24, `structuredMessageDelivery.accountReseat` 25). The K6b cancel fixture attempts its earlier message first (`structuredSwitchCancel` 23).
- Sweep: all 52 test files that touch the delivery reservation, its actuators or the migration drain were run by path, isolated.
  - Green: every file except the known ones below. That includes `coordinator` 106 passing, `bindings` 61, `http` 34, `startup` 87, `structuredMessageDelivery` 61, `structuredHostRetirement` 51, `seatTickController` 134, `flows/engine` 56, `stageHostGenerationClose.integration` 44 and `sendSettlement` 25.
  - Failures identical on the unchanged main tree in this environment:
    - `coordinator`: 9 board-continuity tests;
    - `failedSpawnDelivery`: 7;
    - `registry.sqlite`: 4 timing diagnostics (its first-owner insertion-order test failed once in a whole-file run, then passed three single runs and a second whole-file run);
    - `reaperRuntime`: 3;
    - `severedHostReap`: 2;
    - `structuredSpawn.integration`: 2;
    - `spawnProjection`: 1;
    - `legacyClaudeRecovery.cleanCi.integration`: 1.
  - The 7 coordinator tests and the one MCP recovery test that newly failed pinned commit-time cancellation or an arbitrary admission order. They now pin carrying, and the MCP fixture admits its attempted send first.
- Red checks. Each mechanism was removed in turn and its named tests failed, with the tree restored after each:
  - the commit failing everything again (6 tests, and the stage integration);
  - no claim gate;
  - no section in the drain; no section in the structured send (the paused-actuator test);
  - order by time only; no sequence assigned;
  - carrying turn-bound, injected, request-local or attempted messages; excluding interrupt-bound ones;
  - a foreign fence counted as owned; a settled owner carried; an older generation counted as owned;
  - an exclusion emptying the text; an attempted exclusion marked `lost`;
  - the legacy send outside the section;
  - the section not serial; the section not re-entrant.
- `tsc --noEmit` exits 0. `bun run build` exits 0, with only the known `tts-karaoke` CSS warning.
- The browser test (`issue1695Accounts.browser.test.tsx`) passes against the served prototype and regenerated `k6.json`.
  - Its recorded-switch frame (not committed, under `.artifacts/issue-1695/`) shows the pending row, Cancel switch and both notes.
- The local privacy gate passes, with commit checks.

## Review
- **Independently reviewed:** head 674ddf07 (tree 09eb4af4) against base f1396da9, one round. Verdict **APPROVE, no P0**, under the P0-only rule: nothing is lost or delivered twice.
- **Validation in the review:** `tsc` exits 0. Under Bun 1.4.0, the focused files are green on the head. `coordinator.test` has the same 9 board-continuity failures on head and base, and `registry.sqlite` the same 3 timing-diagnostic failures on both (62 pass).
- **Findings, open and not fixed in this PR** (recorded with acceptance tests in #1709):
  - **P1:** the actuation section decides "already inside" from inherited async context. A migration tick requested inside a section (a legacy send holding it while a switch is pending requests one) inherits that context, and its drain skips the section, so a later send can reach the journal first. Reproduced under Bun 1.4.0 with the real section, tick signal, controller and drain; the control, with the tick requested outside, stays in order.
  - **P2:** the claim gate refuses a later send while an earlier row is `assigned`, and the tick never drains a `failed-recoverable` conversation, so the later send stays `held` after two ticks and a never-claimed reservation blocks every later claim. Reproduced on this PR's own fixture; on main those sends were delivered.
  - **P3 (over-built):** the process-global AsyncLocalStorage section exists only because the legacy send holds it for the whole send; one slow send also holds up that tick's migration work for other conversations. Narrowing the legacy section to claim-to-actuation removes nesting and lets a plain per-conversation chain replace AsyncLocalStorage.
- The body's earlier unconditional "cannot overtake" wording is corrected above.

## Ledger and plan
- `parity-ledger.md`:
  - "Messages across a completed account switch" and "A reseated stage conversation settles on its own verdict" close with K6c;
  - a new open row, "Deliveries a completed switch cannot attribute", carries the release limitation.
- `k6c-plan.md` records the design as implemented, the evidence it relies on, the exact recovery behaviour and the release limitations.

No deployment. No real account was switched, and no runtime production state was used.


Latand:
Closed after the backlog audit in #1738: this branch is wholly contained in #1713 (git merge-base --is-ancestor), which is the one to review.

PR #1714, opened 2026-09-15
Kanban board, K7a (#1695): an operations feed and a pipeline creation receipt for the manager's arrows

Latand:
## What this adds

The board can now show what the manager's `create_pipeline` and `update_task` calls are doing. The data comes from the MCP receipt rows that already decide those calls' idempotency, and from the pipelines their creations stamp, so the arrows and the creation animation (K7b, #1695 AC16–17) have real events to follow.

### Operations feed: `GET /api/mcp/operations`

Query: `project=<key>&after=<sequence>&limit=<1..50>&unresolved=<seq,…>&creationsAfter=<cursor>`

- **Row fields:** `sequence`, `tool`, `requestDigest`, `state`, `claimedAt`, `callerConversationId`, `callerProject`, `target`, `refusal`.
- **States:**
  - `pending`: claimed, with no result, within 120 s of the claim (four tool-call deadlines).
  - `unknown`: still no result after that, or a result that cannot be read.
  - `accepted`: a recorded success, or a stored pipeline stamped with the row's digest.
  - `failed`: a recorded refusal, shortened to 200 characters.
- **Target:** `{ project, taskId, pipelineId }`, or `null` when the target is unknown.
  - Pending, unknown and refused rows carry the target recorded when the call was claimed: the board task the call names (`update_task.taskId`, `create_pipeline.taskId`), validated against the task store, with that task's project.
  - Accepted rows carry what the result or the stamped pipeline names.
  - A call naming no task, or a task that does not exist, reads `null`. No pipeline id is guessed.
- **Scope:** a row belongs to a project by its recorded caller, its claimed target, its result's target, or a pipeline stamped with its digest. Stamped pipelines are applied before paging.
- **Paging:** without `after`, the newest rows answer. The response's `after` is passed back unchanged and always advances, so a page never waits on an unresolved row.
- **Unresolved rows:** a reader sends the sequences it still holds as pending or unknown in `unresolved` (at most 50 per request; it rotates if it holds more) and reads their current state in `refreshed`, with the same sequence and digest.
- **Stamped creations:**
  - The response's `creationsAfter` is passed back unchanged.
  - Each page answers, in `refreshed`, the next 50 or fewer creations stamped into the project after that cursor, in commit order, with `creationsHasMore` while more follow.
  - A creation proven only by its pipeline therefore arrives however late it was stamped and however long the reader was away. A reader without the cursor starts at the newest stamp.
- **Read only:** the route opens the receipt database read-only through a small shared opener, and a missing database answers an empty page. Nothing is claimed, settled or replayed, and no arguments or result bodies are returned.
- **Recorded at claim:** for those two tools the claim row holds the server-derived caller (kind, conversation, project) in `caller_json`, resolved like original-key recovery's caller, and the validated target in `target_json`. Both columns are added in place, and older rows read with neither. A resolver fault records nothing and never refuses the call.

### Creation receipt: `Pipeline.creationReceipt?: { tool, requestDigest, callerConversationId, claimedAt, recordedAt }`

- It is stamped in the create transaction from the receipt the MCP service claimed for the call. Its digest equals the feed row's `requestDigest`, so a placeholder and the card it becomes share one key.
- `recordedAt` is always later than every stamp already stored. Pipeline mutations are serialized, so stamps are ordered as they commit, and a stamp that commits after a reader's cursor is always after it.
- A create under a digest that a stored pipeline already carries answers with that pipeline and stores nothing new.
- `create_pipeline` accepts an optional `taskId`. It joins `taskIds` and is checked like any other task link.

## Compatibility

Everything stored is additive and optional:
- Older builds keep `creationReceipt`: their validators check only the fields they know, and their readers spread the record.
- Older builds never read `caller_json` or `target_json`, and their inserts name their own columns.
- The receipt store's claim answer is unchanged.

## Scope

- This PR ships the K7a backend only.
- The kanban arrows and creation placeholder (K7b) consume it in a separate change, and the frontend owner has the current contract.
- No UI, CSS, fixture or ledger file is touched.
- Discovery reaches the pipelines still on the board. Settled pipelines are archived after three days, together with their cards.

## Review rounds

**Round 1** (fixed in `ea33a39a`):
1. Pending rows lacked a target: claims now record the validated target.
2. An unknown row that settled after the cursor passed was lost: sequences can now be reconciled.
3. A creation proven only by its pipeline was filtered out of that pipeline's project: stamped pipelines now scope rows before paging.

**Round 2** (fixed in `9083930b`): creation discovery was a ten-minute, newest-fifty refresh, so an overflowing batch or a late stamp could be missed forever. It is replaced by the commit-ordered `recordedAt` stamp and the resumable `creationsAfter` cursor.

## Verification

Every run is by path, under Bun 1.4.0, with isolated `HOME`, `XDG_CONFIG_HOME`, `LLV_STATE_DIR` and `TMPDIR`.

| File | Result |
| --- | --- |
| `src/lib/mcp/operationsFeed.test.ts` | 12/0 |
| `src/app/api/mcp/operations/route.test.ts` | 2/0 |
| `src/lib/mcp/server.test.ts` (column pin updated) | 76/0 |
| `src/lib/mcp/schemaParity.test.ts` | 21/0 |
| `src/lib/pipelines/engine.test.ts`: the creation-receipt and stamp-order tests | 2/0 |
| `src/lib/mcp/bindings.test.ts`: the operations-feed target test and `create_pipeline` tests | 4/0 |

What the feed tests cover:
- held bindings expose their targets while pending and unknown;
- more than 50 unresolved rows settle observably while newer rows arrive;
- 52 creations with null or refused receipts and no caller or target, stamped into a project after its cursor passed their rows, in the reverse of their claim order, with the reader back 20 minutes later and one more stamped a day later. All are discovered with unchanged sequence and digest, 51 over two resumable pages, while newer rows page, and no other project sees any;
- a reader without the creations cursor starts at the newest stamp.

`tsc --noEmit` is clean, and the privacy gate passes with known-value fingerprints and commit checks.

**Red checks.** Removing each mechanism turned its test red: 8 in the first commit, 10 in round 1 and 7 in round 2. The two round-1 checks on the ten-minute refresh covered code that round 2 replaced. The 7 from round 2:
- the stamp follows the clock instead of the stored stamps;
- no stamp is recorded;
- discovery is not resumable;
- the discovery page is unbounded;
- the creations cursor does not advance;
- a fresh reader sweeps every past stamp;
- discovery ignores the project.

Part of #1695.


Latand:
Closing: the operations feed has no reader (K7b never started) and conflicts in four files; it comes back as the first commit of K7b if the arrows are scheduled.

PR #1733, opened 2026-09-18
Kanban: stop the board jumping while the seat's composer grows (#1731)

### LINKED ISSUE #1731: fix(kanban): the board jumps under the seat on every typed or dictated character (scroll anchoring on the fixed-height seat)
On the kanban page, while typing or dictating into the manager seat's composer, the whole board below the seat jumps up and down by a few pixels on nearly every character. Dictation makes it continuous because the realtime transcript re-wraps the draft on every revision.

## Cause (reproduced in the issue1695 browser harness, no live state)

- The seat has a fixed height (`.kb .seat`, `src/components/kanban/kanbanBoard.css`) and the composer form sits at its bottom. `useAutosizePinned` sets the field height on every value change; inside the fixed box, a growing field moves the form's top and the feed above it upward.
- The page scroller (`.kb .kb-page`) picks a scroll anchor inside that moving region and compensates `scrollTop` (Chrome scroll anchoring). Everything below the seat moves by the field's height delta, once per height change. `scrollHeight` never changes, so it reads as a jump, not a reflow.
- Controls: blurring the field does not stop it (20 swings); `overflow-anchor: none` on the scroller stops it completely (0 swings) with the field still growing. Both the compact seat and a grip-expanded seat reproduce when the page is scrolled so the composer is the topmost visible content.

## Fix

- `.kb .seat { overflow-anchor: none; }` — scoped to the element that moves; the board keeps its own anchoring. Verified 0 swings at 1440×900 and 1100×800.
- A browser test beside the existing `issue1695*.browser.test.tsx` files on `serveEvidenceFixture` / `openFixture`: scroll so the seat header is out of view, drive the composer value across its wrap boundary N times, assert the board frame's viewport top and the scroller's `scrollTop` are unchanged while the field height does change. The red path is provable by clearing `overflow-anchor` in the same test. happy-dom does not implement anchoring, so a `.dom.test.tsx` cannot hold this.
- Coverage: 1440×900 and 1100×800; compact and grip-expanded seat; at least one scroll position where the composer is the topmost visible content. Check the tabbed layout (≤ 900×700) and the phone surfaces (`MobileFocusView` / `MobileBoard`) on their own terms and report; they use a different scroller and composer.

Not in scope, worth its own issue if confirmed: in the compact seat the composer's grow ceiling computes to the 44 px one-row floor, so the seat composer cannot grow past two rows without dragging the grip.

Latand:
Closes #1731.

## What moved, and why

On the kanban page the whole board below the manager seat moved by a few pixels on nearly every typed character, and continuously under dictation, whose realtime transcript re-wraps the draft on every revision.

The seat has a fixed height and the composer sits at its bottom, so a growing field moves the form's top (and the feed above it) upward inside a box whose own size never changes. Chrome's scroll anchoring on the page scroller `.kb .kb-page` picked an anchor inside that moving region and compensated its `scrollTop`; everything below the seat was displaced by the field's height delta. The scroller's `scrollHeight` never changes, which is why it reads as a jump rather than a reflow.

## The change

One seat-scoped declaration in `src/components/kanban/kanbanBoard.css`:

```css
.kb .seat { overflow-anchor: none; }
```

The seat stops being an anchor candidate; the board keeps its own anchoring. `useAutosizePinned`, `useComposerBox` and `composerScroll` are untouched — the field's growth is correct and unchanged.

## The check

`src/components/kanban/issue1731SeatAnchor.browser.test.tsx`, beside the existing `issue1695*.browser.test.tsx` files on `serveEvidenceFixture` / `openFixture`. happy-dom implements no scroll anchoring, so a `.dom.test.tsx` cannot hold this.

Each case scrolls the page until the seat header is out of view and the composer is the topmost visible content (asserted with `elementFromPoint`), finds the composer's wrap boundary at the field's real width, drives the draft across it 12 times, and requires the board frame's viewport `top` and the scroller's `scrollTop` to be unchanged while the field's height does change.

Coverage: 1440×900 and 1100×800, the compact seat and one dragged open by its grip — four cases, `0` swings each, field 28 px ↔ 44/46 px. Measurements: `evidence/issue-1731/seat-anchor.json`.

Red path, twice over:
- inside each case, `overflow-anchor: auto` back on the seat inline and the identical drive: 12 swings in all four cases, so a green result cannot be an inert check;
- with the declaration deleted from the stylesheet, the test fails all four cases (`the seat's overflow-anchor is auto, not none`, `the board frame moved`, `the page scroller moved`).

Run it with:

```
LLV_KANBAN_BROWSER_TEST=1 bun test src/components/kanban/issue1731SeatAnchor.browser.test.tsx
```

## The other surfaces, checked and left alone

Both were probed in the same harness; neither gets a change here.

**Desktop tabbed layout (≤ 900×700).** At 900×700 and 800×700 the board is in `tabs` mode and the seat starts collapsed, so there is no composer until the operator opens it. With the seat opened and anchoring forced back on, the drive was swept across every scroll position from 0 to the scroller's maximum: `0` swings everywhere, at both sizes. The control run in the same script at 1440×900, same technique and a read-back of the computed `overflow-anchor` it overrode, swings 1–6 times at three of those positions — so the negative is the layout, not a blind probe. At these sizes the board frame is its 440 px minimum rather than a full scrollport, which is the difference. The shipped rule is the same `.kb .seat` element and applies here too; no separate fix is warranted.

**Phone (`MobileFocusView` / `MobileBoard`), 390×844.** `MobileBoard` is a scroller with no composer in it at all. In the focus view the composer is a flex sibling of the transcript feed, not a descendant: the only scrolling ancestor of the field is the form itself (0 px scrollable), the document does not scroll, and `[data-log-feed-scroller]` (1602 px scrollable, parked mid-feed) does not contain the composer, so the composer cannot be an anchor candidate for it. Driving the field 32 px ↔ 54 px left the feed's `scrollTop`, its first item's position and the bar exactly where they were. Nothing to fix on their own terms.

## Verification

- `LLV_KANBAN_BROWSER_TEST=1 bun test src/components/kanban/issue1731SeatAnchor.browser.test.tsx` — 1 pass, and red as described above with the rule removed.


Latand:
Follow-on filed from the note in #1731: #1734 (the compact seat's composer ceiling). Not touched here.

CI note for the reviewer: `bun-runtime` went red on its first run in the native-runtime step (`Codex app-server child exited: could not find bubblewrap on PATH`, plus a 30 s timeout in the same file). That job is red on this PR's merge base too (main @ b3b7434, same step, a different test in it). It passed on rerun at the same commit; nothing in this PR touches that surface.

PR #1782, opened 2026-09-18
Pipelines survive a deploy and a failed spawn: a cut turn resumes, a refused spawn retries (#1747, #1750)

### LINKED ISSUE #1747: fix(runtime): a deploy cuts a Claude turn mid tool call and nothing resumes it — the stage reads running and the card working for ever
## Observed (2026-09-18, production deploy at 06:06 UTC)

Three Claude stage conversations were mid tool call (a Bash command each) when the promote severed their hosts: two pipeline builder stages and one standing builder. The hosts were re-hosted by the new release, but the interrupted turn was never continued. For the next hour:

- each transcript's last record stayed at 06:06, the instant of the cut;
- the pipelines kept the attempts in `running`, the board kept showing the cards as working, and nothing parked or flagged them;
- three lanes spawned after the deploy worked normally, so the hosts and the account were healthy.

The operator noticed by eye and asked the manager to look. A queued "continue" message from the manager was the only thing that could move them.

## Why

`src/lib/runtime/startup.ts` continues an interrupted **Codex** turn at adoption (`INTERRUPTED_CODEX_CONTINUATION_TEXT`, "Continue the interrupted turn from the transcript."). No equivalent exists for Claude rows: a Claude host severed mid tool call comes back idle, waiting for input that nobody sends. Background commands and monitors the agent was waiting on die with the old process, so even an agent that was only waiting never wakes.

## Expected

1. Adoption after a release change continues an interrupted Claude turn the same way it does for Codex: one bounded continuation per severed turn, idempotent by the severed turn's identity, telling the agent that background work started before the cut is gone.
2. A stage attempt whose conversation has had no transcript record and no turn in flight for longer than a bound, while the attempt is `running`, is surfaced: the card stops reading as working, and the pipeline records a stall with its age (it may park for a decision after a second bound).
3. Tests by path at the adoption seam: a Claude row severed mid tool call receives exactly one continuation after adoption; a row that finished its turn before the cut receives none; a replayed adoption sends no second continuation.

Related: #253 (deploys kill pane-less agents), #1732 (re-host without MCP), #1441 (settlement and background tasks), #1744 (graph node hides post-settlement work).

### LINKED ISSUE #1750: bug(runtime): after a promote, direct spawn_agent leaves a Claude process idle with its first prompt expired or held, ownership stuck synchronizing, and kill unavailable
Reported by another project's manager seat, 2026-09-18, the morning of a production promote (the Viewer restarted around 06:06 UTC). Looks like the family of #1583 (closed), seen again on a fresh spawn rather than on a live seat.

## Observed

Two direct `spawn_agent` calls each started a Claude process that never received its initial prompt:

- launch A: lifecycle `delivery_expired … failed after 0 attempt(s)` 25 s after the spawn;
- launch B: `delivery_held`, never released.

For both, `conversation_deliverability` answered: `hostStatus: idle`, `pendingAction: spawn`, `processRecorded: true`, `condition: synchronizing`, reason "the durable host record is still publishing ownership". `conversation_action kill` answered "structured runtime host is unavailable", so the two idle Claude processes could not be stopped through the Viewer at all. The same prompt launched through `create_pipeline` started within seconds.

## Expected

1. A direct spawn after a promote publishes ownership and delivers its first prompt, or fails the launch visibly within a bound and reaps the process it started; a process the Viewer started is never left running with no prompt and no way to stop it.
2. `kill` on a conversation whose host record names a recorded process works while ownership is still `synchronizing`, or says which release owns the process and routes there.
3. Establish why the pipeline launch path publishes ownership where the direct spawn path stalls (same prompt, same account, same minute), and cover the difference with a test by path at the spawn admission seam.

Related: #1583, #1747 (turns cut by the same promote were never resumed), #253.

Latand:
Closes #1747. Closes #1750.

## What went wrong, from the production record

Diagnosed read-only from the pipeline store, the registry receipts, the deployment ledger and the stage transcripts of 2026-09-18 (three deploys, nine lanes). Five root causes, all in the pipeline controller and its spawn path.

**The cut turns (#1747).** A release succession re-claims each registry row and starts a brand-new CLI child over the resumed session. The tool call the previous process was inside is gone, and so are the background commands, monitors and shells it was waiting on; the new child blocks on stdin with nothing pending. The registry keeps the turn `busy` and the runtime ledger keeps reading `running`, so `conversationAgentActive` answers `true`, the attempt stays `running` and the card stays working. Twelve attempts had an open activation spanning a succession; every one went silent about three and a half minutes before the succession instant and the first record after the gap was a message somebody queued by hand — in the worst lane, fifty-one minutes later. No continuation exists for a Claude row: the interrupted-turn continuation at adoption returns nothing for a non-Codex conversation, and the other recovery path iterates orchestrator seats, which a stage builder is not. The container log for the succession carries zero recovery lines.

**The refused spawns (#1750).** Eight attempts parked on a transient class and every one succeeded on the operator's first manual retry.

- Six parked on `pipeline structured runtime host is unavailable` with one spawn call each, no booked wait and no retired launch — while their receipts were `failed` on disk within two seconds. The default ports memoise one registry projection per tick and invalidated it only after `spawnPipelineAgent` resolved, so on the throw path the catch read a projection captured before the launch was reserved, saw `null` for the receipt, and rethrew into the park. The ten-minute host budget existed and was never once entered.
- One parked on `account mutation is busy` after three retries over 146 seconds: that class took the delivery controller's thirty-second budget, which is wall-clock from the activation and therefore spent inside the first spawn calls. A deploy runs about eight minutes end to end and its succession contends for the lock throughout.
- One parked on `delivery was started by an earlier executor; whether it reached the recipient is unverified`, which was not classified as transient at all — although that string is written exactly when a succession leaves two executors over one journal, and the receipt named an artifact path that was never created, so no turn had started.

## The fix

**One continuation per succession, from the controller.** An attempt records the runtime host epoch its launch was hosted under. A later tick that finds a different epoch is looking at a succession — the only event that cuts a running turn while leaving its host alive and idle — so no CPU heuristic is involved. On the first sighting the controller records the transcript's newest record; while that has not moved, the turn reads `busy` and no delivery is outstanding, it sends one continuation after a three-minute silence, attributed to the controller (`origin: agent/controller`) and telling the agent that a deploy cut the turn and that in-flight background work is gone. The client message id is stable per pipeline, stage, attempt and epoch, so the delivery queue's own dedupe stands behind the record. An attempt still silent ten minutes after that continuation parks with a reason that says exactly that.

Every refusal the issue asks for is the same comparison: any record written since the sighting — a resumed tool call, a prompt somebody else delivered, the agent's own answer — means a turn is in progress, and the attempt adopts the new generation and is left alone; an outstanding delivery defers; a witness that already carries its continuation never mints a second.

**Transient spawn classes.** The registry projection is now invalidated on the throw path as well, which alone restores the retired-launch retry and the ten-minute budget for the six host-unavailable parks. The busy-account class rides that same budget — it is the succession seen from another side — while a delivery controller between publications keeps its thirty seconds, because that gap really does close in seconds. The unverified-delivery class is admitted as transient and routed through the receipt gate rather than the immediate handshake retries, and it is re-dispatched only where the transcript the receipt named does not exist; where it exists the attempt parks rather than deliver a second prompt. A non-transient failure still parks on the first sighting.

## Evidence from this branch's own lane

The deploy at about 19:18Z cut this builder's turn in the middle of a tool call — the defect this PR fixes, on the pipeline writing the fix. The attempt kept reading `running`, the transcript stayed silent at the instant of the cut, and the work resumed only when the orchestrator sent a continuation by hand. With this change the controller would have sent that same continuation itself, once, three minutes in.

## Verification

- `bun run build` — clean, including the MCP bundle.
- `bun test src/lib/pipelines/severedTurnResume.test.ts src/lib/pipelines/engine.test.ts src/lib/pipelines/durableEvidence.test.ts` — 326 pass, 0 fail, on an isolated state dir. Also green by path: `controller`, `stageCompletion`, `stageHostLiveness`, `terminalReap`, `graphEdits`, `store`, `listProjection`, `verdict`, and the 44-test `stageHostGenerationClose.integration` suite.
- Red before the fix, checked in a detached worktree at the merge base: all six severed-turn cases, and four of the five new spawn cases. The fifth — a non-transient failure parks at once — passes at the merge base by design: it is the guard that the new transient classes did not widen into everything.
- `bun scripts/privacy-publication-gate.ts --base <merge base>` — PASS.
- `severedStageRetry.test.ts` keeps two failures in this sandbox; both fail identically at the merge base (the kill operation cannot settle without a runtime host socket) and neither touches this change. `pipelineStageHostAccess.integration.test.ts` was not run: it requires an authenticated credential and spends real turns.


Latand:
Review round 1 fixed in 7d6ffa60f733f4a0eb4c2b7257b05665b4e6d010.

The finding was exact: the sighting of a new host epoch is up to one tick later than the succession, and the silence was measured from that sighting, so a record written inside that window — the runtime's own interrupted-turn continuation, or an operator's continue — became the baseline. An agent that really had resumed could then be sent a second continuation, saying its in-flight work was gone, in the middle of a live turn.

The witness now arms only where the transcript had already been silent for the whole resume bound when the new epoch was first seen. That is what makes its newest record older than the epoch change rather than merely older than the sighting. A transcript that moved more recently, or one with no timestamp to judge, gets no continuation at all: the attempt adopts the generation and is left alone however long it then sits inside one tool call. Every other refusal is unchanged.

The suite's clock is now the production one — last record 15:59:11, succession 16:02:32 — so the reported shape still receives exactly one continuation. The six existing severed-turn cases stay green at the new fixture and at the previous head; the new case, an attempt whose transcript moved thirty seconds before the sighting and then went quiet past both bounds, is red at 4acb2733 and green here. tsc, bun run build, 347 tests by path on an isolated state dir, the privacy gate from the merge base, and CI at this SHA are all green.

PR #1791, opened 2026-09-19
Roles: every scaffold carries one standing process-cleanup rule (#1770)

### LINKED ISSUE #1770: roles: a worker cleaned up by port and killed an unrelated local server — scaffolds must limit process cleanup to PIDs the worker started
Filed by the orchestrator seat, 2026-09-18.

A read-only research stage (architect role, pipeline `1240e51e`) started loopback stub servers for a probe, then cleaned up by port. One of its early stubs had targeted a port that an unrelated local server of the operator's was already listening on, so the port-based cleanup killed the operator's process. The agent reported it itself as a P1 finding; the service had to be started again by hand.

`access: read-only` governs repository mutation only, so nothing in the stage's contract spoke to this.

## Wanted

- Every role scaffold carries one standing rule: a worker stops only processes it started itself, by the PID it recorded at start; never by port, name or pattern. A port that is already in use is a reason to pick another port.
- Probe servers bind an ephemeral port (port 0) and read back the assigned one.
- Optional, if cheap: the `restricted` sandbox denies signalling processes outside the stage's own process group.

Latand:
Closes #1770.

A read-only research stage started loopback stub servers for a probe and cleaned up by port; one of those ports belonged to an unrelated local server, which the cleanup killed. `access: read-only` governs repository mutation only, so nothing in the stage contract spoke to it.

## What changed

- `src/lib/roles/defaults.ts`: one `PROCESS_CLEANUP_RULE`, defined once, appended to all eight role scaffolds (orchestrator, reviewer, verifier, builder, architect, cleaner, prod-auditor, deployer). It says: stop only the processes you started yourself, each by the PID recorded at start; never by port, name or pattern (no `fuser -k`, no `lsof` piped into kill, no `pkill`, no `killall`); a port already in use is a reason to pick another port; a probe or stub server binds port 0 and reads the assigned port back. `PROCESS_CLEANUP_MARKER` is the stable phrase the test pins, so the wording around it stays free to change.
- `src/lib/roles/registry.test.ts`: one test walks the whole registry and asserts each role's rendered scaffold carries the marker, the pattern-kill prohibition and the port-0 instruction.

## Delivery

Spawns render the scaffold through `resolveRole` and pipeline stages through `pipelineRoleLookup`, and both go through the same `roleScaffoldBody`, so both receive the rule. Nothing caches or versions a scaffold: a pipeline freezes the rendered text into its stage record at create time, so new pipelines and new spawns get the new text and there is no version to bump.

## Out of scope

The issue's optional third bullet — a `restricted` sandbox that denies signalling outside the stage's process group — has no existing seam in this repo (no sandbox profile carries a signal policy today), so it is not a few lines here. Left for a follow-up issue.

## Verification

- `bunx tsc --noEmit --incremental false` — exit 0, no output.
- `bun run build` — exit 0 (Next build plus the MCP bundle).
- `bun test src/lib/roles/ src/lib/pipelines/roles.test.ts src/app/api/roles/route.test.ts src/lib/pipelines/prompts.test.ts src/lib/workflows src/lib/flows` under an isolated `XDG_CONFIG_HOME` — 320 pass, 0 fail.


Latand:
Enforcement is deliberately out of this PR: the rule here is scaffold prose, so an agent with a terminal can still run a pattern kill. The blocking hook on the Viewer spawn path, and the process-group sandbox behind it, are filed as #1794.

Issue #1792, opened 2026-09-19
fix(delivery): a message to an idle conversation stays queued for ever because its authoritative turn still reads busy after the turn settled


Latand:
Filed by the orchestrator seat, 2026-09-19, from an operator report.

## What happened

The operator opened the conversation of an orchestrator seat that had just been rotated out and wrote it a message from the phone. The message sat as **queued** the whole time and was never sent. The operator gave up and asked the successor seat instead.

## Evidence (prod `e7d5c92e`, about 25 minutes after the conversation's last turn ended)

The conversation is the rotated-out seat `conversation_ff3b5715…`; its last turn ended 02:46:20Z and nothing ran in it afterwards.

- `conversation_deliverability`: `hostStatus: idle`, `deliverable: true`, `condition: deliverable`, `resumeRequired: false`.
- `agent_activity`: `turnState: idle`, host alive (structured), `reason: host_alive_transcript_silent`.
- `/api/files` entry for the same conversation: `activity: idle`, `activityReason: turn_evidence_settled`, `lastTurn.endedAt` set — and in the same object `authoritativeTurn: { state: "busy", source: "assistant", terminalAt: null }`.

So three surfaces say the turn is settled and the host can take a message, while the authoritative turn the composer reads still says busy with no terminal time. A send into a busy turn is held for the turn to end; this turn already ended, so the hold never releases.

For contrast, the live successor seat's entry carries `authoritativeTurn.source: "lifecycle"`. The rotated-out one fell back to `source: "assistant"`, which suggests the lifecycle evidence for its final turn is missing or no longer consulted after the seat was revoked, and the assistant-record fallback cannot see a terminal marker.

## Wanted

- A conversation whose turn evidence is settled (`turn_evidence_settled`, `lastTurn.endedAt` present, host idle and deliverable) never reports a busy authoritative turn. One predicate decides it, and the composer, the card and the files projection read that one predicate.
- A message queued behind a turn that has in fact ended is delivered at the next reconcile without the operator doing anything.
- A queued bubble that has waited longer than a bound says why it is waiting, in words, on the bubble.
- Regression test runs the production projection over a transcript whose last record is an assistant message from a settled turn with no lifecycle terminal event, and asserts the turn reads settled and a queued send drains.

## Notes

- Whether rotation itself plays a part (authority revoked while the host stays alive) is for the diagnosis to establish; the same final-record shape may affect any conversation.
[...]

Issue #1244, opened 2026-08-28
fix(board): keep unfinished work expanded — collapse on completion, not on idle age


Latand:
## Operator requirement

Unfinished work must stay on the board. Collapsing must key off **whether the work is finished**, not off how long a transcript has been quiet.

Once work *is* finished (merged / closed / abandoned), the operator's stated allowance is either of:

- hide the whole thing, or
- keep exactly one representative card — the agent that managed the work, if such a card exists.

Both are acceptable. What is not acceptable is the current outcome: work the operator still owes attention to disappears because nothing wrote to its transcript recently.

## Current mechanism

`shouldCollapseWorker` in `src/components/scheme/workerCollapse.ts` is completion-keyed for exactly one class and age-keyed for everything else.

- Reviewers collapse *on verdict* — correct, and the comment says so explicitly (`Reviewers collapse exactly on verdict, never on the idle window`).
- An implementer is a candidate only once its flow is closed, then still gated on the idle window.
- Every remaining worker falls through to the terminal line:

```
return context.nowMs - file.mtime * 1000 >= context.idleMs;
```

`DEFAULT_WORKER_COLLAPSE_IDLE_MS` is 15 minutes; the per-project board idle window defaults to two hours.

`isCollapseExempt` protects live work, pinned cards, and anything carrying a human-authored message — so the exemption vocabulary already exists. It has no notion of *unfinished*.

## Why the age fallback is the wrong proxy

Quiet and finished are different states, and the failure mode is one-directional: it hides work that is still owed.

A lane parked at a provider limit, a stage in `needs_decision`, a builder waiting on a relay, a blocked pipeline, or an agent that simply stopped — all stop writing to their transcript. Each ages past the window and folds into the idle stack while the work is still unfinished. Meanwhile a genuinely finished lane whose transcript happened to get a late write stays expanded.

The operator observed this directly: work set up over the preceding two days was no longer visible on the board, and it was not possible to tell from the board whether it had completed.

## Required behavior

1. A card stays expanded while its work is unfinished, regardless of transcript age.
2. Collapse is driven by a terminal outcome. Reuse the terminal vocabulary the codebase already carries (pipeline `state`/`stateDetail`, flow closure, reviewer verdict, PR merged/closed) rather than introducing a parallel notion.
3. When work is finished, the project setting chooses between: collapse everything, or keep one representative card for the managing agent when one exists.
[...]

PR #1803, opened 2026-09-19
Claude turns settle on the evidence a session transcript carries (#1792)

Latand:
A finished Claude conversation reported a busy authoritative turn for ever. In
the same `/api/files` object, the activity axis said `jsonl_turn_completed` with
`lastTurn.endedAt` set and the host said idle and deliverable, while
`authoritativeTurn` read `{state: busy, source: assistant, terminalAt: null}`. A
message written to such a conversation is held for the turn to end, and that
turn had already ended.

## The predicate

The authoritative branch of `turnStateFromRecords` closed a Claude turn only on
a top-level `result` record. The CLI writes that record into `--print` stream
output and never into a session transcript, so the branch had no terminal
evidence at all and every settled Claude turn projected busy.

A turn now closes on the assistant record's own stop reason, `end_turn` or
`stop_sequence`, read through one definition both projections share. Only a
record the provider authored counts: the CLI stamps a stop reason of its own on
an API-error record it may retry inside the same turn, and on a `<synthetic>`
record that never reached the provider at all (the queued-prompt no-op and the
replayed recovery prose of #516). Both keep their busy reading, and every #516
release still lands on its own boundary.

Replayed over the 128 KiB tail of 1276 local Claude transcripts, the two axes
agreed on 145 before and on 1264 after; 1200 busy authoritative readings become
1190 terminal ones. All 12 remaining disagreements are the two guarded shapes —
11 non-terminal API errors and one `<synthetic>` record — and they are a
different, pre-existing defect on the reverse branch, filed as #1811.

### A stop reason ends a message, not necessarily the turn

A stop reason is a statement about the MESSAGE, and the two come apart. When the
provider asks for several tools at once the CLI writes one record per content
block — same `message.id`, rising `apiBlockIndex` — and every one of those
records carries the whole message's `end_turn` stop reason while its own tool
call is still pending. A scan of the local corpus finds 183 such records across
29 of the 1276 transcripts, each answered by its tool result within seconds;
none is the last record of its file, but the window widens with tool duration.
While such a record is the tail, reading its stop reason would answer terminal
with tool calls outstanding and hand a queued message to an agent in the middle
of a turn.

So the `tool_use` block a record carries outranks its stop reason, in the same
one definition both projections read — which is what keeps the two axes
agreeing on it.

## The regression tests

`src/lib/scanner/settledTurnProjection.test.ts` runs the production scan —
`listFiles()`, the walk `/api/files` serves — over transcripts written the way
the CLI writes one, and reads the projected turn out of the entry the projection
produces:

- a prompt, a tool call, its result, an assistant record that ends the turn,
  then the harness bookkeeping, and no `result` record anywhere — terminal, and
  the card's own `turnLeftOpen` agrees. Red at the merge base with exactly the
  reported `{busy, assistant, null}`;
- the same conversation one record earlier, with an outstanding tool call —
  busy;
- one block of a split parallel-tool message, `end_turn` with a `tool_use` block
  still in its content — busy on both axes, `activityReason` included. Red
  before the guard above with `{terminal, lifecycle}`;
- the board fold, over the same two scanned entries: the settled one folds and
  the busy one does not. Red at the merge base on the fold.

## Which surfaces read this predicate

The card (through `turnLeftOpen`) and the files projection read
`FileEntry.authoritativeTurn`, and the registry's `conversation.turn` is derived
from the same projection. The composer does **not**: its turn axis is the host
ledger's `activeTurnRef`, which the structured delivery controller reads and the
send queue parks on. So this brings the transcript-derived axis back into
agreement with the host one rather than collapsing them into a single reader.

## What this changes downstream

Claude conversations now reach `terminal` the way Codex ones already do, so
several settled-turn behaviours that were silently unreachable for Claude become
reachable: a settled host may be retired by the idle sweep, a reseat may proceed
on transcript evidence rather than only on the live host's own lifecycle, and
boot stops re-hosting a conversation on the strength of a turn claim nothing had
confirmed. Work that is actually owed still wins everywhere it did before — the
startup adoption filter answers a pending delivery or a pending operation before
it consults the turn at all.

One of those newly reachable paths has a pre-existing hole, filed as #1810 and
not changed here: `completeProviderTurnObservation` consults its
transcript-terminal branch after the structured-host release check without being
fenced by it, so a reseat can proceed in the window between a host taking a turn
and the CLI journaling that turn's prompt record. The same hole already existed
for Codex; this makes it reachable for Claude.

**Board presentation.** `conversationSettled` folds a card on
`authoritativeTurn.state === "terminal"`, and that branch never fired for
Claude, so this does change what the board shows. The branch is genuinely
reached: a scanned session transcript carries `proc: null` — the scanner only
ever marks a transcript `running`, and `done` belongs to a claude-tasks output —
so neither the `done` short-circuit above the turn branch nor the `running`
collapse exemption holds these cards. Every quiet Claude card not held by
`isCollapseExempt` (live activity, running proc, pending question, rate limit,
held delivery, pin) now folds where it used to stay expanded. The fourth
regression case asserts exactly that off the production scan, rather than
leaving it as prose. That is the intended #1244 behaviour — completion-keyed
collapse, which #1244 made the only thing that folds a card — which this defect
had been suppressing for Claude. No collapse rule is touched here; the evidence
it keys off is simply correct now.

**Restart cost.** A terminal authoritative Claude turn is the one shape
`primeTranscriptTurnEvidence` deliberately refuses to carry across a restart,
because a persisted snapshot cannot say whether the release hinged on
`recoveryReleased`, which the migration coordinator reads to re-impose busy while
a live host still owns the transcript. Settled Claude turns are now that shape,
so each costs one bounded, identity-keyed tail read on a restart that hydrates
from a snapshot — where before it hydrated as busy for free and wrongly. The
fence is correctness for account migration and is left alone; the cost is filed
as #1808, and `scanCache.real.test.ts` pins the counts so any change to the
fence is visible.

## Why issue 1792 stays open

#1792 reports two things, and this delivers one of them. The turn-projection
half is fixed and asserted here. The delivery half — a message written to an
idle conversation being sent rather than left queued — is not, so **#1792 stays
open** and this pull request only references it.

The diagnosis found no journaled send for the reported conversation after its
last delivered seat tick, so no server-side turn-boundary hold ever existed for
that message: the delivery queue parks on the live host's own `activeTurnRef`,
not on this projection, and making the composer read this predicate instead
would be the wrong fix. The queued bubble the operator watched was the
composer's own local row for a send whose admission never landed. That is a
different axis with a different defect behind it, filed as #1804, and #1792
should be closed on that one.

Refs #1792


Latand:
Follow-up for the delivery half this PR deliberately leaves out: #1804

Issue #699, opened 2026-07-26
fix(board): make mobile conversation targets deliberate and safe to tap


Latand:
Origin: #693 UX audit; extends #474.

Overview-board conversation rows are nested inside a project button and have a narrow gap that navigates to a different destination. A near miss is never a no-op.

Scope:
- remove nested interactive controls;
- give conversation targets an intentional mobile hit area and unambiguous destination;
- preserve project-level navigation separately.

Acceptance: taps around a row cannot open a different conversation or the parent project accidentally.

Issue #1796, opened 2026-09-19
limits: Fable's own weekly window is never shown — the Viewer reads only seven_day_opus and assumes Fable shares it


Latand:
Filed by the orchestrator seat, 2026-09-19, from an operator request: "I want to finally see the limits for Fable — at Claude they are separate."

## State on prod `e7d5c92e`

`GET /api/limits` answers for Claude with `session` and `weekly` only; `flagship` is `null`, so the row #1358 added never renders for the operator. (At the moment of reading the answer was served from cache, `provenance.claude.reason: oauth-rate-limited`.)

`src/lib/limits.ts` reads exactly one tier bucket from the OAuth usage response, `seven_day_opus` (`OAUTH_FLAGSHIP_BUCKET`), and `src/lib/agent/models.ts` assumes Fable and Opus share it (`CLAUDE_FLAGSHIP_FAMILIES`). The operator's account shows Fable metered on its own window at the provider, separate from Opus, so that assumption no longer holds: either the response carries a Fable bucket under a key the Viewer ignores, or the Opus key is absent for this plan and nothing is shown at all.

## Wanted

- Establish the real shape first: run the production fetch against a signed-in account and record the KEY NAMES of the usage response (names only; never the token, never values tied to an account identity) in the PR body.
- Every tier window the provider reports (`seven_day_<tier>` that is a model tier, so Fable, Opus, Sonnet as present) is parsed and shown as its own labelled line beside 5h and Week, on the accounts screen, in the header limits surface and wherever the weekly line shows today. Non-tier keys (`seven_day_oauth_apps`, `seven_day_overage_included`) stay excluded.
- Spawn gating follows the same data: a Fable spawn is gated by Fable's window, an Opus spawn by Opus's, and a model with no bucket of its own by the general week only. This also settles #1431 (the flagship window gating Sonnet and Haiku spawns).
- The limits history keeps recording; a tier that was never reported simply has no line.
- Regression test feeds the production parser a usage response carrying a Fable bucket, an Opus bucket and the non-tier keys, and asserts the lines and the per-model gate.

PR #1831, opened 2026-09-19
Overview: the project kanban board, over every project, filtered to live work (#1820)

### LINKED ISSUE #1820: Overview shows the existing kanban board across all projects, filtered to cards with working workers
Filed by the orchestrator seat, 2026-09-19. REVISED the same day: the seat's first version of this issue invented a new column-per-project page, and the operator rejected it.

## The requirement, in the operator's words

First: "Overview should show a kanban overview for all projects, but only the status of where agents are working right now, without the other tasks."
Correction: "THE KANBAN AS IT IS NOW! JUST WITH A FILTER OVER ALL PROJECTS AND WORKING WORKERS."

## What that means

The Overview page (`src/components/OverviewBoard.tsx`, behind the rail's Overview row) shows the SAME kanban board a project shows today (`src/components/kanban/KanbanBoard.tsx`, mounted from `src/components/ProjectDashboard.tsx` around line 2327): the same status columns, the same task cards with their pipeline rows and conversations, the same card interactions. Two things differ, and only these:

1. **Scope: all projects.** The board is fed the tasks, pipelines and conversations of every project instead of one.
2. **Filter: working workers.** Only cards that have a worker working right now are shown (a running pipeline stage or a conversation with a live turn); a card waiting on the operator counts as live work too. Everything else (idle Inbox, Blocked and Done cards, finished pipelines) is filtered out, the way the board's own search filter hides non-matching cards today.

Because cards from different projects now share columns, each card carries a small label with its project's display name, and clicking that label opens the project's own board.

## Not wanted

No new page design, no new card design, no column-per-project layout, no second board implementation, no copy of the kanban model. If the board cannot take more than one project today, the change is to let the existing board and model accept a cross-project input and a predicate, not to fork them.

## Constraints

- Whatever needs a single project to make sense is hidden or disabled on the Overview rather than faked: the orchestrator seat panel, "+ Task", "+ Agent", drafts, drag between columns if the status write needs the project (it may simply use the card's own project).
- "Working" reuses the predicate the board already uses for its "N working" counters; no second definition. Known trap: until #1803 and #1815 are live a finished Claude conversation can read as working; take a running stage attempt or a live turn from the same evidence the board's counters use.
- Data comes from the projection `Viewer.tsx` already holds; no per-project fetch fan-out and no new polling loop. If tasks are fetched per project today, one request for all projects' tasks is acceptable; N requests are not.
- First-run behaviour of the Overview (no projects yet) stays.
- Rendered evidence once, at the end: Overview with working cards from three projects, and the empty state; desktop and phone.

Latand:
Closes #1820.

The Overview drew a grid of project summary cards. It now draws the **same kanban board a project draws** — `KanbanBoard` over `buildSchemeLayout` → `buildTaskBands` → `buildKanbanModel` — with exactly two differences and no second implementation: its scope is every project, and only the cards a worker is working on right now are shown, hidden the way the board's own search hides what does not match. Because projects share the columns, each card carries a small label with its project's display name, and the label opens that project's board.

## What was generalized, and what stayed project-only

The first step the issue asked for, written before any code (it is the body of the first commit as well).

| Input | Today, on a project's board | On the Overview |
|---|---|---|
| `project` | the one project key: the task projection's unlinked fence, the created-task filter, the seat read, the `ReaderMemory` key, the loose reader's scope, the focus index, the draft context, the task composer | stays the board's **identity** (the rail's own `__overview__` key), so reader memory, focus and the loose reader keep working unchanged. Only the projection's argument changes: `projectTaskWorkflows` uses `project` to fence its **unlinked** buckets and nothing else, so passing none *is* the cross-project answer — one call, not one per project |
| `groups` | `buildBranchGroups(files, project)`, narrowed by that project's board preferences | one grouping pass per shown project, concatenated. The preferences are per-project server state the Overview has none of, so it groups unfiltered |
| `manual`, `drafts`, `draftBands`, `favorites`, `isolatedManualPaths` | per-project preferences and in-flight creation | empty — the Overview creates nothing and pins nothing |
| `files` | already every scanned conversation | unchanged |
| `flows` | already global (`useEffectiveFlows`) | unchanged |
| `pipelines` | already global | unchanged |
| `surfacePipelines` | `pipelinesForProject` — the fence that lets only this project's pipelines grow memberless stage rows | new `pipelinesForProjects` over the union, so the fence admits exactly the projects on the board. One pass, not one per project |
| `tasks` / `allTasks` | `tasks.filter(project)` | every shown project's tasks, out of the **one** payload the Viewer already polls. No new request, no fan-out, no new loop |
| `layout` | the dashboard's prebuilt `pipelineLayout`, shared with the scheme leaf | omitted; the board builds the same layout from the cross-project inputs itself |
| `now`, `loaded`, `catalogFailures` | board clock and scan state | unchanged. The clock is literally the dashboard's: `BOARD_CLOCK_MS` moved beside `useNowSeconds`, which shares one timer per cadence |
| `selection` | the scheme's selection | empty |
| `seat`, `seatRefs`, `projectCwd` | the project's orchestrator seat | omitted. A seat belongs to one project, so the panel is absent rather than faked, and no seat is read |
| `onNewAgent`, `onAddAgent`, `onDraftClose`, `onDraftSpawned`, `onHandoff`, `onSpawnRetry` | creation bound to one project's checkout | omitted: «+ Agent» is absent, on the bar and on every card |
| «+ Task» | `KanbanTaskComposer` writes into one project | absent; it would have to choose a project for the operator |
| `closedPaths`, `onCloseConversation`, `onRestoreConversation` | per-project board preferences | omitted |
| `onOpenConversations` | the project's conversation list | the Overview's own global search |
| a status move | already writes the task's own `expectedProject` | unchanged — which is what makes moving a card between columns correct across projects |

Nothing in that list needed a rewrite to accept a cross-project input, so there was no `needs_decision` to report.

## Working reuses the counters' own evidence

`cardHasLiveWork` is `card.working` (the number the column's «N working» shows — a member conversation whose row state is working or held), `card.needsYou` (the number its «N need you» shows), and a stage chip in flight. No second definition: a Claude conversation whose turn closed is not working to the counter and is not working here (#1803, which is in this branch's base).

## What the model gained

Two things, both inert on a project's board: every card now carries its own `project` (the label reads it, and a status write already used the task's own), and `cardFilter` narrows `shown`/`unlinkedShown` exactly where `query` does — every count is still taken over the whole inventory, as it is under a search. The Overview's columns therefore read «3 of 41», and the page's subtitle says «Only what is working right now» so that reads as the truth rather than a bug.

## The grid is gone

It was a second, smaller board beside the real one, and the rail already lists what it listed. Its structural lesson from #699 — nothing that navigates nested inside anything else that navigates, the card itself never a target — moves onto the new surface and is asserted there. `OverviewBoard.targets.dom.test.tsx` is replaced by `OverviewBoard.kanban.dom.test.tsx`, which carries that case plus this issue's.

## Verification

- `tsc --noEmit`: clean (logged to a file, exit code read).
- `bun run build`: clean, MCP bundle included.
- Tests by path: `OverviewBoard.kanban.dom`, `OverviewBoard.render`, `OverviewBoard.firstRun.dom`, `OverviewBoard.catalog.render`, `kanbanModel`, `KanbanBoard.dom`, `BoardSearchEntry.dom`, `ProjectRail.firstRun.dom`, `pipelineModel`, `useNowSeconds.dom`, `i18n` — all green.
- Privacy publication gate from the merge base: PASS.
- Rendered evidence: a case added to the kanban board's one driver (`kanbanBoard.browser.test.tsx`, `#1820` block) over the fixture the other cases use, at 1440×900 and 390×844, working cards and the first run. Measurements in `evidence/issue-1820/overview.json`; frames go to the uncommitted `.artifacts/issue-1820/`.

One note on a directory-wide run: `bun test src/components/kanban/` fails `KanbanAccountsRuntime.dom.test.tsx` in this checkout. It fails there with the branch's sources **and** with `main`'s sources pasted over them, and passes when the file is run on its own — a cross-file `mock.module` leak this checkout provokes, not a change from this branch. Every file above was run by path, which is how CI runs them.


Latand:
Merged `main` (through #1833) into the lane and pushed `d51a510a` as a fast-forward of the branch.

The only conflict was `src/components/kanban/kanbanBoard.browser.test.tsx`, where this lane's `#1820` describe block and main's `#1796 per-model limits` block were appended at the same place; both are kept whole, and the same is true of the earlier merge of #1824/#1819 that this branch carried.

The review's P2 is answered at this head: the quiet Overview draws `overview.noneWorking` with a body naming the Overview's own filter, `OverviewBoard.kanban.dom.test.tsx` pins that copy in both directions (the quiet state must not read as a search; a fruitless search must still read as one, on the Overview and on a project's board), and the rendered evidence covers the quiet state at both viewports. The browser driver re-ran at this head and reproduced `evidence/issue-1820/overview.json` byte for byte, with no failures.

The review's non-blocking P3 is filed as #1838 with the measured numbers; nothing about it is changed here.

Gates at this head: `bunx tsc --noEmit --incremental false` clean, `bun run build` clean, the touched tests green by path (`OverviewBoard.kanban.dom.test.tsx`, `OverviewBoard.render/catalog/firstRun`, `BoardSearchEntry.dom.test.tsx`, `ProjectRail.firstRun.dom.test.tsx`, `kanbanModel.test.ts`, the i18n parity tests), and the privacy gate passes from the merge base.

Issue #1853, opened 2026-09-19
Kanban K6a browser case fails on the conversation picker head since the #1837 fixture change


Latand:
The gated board driver's K6a case (\`LLV_KANBAN_BROWSER_TEST=1 bun test src/components/kanban/kanbanBoard.browser.test.tsx -t \"#1695 K6a: account chips\"\`) now fails on two checks at current main:

\`\`\`
conversation picker light: head Account · Verify · Claude
conversation picker dark: head Account · Verify · Claude
\`\`\`

The check at \`kanbanBoard.browser.test.tsx\` (\`measure.head !== \"Account · Verifier · Claude\"\`) still expects the old conversation name, while the evidence fixture's task titles were rewritten in #1837, and the running stage conversation's picker head now reads \`Verify\`. Every other K6a check passes. CI does not run this driver, so nothing went red.

Fix: decide which name the picker head should carry for that fixture conversation, then align either the fixture or the expectation, and regenerate \`evidence/issue-1695/k6.json\`.

Found while running the driver for #1846.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #1705, opened 2026-09-15
A pending account switch made through reconfigure can't be cancelled: rollback is undone, superseding cancels held messages


Latand:
## What happens

A conversation's account switch requested through `reconfigure` can't be cancelled without side effects:

- **Rollback is undone.** Suppose a switch to another account is waiting for the running turn (`waiting-turn`) and is rolled back through `POST /api/conversations/:id/migration {action: "rollback", expectedRevision}`, or the MCP `conversation_migration` rollback. The structured delivery queue then retries the reconfigure that owns the switch. That retry requests the reseat again, and when the turn ends the conversation moves to the account the rollback was meant to cancel.
- **Superseding it cancels held messages.** A newer `reconfigure` to a third account, or back to the account the conversation is on, does replace the pending switch. It also fails every message held for the switch, with "delivery cancelled because its owning account migration was stopped; send again to authorize a fresh delivery". Switching back also releases and restarts the host at the turn boundary.

No path today cancels a pending switch while keeping the messages held for it.

## Evidence (isolated test, main 098932f8)

Each run starts from the `switchRequestedMidTurn()` fixture in `src/lib/runtime/structuredAccountSwitch.test.ts`:
- the conversation is on `account-a`;
- a `reconfigure` to `account-b` has been claimed;
- the migration is in `waiting-turn`.

A message was then held with `registry.holdDelivery(...)`.

| Step | Migration | Held message |
| --- | --- | --- |
| Rollback | `rolled-back`; intent still `draining`; reconfigure still `applying` | `assigned` (re-armed) |
| Queue retries the same reconfigure | `waiting-turn`, target `account-b` again | `assigned` |
| Turn ends, reconfigure runs again | `committed`; conversation on `account-b` | — |
| Instead: newer reconfigure to `account-c` | `waiting-turn`, target `account-c` | `failed` (cancelled, stopped migration) |
| Instead: newer reconfigure back to `account-a` | `rolled-back`, host released once | `failed` (cancelled, stopped migration) |

Repro, appended to `structuredAccountSwitch.test.ts` (uses its helpers; run by path with an isolated `HOME`, `XDG_CONFIG_HOME` and `TMPDIR`):

```ts
test("a rolled-back switch owned by a reconfigure stays rolled back when the queue retries it", async () => {
  const c = await switchRequestedMidTurn();
  const id = c.conversation.id;
  const held = c.registry.holdDelivery(id, "sent while pending", "held-rollback");
[...]

Issue #86, opened 2026-07-10
Account migration floods the board: every drained successor fork appears as a fresh manual card


Latand:
## Observed (2026-07-10 ~15:25, prod)

When migration intent efd36421 drained (46 conversations retried after #82), the Codex successor provider forked every thread via app-server. Each fork wrote a new rollout under ~/.codex/sessions, the scanner picked all of them up as brand-new active roots, and the board's manual list jumped from ~2 to 53 cards within a minute — the operator saw "миллион окошек" appear out of nowhere and had to bulk-close 51.

## Expected

A committed successor REPLACES its predecessor on the board:
- inherit the predecessor's placement (manual/auto/expanded slot) via the existing `remap-paths` board mutation — the registry knows the source path → fork path pairing at commit time;
- a successor whose predecessor was hidden/quiet stays hidden/quiet;
- mass-drain must never enqueue N new manual roots.

## Where to look

`src/lib/accounts/migration/coordinator.ts` (`commitSuccessor`) — nothing tells the board about the path change; board membership then treats forks as new conversations. The board already supports `remap-paths` mutations (src/lib/board/mutations.ts) built exactly for path continuity.

Refs #40, follows #82.

Latand:
Implementation is published as draft PR #101: https://github.com/Latand/live-log-viewer-next/pull/101\n\nValidation completed: 1166 tests passed and TypeScript completed without diagnostics. The PR body records the remaining audit caveats after the self-spawned review flow was stopped.

Issue #40, opened 2026-07-09
In-UI multi-account switching per engine: store, toggle & quick-login Claude / Codex accounts (fast swap, no re-entering passwords)


Latand:
## Goal

Manage **multiple accounts per engine** (multiple Claude subscriptions, multiple Codex/ChatGPT subscriptions) **directly in the viewer UI**, and switch between them with a single flick — a toggle/dropdown, no re-entering passwords, no opening extra apps.

Concretely, I want to:
- Store several accounts per engine and label them.
- Switch the active account with a toggle/quick-pick right in the UI.
- Add an account fast via the engine's own **code/device login** ("with digits") — `chik` add, `chik` switch, `chik` sign out — without leaving the viewer.
- Quick sign-out / sign-in without fiddling with password managers.

## Where creds live today (grounding)

- **Codex:** `~/.codex/auth.json` — `auth_mode: "chatgpt"`, `tokens: { id_token, access_token, refresh_token }` (or `OPENAI_API_KEY`).
- **Claude Code:** `~/.claude/.credentials.json` — OAuth tokens (0600).
- The viewer **doesn't touch auth at all today.** Spawns just call `resolveBinary("claude"|"codex")` — `src/lib/agent/cli.ts` (`freshSpecFor` :75, `resumeSpecFor` :112). So account = whatever global file exists on disk.

### Approach A — per-spawn env-scoped config dir (preferred, no global mutation)
Both engines support redirecting their config/creds dir via env:
- Codex: **`CODEX_HOME`**
- Claude Code: **`CLAUDE_CONFIG_DIR`**

Keep one dir per account, e.g. `~/.config/agent-log-viewer/accounts/{codex,claude}/<label>/`, each holding that account's `auth.json` / `.credentials.json`. At spawn time inject the chosen account's env var into the command spec (`freshSpecFor`/`resumeSpecFor` in `cli.ts`, validated in `src/app/api/spawn/route.ts`).

**Why this is the right shape:**
- No global file swapping → different running agents can use **different accounts simultaneously**.
- No race where switching accounts yanks the token out from under an already-running agent.
- Sidesteps the revocation worry below (each account keeps its own refresh token in its own dir; nothing overwrites another).

*Needs verifying these env vars are honored by the installed versions — see AGENTS.md note "read the guide before writing code".*

### Approach B — global auth-file swap toggle (simpler, but blunt)
[...]

PR #1854, opened 2026-09-19
Account switches are instant and optimistic: the pick shows at once, the conversation moves with its next message

### LINKED ISSUE #1846: Account switches are instant and optimistic: record the intended account at once, move the conversation at its next engagement
## Operator request, 2026-09-19 (paraphrased in English)

> About the model sheet and switching between accounts in general: a switch between accounts must be instant and optimistic. The conversation is actually carried over to the other account only later, at the moment it is engaged again: when I send a new message, or the orchestrator acts on it, or anything else uses that conversation. We must not wait for a chain of confirmations, or for agents to finish writing their answer. The choice changes at once.

## Today

Picking another account for a conversation starts the migration machinery immediately (#40, #86: ownership is rebound, never copied) and the UI waits on it. While the agent is mid-turn the switch waits or is refused, and the sheet shows nothing until it settles. The UI critique of PR #1822 measured exactly that on the phone sheet: a tap on a ready account sends one select and then nothing on screen changes. PR #1822 adds a "next message" mark for the picked account; this issue is the behaviour behind that mark, for every surface that switches accounts.

## What to build

1. **The choice is recorded at once as the conversation's intended account** — durable, per conversation, separate from the account it is running on now. The sheet, the desktop popover, the header and the accounts panel show it immediately: "runs on A · next on B". No spinner, no wait, and it works while the agent is mid-turn.
2. **The move happens lazily, at the next engagement**: an operator message, a seat `send_message`, a stage retry, a resume, a wake. Before that turn starts, the Viewer performs the existing ownership rebind to the intended account and then delivers. A running turn is never interrupted for a switch; per the limit semantics a running turn continues and only new turns need the account.
3. **Changing one's mind is free**: picking A again before any engagement clears the intent and nothing ever moved.
4. **If the move fails at engagement** (account signed out, limit reached, migration refused), the message is not lost: it stays held with a plain reason and the two obvious actions (send on the current account, or pick another).
5. One mechanism for the operator's pick, the seat's tools (`conversation_migration`, pipeline `override-stage` account for the next attempt) and the automatic switch under rate-limit pressure: all of them set the intent; only engagement performs the move.

## Out of scope

Copying conversations between accounts (#86 forbids it). Changing how limits are read.

## Acceptance

- Picking an account while the agent is mid-turn updates every surface within one frame, sends no migration, and the running turn finishes on the old account.
- The next message runs on the picked account, and the transcript keeps one continuous conversation.
- Pick B then A before engaging: no migration is ever attempted.
- A failed move leaves the message held with its reason; nothing is sent twice.
- DOM and integration tests go red without the change.

Touches `RuntimePill.tsx` and the accounts migration code: start after PR #1822 merges.

Latand:
An account pick for a structured conversation now takes effect in the UI straight away. The conversation itself moves to the new account the next time something engages it. No surface waits for a migration, and a running turn is never interrupted.

Closes #1846

## What happened before this change (read from the code, then pinned by a test before any change)

- **Phone sheet (`RuntimePill` Account group):** a tap on another account called `useEngineAccounts.select`, i.e. `POST /api/accounts/<engine>/active`. That moved the **engine-wide** launch account, not this conversation's. This conversation (and every other unpinned one on that engine) was then moved lazily on its next send by `requestConversationMigrationToActiveAccount`.
- **Board picker and header chip (`accountChoice` / `AccountPicker`):** these sent a per-conversation `reconfigure {accountId}` into the runtime journal. The structured delivery queue applied it **eagerly**: while the agent was mid-turn the queue blocked, and the moment the turn ended it ran the ownership rebind (`requestConversationReseat` → `advanceConversationMigration`) even if nobody ever sent another message. With the agent idle it migrated at once. A pick that waited more than two minutes behind a long turn was also failed by the control settlement deadline.
- **A move that failed** (signed-out target, refused migration) was marked failed, and the message queued behind it was **delivered on the old account** in the same pass.
- **`conversation_migration reseat`** created the migration record and advanced it right away.
- **Automatic switch under rate-limit pressure:** `chooseAutoBalance` has no production caller. The only automatic move is the engine active-account migration on send, which was already lazy at engagement.

`src/lib/runtime/structuredAccountIntent.test.ts` recorded that before the fix: on unchanged code 7 of its 8 cases failed. The pick migrated when the turn ended with no engagement, an idle pick migrated immediately, and a failed move let the held message go out on the old account.

## What changes

One mechanism, the existing one: the per-conversation `runtime.reconfigure` with an `accountId`, which the runtime session already projects as `pendingReconfigure`. It moves the conversation through the **existing** rebind (ownership rebind, never a copy: #40, #86). No second migration path.

1. **The intent is recorded and shown at once.** The phone sheet and a new **Account** panel in the desktop popover send this conversation's reconfigure and show "runs on A · next on B" in the same frame, while the agent is mid-turn and before the server answers. A waiting pick shows no spinner. The board picker says "Account A → Account G with the next message" until a message engages the switch. After that, the migration record's "after this turn" wording applies.
2. **The move is lazy.** `StructuredDeliveryQueue` keeps an account-switching reconfigure parked until a message (`send`, `steer`, `native-queue`) for that conversation is open, then applies it before that message. A mid-turn engagement waits for the turn to end and never interrupts it. A parked pick is exempt from the two-minute control deadline, and on its own it no longer makes startup re-host the conversation.
3. **Changing your mind is free.** Picking the account the conversation already runs on withdraws a waiting pick (the #1705 withdrawal). If nothing else changed, no command reaches the host, and at the next engagement the queue settles the pick as `cancelled` without moving anything. A settings change made while a pick waits carries the picked account forward, so the newer reconfigure does not drop it.
4. **A failed move holds the message.** When a move fails at engagement, the registry keeps a durable per-conversation `switchHold` with the reason. The message stays `queued` where it is, so nothing reaches any engine and nothing can be sent twice. The composer shows "Not sent: moving to B failed — <reason>" with **Send on A** (`conversation_migration keep-current`, or picking A) and **Pick another account**. A new pick's claim ends the hold.
5. **Every caller uses the same path.** The operator's pick (sheet, popover, board picker), the seat's `conversation_migration reseat` (on a structured conversation it now answers `reseat: intended`) and the rate-limit reseat button all record the intent. Only an engagement moves the conversation. A pipeline stage's `override-stage` account was already an intent for the next attempt. `bindings.ts` is untouched: the new `keep-current` action is added to the tool's schema in `server.ts`.

Not covered: a pick on a conversation without a structured host (the legacy tmux path) still switches as it did before. A kill ends a parked pick, as it ends any queued reconfigure.

## Verification

All of it ran on the Viewer's code paths with a private state directory. The runtime host process was not started.

- `bunx tsc --noEmit --incremental false`: clean.
- `bun run build`: passes.
- Privacy gate `--base <merge-base> --check-commits` with the fingerprint catalog: PASS.
- Touched tests, run by path with `LLV_STATE_DIR` pointing at a scratch dir, all green:
  - queue: `structuredAccountIntent` 8, `structuredDeliveryQueue` 68, `structuredSwitchCancelQueue` 2, `structuredDeliveryQueue.inject` 13, `.recoveryContention` 9
  - dispatch and command: `structuredControls` 41, `structuredSwitchCancel` 23, `structuredReconfigure` 20, `structuredDelivery.integration` 38, `structuredDeliveryRebind` 13
  - message delivery: `structuredMessageDelivery` 62, `.accountReseat` 25
  - routes and wiring: `startup` 97, `files/route` 96, `tmux/route` 30, migration route 7, `reseatCommandBinding` 2, `conversation/actions` 8, `mcp/bindings` 69, `mcp/presentation` 3, `i18n` 20
  - UI: `RuntimePill.accountIntent` 6, `RuntimePill` 18, `RuntimePill.limitAccounts` 9, `KanbanAccounts` 17, `KanbanAccountsRuntime` 5, `RateLimitBadge` 4
- The new DOM file (`RuntimePill.accountIntent.dom.test.tsx`) fails 6 of 6 against `RuntimePill.tsx` at HEAD. The queue file fails 7 of 8 on the unchanged queue.
- **Rendered, in Chromium** (the existing K6a case in `kanbanBoard.browser.test.tsx`): after a pick, the board picker reads `Pending: Account G · moves with the next message` and the chip reads `with the next message`. Exactly one `reconfigure {accountId: account-g}` left the page, with no account select. One K6a check fails, and it is unrelated to this change: the picker head reads "Verify" where the driver expects "Verifier", because of the fixture rewrite in #1837. It is filed as #1853, and the committed `k6.json` is left as it was. The phone driver's fixture has no runtime plane, so its pill is not on the structured surface; the DOM tests cover the sheet and the popover.

## Left unchanged on purpose

- **The engine accounts panel.** It picks the account used for new launches of an engine. Its model holds no conversation, so it has no conversation whose next message it could move. A per-conversation pick is made in the runtime pill, the card's account chip or the board picker, and all of them share one intended-account store.
- **The phone composer chip.** The phone composer's model chip does not show "→ next" while a pick waits. On the phone, the conversation's title line names the next account, and the sheet the chip opens names both accounts.
- **The desktop conversation-pane badge ("@ A → B").** No desktop surface mounts it today. A kanban reader replaces the pane's header with its own account chip, the scheme panes' board is mounted nowhere, and the phone has no pane header. The rendered readings cover the chip the reader does draw: `evidence/issue-1846/desktop-deck.json` (1280×900, the deck surface, en and uk, short and long ids, in the reader and in the same reader held to 426 px) and `evidence/issue-1846/shared-pick.json` (the reader's chip in its board column at 1440 and 1280, en and uk).
- **Phone title vs. a long next id.** The phone title now keeps at least 6rem beside a waiting pick. At 390 px a 20-character next id therefore shows its head (95 of 116 px) instead of pushing the title down to seven letters. Short ids still show both accounts whole (`evidence/issue-1846/phone-header.json`).


Latand:
Fix round for the render critique (head 8da0e344):

- **One pick, every surface, same frame.** A page-wide store (`src/lib/accounts/intendedAccount.ts`) holds the pick. The runtime pill and sheet, the phone title line (`@ A → B`) and the card account badge all read it and fall back to the session's projected intent.
- **The legacy badge is optimistic.** It shows no spinner, has no 60 s "no confirmation" error and keeps no disabled menu. The running account stays pickable, marked "cancel switch".
- **Taking a pick back is quiet.** The pill treats a `cancelled` or `superseded` reconfigure receipt as settled: no danger face, no toast. It also no longer adopts a projected account pick as an apply in flight, so no stale `pending` phase reaches localStorage.
- **The failed-move notice has its own line.** It is now `RuntimeSwitchHold`, a full-width block in the composer's receipts region with 44 px actions on the phone. "Pick another account" opens the pill's account choice.
- **The running account reads as the way back.** Its sheet and popover rows say "cancel switch" while a pick waits.
- **The board chip is optimistic from the first frame.** It says "with the next message" where it used to say "asking…".

Tests: `RuntimePill.accountIntent` (quiet take-back, no stale pending phase, notice outside the pill row), `AccountBadge.dom` (5 of its cases fail against the previous head), and `MobileFocusView.conversation` (the header names the next account). tsc, build and the privacy gate pass. `TmuxComposer.liveRefreshFocus`, `MobileFocusView.keyboardInset` and `EffortPills.slot` fail identically at the previous head. No new rendering was done this round.

Latand:
Fix round for the render critique (55195761):

- **P1:** the board's account chip and picker now read and write the shared picked-account store that the runtime pill, card badge and phone title already use. A pick or take-back made on either surface shows on the other in the frame of the click. The rendered case in `kanbanBoard.browser.test.tsx` (`#1846`, structured runtime session in the fixture) measured 13–29 ms, with no server projection involved (`evidence/issue-1846/shared-pick.json`). While a pick waits for the next message, the board picker replaces or takes it back with one reconfigure. A switch known only to this page, with no runtime plane, keeps the old withdraw-by-operation path.
- **P2:** on the phone title line, the running account keeps up to 60% of the tag and the next account truncates first. The tag is capped at 55% while a pick waits (45% otherwise).
- **P3:** the popover's way-back row reads "cancel switch" as an accent action. A held message translates the known failure causes (signed out, usage limit, refused); any other reason is shown as the server sent it.

Gates: tsc clean, `bun run build` ok, the touched test files pass run by path, privacy gate PASS. The phone header's long-id layout is covered by class assertions in a DOM test, with no browser render this round.

Latand:
Fix round for the render critique (315a44b5):

- **One name per account.** The "runs on · next on" line in the sheet and the popover, the held message, the card chip and the phone header now use the account's label, the same one the rows and the board show. The id appears only for an account the list does not enumerate, such as the legacy home.
- **Phone title line.** While a pick waits, the next account comes first and shows whole. The running account shows before it only when both fit; otherwise it wraps onto a second line that the one-line tag clips, so no sliver of it shows. At 390 px with `review-relief-2` → `production-backup-7`, the next account draws at 116.1 of the 116.1 px it needs, in en and uk. The model and tier stay whole.
- **Desktop pill.** While a pick waits, the pill's face shows `→ <next account>`, because the popover closes when you pick.
- **Rendered evidence** goes into the existing drivers. The phone driver has a new `#1846` case: 390 px, en and uk, short and long ids (`evidence/issue-1846/phone-header.json`). The kanban `#1846` case now runs in en at 1440 and in uk at 1280, and checks the pill mark and the board chip's longer Ukrainian tail (`evidence/issue-1846/shared-pick.json`). The kanban fixture's reader has no account chip, so the reader-chip reading is recorded as absent.

Gates: tsc clean, touched DOM tests by path (80 pass), both browser cases green, `bun run build` ok, privacy gate (with fingerprints, `--check-commits`) PASS.

Issue #1881, opened 2026-09-19
A seat is not woken when agents it spawned outside a pipeline finish: their completion is 'unreadable' and the fallback wake is hourly


Latand:
Second report from the same new user (relayed by the operator, paraphrased in English).

## Symptom
The orchestrator spawned three reviewers as separate agents (spawn, outside any pipeline) and ended its turn to wait for the Viewer's wake. All three finished with an approve verdict; nothing woke the orchestrator until the user wrote to it, an hour later. The seat tick it did receive said its spawned children were 'not listed: the Viewer cannot resolve their transcript, so no seat can read them'. The user had never touched the tick settings. The orchestrator then worked around it: it set its own tick to two minutes and told every builder and reviewer to end by sending it a message, which does wake it at once.

## What the code says
- `DEFAULT_SEAT_TICK_POLICY.checkIntervalMs` is 5 minutes, and a settled child is a wake reason only when the child is harvestable (`childFactsSkipReason` in `src/lib/monitor/seatTick.ts`: transcript resolvable and a finite own instant). An unreadable child is dropped into the parenthetical count and wakes nobody.
- With no other reason, the only wake left is the interval: `SEAT_TICK_WAKE_INTERVAL_MS` = 60 minutes.
- This seat's own ticks on the production machine show the same parenthetical on every wake ('14 spawned child(ren) not listed: the Viewer cannot resolve their transcript'), so unreadable children are common, not a fresh-install curiosity.

## To find first
Why a just-finished child of the seat is unreadable. Candidates: the project identity split of #1874 (seat under `dir-…`, children under `repo-…`), the child's transcript living in the shared store while the resolver looks in the account home, or the registry entry losing its path once the host is reclaimed. Reproduce on a fresh state directory: seat spawns a child with `src` lineage, the child ends its turn.

## Wanted
- A child the seat spawned settling its turn is a wake reason by itself, read from the registry's turn state (no transcript needed to know that it ended), with the child's final message attached when the transcript is readable and a plain 'finished, transcript not readable: <why>' line when it is not. Never silence.
- Unreadable children are diagnosable: the wake names the reason per child once, instead of a count.
- The fallback interval while the seat has running or just-settled children is minutes; the hourly bound stays for a board with nothing moving.
- Test: fresh state dir, seat spawns a child, child settles: the next check raises a wake naming the child within one check interval.

Issue #1876, opened 2026-09-19
First-run onboarding: connect engines, pick model defaults, phone access, and a short tour of what the Viewer does


Latand:
Operator idea after watching a new user's first evening (paraphrased in English): the product needs an onboarding.

## Scope to design
1. **Engines and accounts**: detect Claude and Codex CLIs and their sign-in state, offer sign-in, and state plainly what works with one engine only.
2. **Agent mapping**: which model and effort build, review and critique by default on this install (the role presets), editable in one place, with cost hints. Ties into the defaults issue for installs without Codex.
3. **Phone access**: an optional step for reaching the Viewer from a phone (Tailscale or another private network), with the access token and the URL to open.
4. **A short tour**: the board, tasks, pipelines and stages, the orchestrator seat and how it is woken, review rounds, where to look when something waits for you.
5. **Health check** at the end: a throw-away one-stage pipeline on the cheapest model proves spawn, delivery, stage report and the seat wake on this machine, and points at the failing part when it does not.

## Notes
The wizard is reachable again later from settings. Text in English and Ukrainian. Design first (rendered at desktop and phone widths), then build.

Latand:
Scope additions by the operator's decision (paraphrased in English), moved here from #1875:
- The agent-mapping step decides, per install, which engine, model and reasoning effort each role runs on; the shipped defaults stay as they are (review on Codex) until the user chooses. The step shows cost hints, since a new user's first review ran on gpt-6-astra at xhigh.
- The same design decides what a launch does when a stage names an engine with no signed-in account (refuse at once with a plain message, offer to connect, or use the mapping), so that a stage never starts and dies.
Priority raised: next design lane after the state migration's first slice.

Issue #1874, opened 2026-09-19
A seat designated under a project's name never gets woken when its lanes are recorded under the repo key (fresh install)


Latand:
Report from an external user on a fresh npm install (relayed by the operator, paraphrased in English).

## Symptom
The user's orchestrator started a pipeline; the builder fixed all review findings and pushed (its PR head moved, builds green). For the next hour, then two, nothing happened: the orchestrator waited for the Viewer's wake and none came, so the user had to ping it by hand between stages. `request_attention` from the same orchestrator was refused as well. The orchestrator's own reading: its seat is bound to the project under its plain name (the directory name), while the pipeline and its agents are recorded under the canonical key `repo-<hash>`.

## Likely cause (to confirm)
Seats are keyed through `canonicalOrchestratorProject` → `canonicalProject` (`src/lib/projects/aliases.ts`), which maps a named project to its repository identity only when an alias record exists. On a fresh install the seat can be designated before that alias exists (or the alias is never written for a project first seen by name), so the seat lives under the name and everything the pipeline engine writes lives under `repo-<hash>`. The seat tick looks up lanes and events by the seat's project, finds none, and never wakes it; attention authority compares the same two keys and refuses.

## Wanted
- One identity: designating a seat, creating a pipeline, the seat tick, attention authority and the task board all resolve a project through the same function, and a seat stored under a name is re-keyed (or matched) as soon as the repository identity is known. An existing mis-keyed seat heals at boot.
- A test on a fresh state directory: create the orchestrator for a directory that is a git repository nobody has scanned yet, start a pipeline, settle a stage: the seat tick names that lane in a wake, and request_attention from the seat is accepted.
- The refusal text of request_attention names both keys when they differ, so the next report is diagnosable in one read.

Latand:
Cause confirmed from the code and the user's second report (paraphrased in English).

**Observed by the user:** the seat holds the key `dir-<hash>` (derived from the folder path); every pipeline created with `repoDir` pointing at the same folder holds `repo-<hash>`. The sidebar shows the project twice under the same name, one entry with the seat, the other with the lanes. `create_pipeline` has no project field, and the agent's attempt to designate itself in the repo-keyed project was refused as an operator-only action.

[...]

PR #1897, opened 2026-09-19
Setup guide, slice 2: the health check proves spawn, delivery, stage report and the seat wake (#1876)

Latand:
Slice 2 of the first-run setup guide (#1876, `docs/design/onboarding.md` §5): the **Check** step. It proves on this machine that an agent starts, a message reaches it, a stage reports back and a sleeping orchestrator is woken. When one of those fails, it names the part that broke.

## What it does

A run uses the cheapest connected model at its lowest effort (Claude Haiku low, else Codex Luna). Only one run exists at a time. It goes like this:

1. A test orchestrator starts in a scratch repository under the state folder (`state/onboarding/viewer-health-check`, no origin).
2. That orchestrator becomes the creator of a one-stage, read-only pipeline, and the stage agent starts.
3. The stage calls `stage_report`, and the pipeline completes.
4. The production seat tick check (`runSeatTickCheck`) runs over the scratch project. It must decide `wake` for "a lane you launched is completed", and the send must reach the idle test orchestrator.
5. Every real seat is checked, read-only, for the #1874 condition: `projectSuccessionFor` must find no key its folder has moved on from.

Each row is read from a fact the Viewer already records: the transcript on disk, the launch receipt's `prompt-delivered`, the settled attempt, and the check's own journal line plus the delivery outcome. The agent's own claim counts for nothing. A failed row turns red and expands in place with its code, one sentence on what happened, one on what to do, one action (Go to Engines, Open Accounts, Open the agent, or Copy details) and the redacted machine detail. The rows after it stay waiting. Row bounds are 60 s per launch, 30 s, 120 s and 60 s, and the whole run is capped at 5 minutes. Stop cancels the run.

Cleanup runs whatever the outcome: the pipeline is closed, both hosts are stopped, the worktree and its branch are removed, both conversations are archived off the scratch board, and the task cards are hidden. The result goes onto the guide's marker as `lastHealth`.

## One reduction from the design, as the design anticipated

A seat can only be ended by a rotation or a stillborn rollback, and both live under `src/lib/orchestrator/`, which is outside this slice's fence. A designated scratch seat would outlive the run and be woken by the real tick every day. So the test orchestrator reaches the check through the check's own sources, on a tick row the run owns. The gather, the project identity, the decision and the send are all the production ones. The one thing left unproven is the seat store's own lookup, and row 5 reads that store for every real seat. A throw-away seat that goes through the intent path needs a way to end a seat, which belongs in a separate lane.

Two more things differ from the design. There is no project archive, so the scratch project stays in the rail under the folder's name, with its conversations archived and its cards hidden. And row 1 covers both launches, because the pipeline needs the orchestrator's transcript before it can name it as creator.

## Wake behaviour after #1881 and #1874

Tour card 4 and §1 of the design now describe the real behaviour. An agent the orchestrator started itself with `spawn_agent` wakes it at the next 5-minute check. A finished stage or stuck work wakes it at most once an hour, because a stage agent's lineage parent is the stage before it, so it never counts as the seat's child. The `SEAT_MISFILED` remedy now says the next check re-files the seat, since #1874 records successions on every tick sweep.

## Verification

- `bunx tsc --noEmit --incremental false`: exit 0.
- Tests run by path with an isolated `LLV_STATE_DIR` (23 pass): `src/lib/onboarding/{healthCheck,healthRoute,marker}.test.ts`, `src/components/onboarding/{CheckStep,OnboardingDialog}.dom.test.tsx`, `src/app/api/routeExports.test.ts`. They pin an empty tick row waking at once for a completed own lane, every failure code at its row with the later rows left waiting, Stop and the run bound, one-run-at-a-time, and the worktree and branch removal on a real temporary repository.
- `bun run build`: exit 0.
- Privacy gate from the merge base, with the fingerprint catalog: PASS.
- Rendered evidence: the onboarding case of `scripts/capture-board-geometry.ts` now walks the Check step at 1440×900, 1280×900 and 390×844, light and dark, English and Ukrainian, on a production build over a seeded home. The frames cover before a run, running, passed, stopped, each of the eleven failures and the opened details. `/api/onboarding/health` is served from records, so no agent is spawned. All 12 combinations passed their overflow, clipping and 44 px checks. The measured record is `evidence/onboarding/geometry.json`.

Not verified: a live end-to-end run against real engines. That needs the running Viewer and its runtime host, so the first real run happens after a deploy.

Refs #1876


Latand:
Fix round for the render critique of the Check step (its six findings, all addressed).

**The pass a new install ends on.** With no orchestrator, row 5 is skipped with "No orchestrator yet." That note is a sentence where every other row carries a single word, and beside the longest label at 390 px it pushed that label to three lines. The note now drops to its own line under the label on a phone, and the empty time column goes with it, so the label is one line and the row two. The state is a capture case of its own now (`check-passed-no-seat`).

**A cleanup that could not finish.** It named what was left and stopped there. A run whose cleanup reports problems now keeps its run file, so the next read or start of the check undoes the rest from what that file names — the sweep then drops the file whatever the second attempt does, so a step that can never succeed is retried once and not on every read. The amber line says that, and what to do if the next run repeats it. Also a capture case now (`check-cleanup-problem`).

**The footer no longer tells the user to fix what it just said they cannot.** `DELIVERY_FAILED`, `WAKE_NOT_OWED` and `WAKE_UNDELIVERED` end with where the check stopped; every other failure keeps "Fix that, then run it again".

**`SPAWN_TIMEOUT` closes both branches:** the version command answering, and not answering.

**The machine detail no longer pushes "Run it again" under the fold.** Opening a detail scrolls the controls into view rather than the detail, and the dialog body carries scroll padding, so the button keeps its breathing room above the footer. The driver requires it.

**Nothing is filled while the check runs.** "Stop" and "Open the board" are both bordered, so the brightest control during a two-minute wait is not the one that leaves the step; the driver expects zero filled accents in that state. A failed row's state word carries the danger colour with its glyph.

Verification: `tsc` clean, `bun run build` green, the three touched test files pass (the run-file retention test was checked red against the previous head), the privacy gate passes, and the capture driver's own acceptance is green at 1440×900, 1280×900 and 390×844, light and dark, English and Ukrainian — 29 frames per combination, committed as `evidence/onboarding/geometry.json`.

https://github.com/Latand/live-log-viewer-next/pull/1897

Latand:
Fix round for the five findings on d851be3.

**P2 — cleanup destroyed its own recovery record.** `cleanupHealthRun`'s last step removed the whole run folder, and that folder is where `writeHealthRunFile` puts `run.json`. It now removes only `seat-tick.json`; the run file is dropped by `dropRunFile` — at once on a clean run, and after the sweep's one retry otherwise. The new test drives the real `cleanupHealthRun` over a temporary state root with `closePipeline` answering "device or resource busy", and asserts `run.json` survives, `readHealthRunFiles` still names the run, and the sweep then removes it. Verified red against the old step.

**P2 — the step polled a 404 for ever after a Viewer restart.** `useHealthCheck.read` now falls back to a read without the run id when a named run answers 404, so the step returns to idle and the interval stops. A request that merely failed still retries. Two DOM tests: the 404 case (polling stops, spinner goes, "Run the check" returns) and a 503 case (the run and the poll are kept). Verified red against the old `read`.

**P2 — two failure states painted an unsubstituted placeholder.** `filingVerdict(null)` is now `SEAT_UNREADABLE` with its own copy in both locales, instead of borrowing `SEAT_MISFILED`'s sentence and naming a cause nobody observed. The whole-run bound is `RUN_BOUND`, raised by its own error class and given a footer of its own because it can stop the first row as well as a later one; the catch-all carries the engine and command params every row-1 sentence names. The capture's guard now reads the whole failure block rather than its first paragraph, and both states are captured in both locales.

**P3 — `marker.lastHealth`.** The menu dot is slice 1's fence, so the docstring now says the reader lands with the menu work rather than claiming it exists. `parseLastHealth` is covered: a round trip and each of the four rejection branches, written straight into the marker on disk.

**P3 — two uncaptured rendered states.** `check-cleaning` (settled run, cleanup still running) and `check-start-refused` (a 409 the step reports) are new capture cases. The refused-start frame turned up an untranslated server sentence in the Ukrainian interface, so a refusal the step has its own sentence for is now written in the interface language; the capture asserts the exact sentence per locale.

Gates at this head: `bunx tsc --noEmit --incremental false` exit 0, the six touched test files by path (72 pass) against isolated state directories, `bun run build` exit 0, and the privacy gate from the merge base with `--check-commits` and the fingerprint catalog: PASS. Rendered evidence re-captured at 1440×900, 1280×900 and 390×844, light and dark, English and Ukrainian — 33 frames per combination, acceptance clean.

PR #1898, opened 2026-09-19
Orchestrator seats get their own place on the board; the panel docks and collapses; columns expand (#1841)

### LINKED ISSUE #1841: Board: orchestrator seats leave the task columns and get a place of their own, previous seats included
## Operator request, 2026-09-19 09:05Z, orchestrator seat chat (paraphrased in English)

> I do not like orchestrators appearing in my list of tasks. An orchestrator has nothing to do with a task. Either do not show it there, or give orchestrators a separate column, if that makes sense given the orchestrator panel at the top. Previous seats should not sit there either. I still need to be able to see them, so they need a special place, outside the other tasks. I can see a manager seat from 20 hours ago and an orchestrator seat from 6 hours ago.

## Today

Every orchestrator seat launch mints a board task ("Orchestrator seat …", "Manager seat …", "orchestrator · You are the viewer's built-in Manager …") with the seat conversation as its assignment. It draws a band in the Assigned column beside product work, and it stays there after a rotation, because nothing ever marks it done. On 2026-09-19 the board carried five such cards (tasks 71182567, f4e939ef, c4be6c78, 9feead69, 08ace29d); the seat closed three by hand. The seat's own card also cannot be hidden (`TASK_HIDE_PROTECTED`).

## What to build

1. **Seat conversations draw no band in the task columns.** A task whose only agents are orchestrator seats (current or rotated out) is excluded from Inbox / Assigned / Blocked / Done, from the column counters and from "N working".
2. **A place of their own.** The board already has the orchestrator panel at the top. It becomes the home of seats: the current seat as now, plus a compact "Previous seats" disclosure (closed by default) listing rotated-out seats newest first — title, when it held the seat, a click opens that conversation. Nothing else is invented: no new column unless the panel proves unable to hold it, and then say why in the PR.
3. **Rotation retires the old seat's card by itself**: when a seat is rotated out, its card moves to "Previous seats" with no manual step.
4. The seat's working notes (the long rules-and-state text a seat keeps) stay readable from that place; they are agent text and belong to Details once #1834 lands.
5. Work that a seat launches keeps grouping under the outcome's task, exactly as now.

## Acceptance

- A project with one live seat, two rotated-out seats and three product tasks shows three bands in the columns and zero seat bands; the panel shows the live seat and "Previous seats (2)".
- Rotating the seat moves the old one into the list with no board write by an agent.
- DOM tests go red without the change; rendered evidence at desktop and 390x844, once, at the end.

Related: #1801 (orchestrator panel placement and collapse) touches the same panel — one owner for `OrchestratorPanel` at a time.

Latand:
Builds section 2 of `docs/design/ui-batch-2026-09.md`: orchestrator seats leave the task columns and get their own place, the seat panel docks and collapses, and a shelf column can take the wide share. The design planned this as two PRs (A+B, then C+D). This pipeline stage ships all four parts in one PR, one commit series per part.

Closes #1841.

## What changes for the operator

**A. Seat conversations draw no band.** Every conversation the project's seat record names (the live seat, a pending intent, every revoked seat) leaves the bands before anything is counted. A task holding nothing but seat conversations draws no card, adds nothing to any column counter, `N working`, the Hidden tray or `Tasks N`, and has no row in the Tasks panel. The task row stays in the store, because that is where the seat keeps its notes. A band that mixes a seat with real work keeps its card and loses only the seat's tile. A band with no task whose only conversations were seats is gone too. A rotation retires the old seat's card with no write, because the revocation it already records is what moves it. If the seat read fails, the bands draw as before, seat included. The phone board applies the same filter. The filter is not limited to the open project: the seat read carries every project's seat conversations, so the Tasks panel in its `All projects` scope leaves other projects' seat tasks out too, and the Overview board — which names no project and so cannot ask the per-project read — asks for that set on its own (`?scope=all`) and keeps seat cards, and their share of its hidden count, off it.

**B. Previous seats in the seat head.** The `Stays on the board` lock pill is gone. In its place is an outlined 28 px `Previous seats N` control. It opens a 340 px popover with the live seat under `Current` and the previous seats newest first. Each row shows its span (`18 Sep 01:01 – 19 Sep 05:01 · 28 h`, `until …` for a revocation recorded before spans existed), links to its conversation, and has `Notes ›`, which expands that seat task's `details` in place (one row open at a time). With a live seat and no previous ones, the control reads `Seat notes`. It is hidden at zero and while the seat read is loading or failed. On the phone, the seat sheet gains a `Previous seats 2 ›` row that opens a list screen, with the notes as a sub-screen.

**C. Placement and collapse.** A quiet icon in the seat head moves the panel between **Top** (as before) and **Side**: a full-height 380 px column left of the columns, resizable from 320 to 560 px on its right edge. The choice is remembered per browser in `llv:kanban-seat:v2`, and the v1 height and collapsed flags carry over once. Collapsed on top, the panel is a 40 px strip with the state word, the unread marker, the Previous seats count and the placement switch. Collapsed at the side, it is a 44 px rail: one button showing the state dot and the unread dot. `O` toggles collapse while the board has focus and no field, sheet or menu does. While collapsed, the header's Orchestrator toggle carries a 6 px dot in the seat's state colour. At the side, the board's layout mode now accounts for the panel's width.

**D. A column can take the wide share.** Each shelf's head has `Widen {column}`. The widened shelf takes the workspace width and dot grid, and Assigned takes the shelf's share and fill. Only one column is wide at a time. A wide shelf gives the space back when the operator works in Assigned again (a pointer down or a focus on a card there); reading, scrolling or opening a card inside the wide column never narrows it. `Keep wide` pins it through that and across reloads (`llv:kanban-wide:v1`). In scroll mode, the same control swaps the 480 and 280 px bases. Tabs draw no control.

All copy is in English and Ukrainian, as the design's tables give it. `orchPanel.seatStays*` are deleted.

## Server

- A revocation now also records the ended seat's `activatedAt`, `path` and `engine`. Older revocations read back unchanged.
- `previousOrchestratorSeats(project)` returns the revoked seats, newest first, at most 20, each conversation once, never the live one.
- `GET /api/orchestrator/seat` adds `previous[]` (with `title`, `taskId` and `hasNotes` from the task whose assignment names the seat) and `currentTask` (the live seat's `taskId`, `title` and `hasNotes`). `hasNotes` and the current seat's title are on the answer because the phone's seat screens carry no task list: without them they cannot tell a seat with notes from one without, or name the live seat at all.
- The same answer carries `all`: every conversation ANY project's seat record names, split into the seats held now and the seats retired, from the same read. `GET /api/orchestrator/seat?scope=all` answers with that set alone, for the Overview, which names no project. Null when the record cannot be read, which hides nothing anywhere.
- The status read takes ONE read of the seat record and derives all three parts of the answer from that parse (`orchestratorSeatIn`, `previousOrchestratorSeatsIn`, `allSeatConversationsIn` beside the reading functions). The task store is read once, and only when there is a seat to find notes for.
- The rotation and designation logic, `src/lib/attention/`, the task store's write paths and `TASK_HIDE_PROTECTED` are untouched.

## Fit with the header and the stage labels

The header (`board-header.md`) and the stage card labels (#1887) are not redesigned. The only header change is the one §2 C asks for: the state dot on the Orchestrator toggle's icon (`BarPanelToggles` in `ProjectBar.tsx`; the design's `OrchestratorPanelToggle.tsx` does not exist). The bar stays 48 px in every capture. At the 380 px side placement, the head holds its title, state, Previous seats (as a count) and both controls on one row: Fold draws as its chevron there, with its word kept in the label and tooltip. The incumbent line and host controls take a second row. On top, the head is as before.

## Verification

- `bunx tsc --noEmit --incremental false`: exit 0.
- Touched test files, by path, against an isolated state dir: 20 files, all green. New cases:
  - `kanbanModel.test.ts`: live + two revoked seats + product tasks give zero seat cards and counters without seats; the mixed band keeps its card without the seat tile; taskless seat bands; a failed read draws as before; rotation with no task write.
  - `seats.test.ts`: the previous-seat reader, span fields on new revocations, old revocations unchanged, the seat task lookup.
  - `seatFold.dom.test.tsx`: both placements and their collapsed shapes; the v1 to v2 migration; the popover closed by default, with its count, order, links and one open Notes row; hidden at zero; `Seat notes`.
  - `KanbanBoard.dom.test.tsx`: widen, one wide at a time, auto-narrow on a card focus in Assigned, the pin surviving it and a remount, no control in tabs, `O`.
- `bun run build`: passes.
- Privacy gate from the merge base: PASS.
- Rendered evidence: three new cases on the existing driver `scripts/capture-board-geometry.ts` (`BOARD_CAPTURE_CASE=seats | seat-placement | columns-wide`), run against the production build on a port the driver binds itself, with playwright-core and `CHROME_BIN=google-chrome-stable`. The seeded home under `/var/tmp` has two projects, a live seat, two retired seats with notes, and product tasks across all four columns. Records are in `evidence/issue-1841/`; no rasters are committed. All three pass at 1440 and 1280 (en, uk; light, dark), `columns-wide` also at 1920 (the only width where the board leaves scroll mode beside the rail), and `seats` also at 390 × 844. The checks read red before the fixes they drove: the first `seats` run found taskless seat bands still drawn in Inbox, and the first `seat-placement` run found the side rail hidden by the folded-panel rule and the side head overlapping. Both are fixed here.

## Review round 2

Four findings, all the same shape: a fact one surface knew was guessed on another.

- **The Tasks panel and `Tasks N` left the seat's tasks out only on the Board face.** The seat answer they read belonged to the board, and on the Conversations face the board is not mounted, so the filter emptied and the orchestrator rows came back into the list and the count. The desktop's seat read now belongs to the dashboard, which owns the panel and both faces; the board takes that answer as a required `seatRefs` prop and has no read of its own left. That left the dock's own panel — a sibling of the dashboard, open on the Conversations face — reading the route a second time per interval; round 3 below is where the poll itself was made shared. New DOM cases mount the dashboard on both faces — including one that opens straight onto Conversations, where nothing used to read the seat at all.
- **A retired seat left the phone board by its stored path only.** A stored revocation path goes stale on its own (the identity wave rekeys the seats and not the revocations; a conversation that moves accounts keeps its id and changes its transcript), and preferring it drew the retired seat as a row beside product work. `retiredSeatPaths` now takes both paths and is a tested function; the case asserts through the phone board's own projection.
- **A seat with no notes still drew `Notes` where the page carried no task list.** That was optimism, and on the phone it opened an empty box. The answer now says `hasNotes` for every seat it names, so the control is drawn only for notes that exist. If an on-demand read lands empty anyway (the task was edited after the answer), the control stays while its row is open and the row says there are none, rather than vanishing under the panel it opened — one new string, `orchPanel.seatNotesEmpty`, in both languages.
- **The phone showed the live seat as `Orchestrator seat`** while the desktop popover showed its task's title, because the phone builds its rows with no task list. `currentTask.title` on the answer names it. The re-captured evidence is where this shows: the phone list's live row read `Orchestrator seat` and now reads `Orchestrator seat, board batch`.

Re-verified at the fix commit: `tsc` exit 0, `bun run build` passes, the privacy gate from the merge base PASSes, the touched files green by path against an isolated state dir, and all three evidence cases re-captured with `failures: []`. Each new case was checked red first — against `HEAD`'s dashboard and hook for the panel cases, and against the replaced predicates for the notes, title and stale-path cases.

## Review round 3

Three findings: one behaviour regression and two about the record.

- **The desktop polled the seat route twice per interval on the Conversations face.** Making the dashboard read the seat on every face (round 2) was right, but the dock's panel is a sibling of the dashboard, not a child, and it starts a read of its own — and every MOUNT of `useOrchestratorSeat` owned an interval, so the shared answer map shared the answer and not the request. The poll now belongs to the project and cwd: the first surface for a key starts it, later surfaces join and are published the same answer, and it stops when the last one leaves. An arriving surface still revalidates once, and arrivals in the same commit share that one read. This also collapses the phone's pre-existing overlap between the focus view and the seat card under it. A new DOM case mounts the dashboard on the Conversations face with the dock open and asserts one request per interval with the dock still painting the live seat; it reads red against the previous head, at two.
- **The status read read and re-parsed one document three times.** The active seat, the previous seats and the cross-project set each read the record for themselves — on a poll this PR had also made heavier. One read now, all three derived from that parse. The route test counts both: one parse of the record per request in all three shapes (project, `?scope=all`, a project with no seat), and one call into the task store when there is a seat, none otherwise. Red against the previous head, at three parses.
- **The three evidence records named three different commits.** All three cases are re-run at the head that carries the changes above, so each record names the last commit that changed anything a seat surface renders; all pass with `failures: []`. Nothing that follows them on this branch renders anything: the evidence commit itself, a merge of `origin/main` whose only overlap with this PR is the capture driver's unrelated onboarding case, and an indentation fix in a test file.

Re-verified on the branch head, after merging `origin/main` (the build at the commit before the test-only indentation that follows it): `tsc` exit 0, `bun run build` passes, the privacy gate from the merge base PASSes, and the touched suites green by path against an isolated state dir (the seat route, the seat hook's own cases, the dashboard's seat cases, the phone's seat surfaces, the seat library and the fold). `src/lib/mcp/orchestratorTools.test.ts` has four failures that are equally red at this branch's previous head and at the merge base, so they are not from this round; each of the four passes when run alone, and that is filed as #1906.

## Where this deviates from the design

§2 B says a revocation written before spans existed falls back to the conversation's first message time and otherwise omits the start. The fallback is not implemented: such a row reads `until {to}`, which is the row the same section's format paragraph specifies for it. Reading a first message time per retired seat would mean up to 20 transcript reads on every tick of a status poll that this PR has just finished making lighter, and it applies only to revocations recorded before this branch. If it is wanted, it belongs with a stored `activatedAt` backfill rather than on the read path.

## Left for follow-up

- A seat's notes are read from the task list the page already holds. When that task is not in the list, they are fetched from `GET /api/tasks`; there is no single-task read route.


Latand:
### Fix round: the Overview, the panel's «all» scope, and the notes well

All three findings from the previous review are addressed.

**P2 — the Overview still listed every seat under Hidden.** The rule only ever had the *current project's* seat to work with, and the Overview names no project, so it was handed none and filed every seat-only task as an off-board task. The seat record is one document, so it is now read whole: `allSeatConversations()` answers every project's seat conversations — held, pending and retired — from a single read, and `null` when the record cannot be read, which hides nothing anywhere. `/api/orchestrator/seat` carries that set on its per-project answer (so the dashboard makes no second request) and answers it alone for `?scope=all`, which is what the Overview's board asks for on the panel's own slow cadence. On the seeded home the Overview's hidden count is now 0 with the tray naming no seat, at 1440 and 1280, en and uk, light and dark.

**P3 — the Tasks panel's «all» scope.** The panel, its `Tasks N` count and the phone's task screens and count now all filter with the cross-project set, so a seat of another project is no more a task there than this project's is.

**P3 — the dark Notes well.** The open notes block carries a 1 px `--border-default` edge and the deeper `--surface-board` fill, so it reads as an inset instead of blending into the row under the pointer. The phone's notes screen gets the same edge. The evidence records how the block is painted: dark `rgb(13,13,17)` fill, `rgb(38,38,46)` border, against a row at `rgb(19,19,24)`.

**Verification.** `bunx tsc --noEmit --incremental false` (exit 0), `bun run build` (exit 0), the privacy gate, the touched tests by path (136 pass) — including a new red-first case in each of `seats.test.ts`, `seatState.test.ts`, the seat route test, `OverviewBoard.kanban.dom.test.tsx` (the Overview's hidden count and seat card, verified to fail with the seat read removed) and `ProjectDashboard.seatTasks.dom.test.tsx` (the «all» scope, verified to fail with the cross-project set removed) — and the `seats` capture case, extended to read the hidden tray on the project board and on the Overview and the notes paint, re-run at this head with no failures.

Head: ea7f01f0c801be7611780c11eff201994e8d26c1

Issue #1870, opened 2026-09-19
A hard reboot zeroed the task store: durable writes, rolling backups and a restore that replays updates


Latand:
## Observed (2026-09-19)
- 14:14–14:15Z: the kernel OOM killer took down the desktop session (browser, Telegram, the GNOME session scope); the machine hard-rebooted at 14:16Z. Six pipeline agents were running tsc and production builds at once, swap had been above 19 GB all day.
- After boot the Viewer's task store was a file of NUL bytes ('JSON Parse error: Unrecognized token' on every task route, and on a pipeline's task-membership write, which parked lane 5f800101). The WakaTime activity state file was zeroed as well.
- An agent rebuilt the store from a full snapshot dated 2026-09-12 plus stored mutation results: 1 273 tasks. Creations came back. **Updates did not**: 96 tasks had regressed status (66 closed cards back to blocked, 12 back to assigned, …), titles, descriptions and `details` written after the snapshot were gone (the orchestrator seat card lost four hours of state), board visibility was reset, and 178 tasks that existed before the crash were missing (mostly launch placeholders). The seat repaired statuses only because it happened to hold an id→status map read four minutes before the crash.

## Cause to confirm
The store is rewritten in place, or renamed without fsync of the file and its directory, so a crash shortly after a write leaves a zero-length or NUL-filled file. Nothing keeps a recent good copy.

## Wanted
1. Every Viewer state file that is rewritten whole (task store, WakaTime state, accounts registry, limits cache, worktree map, pipeline store if it is one) goes through one helper: write temp file, fsync it, rename, fsync the directory.
2. Rolling backups of the task store (for example the last 10 generations plus one per hour for a day), written by the same helper, so the newest readable generation is minutes old.
3. On load, an unreadable store falls back to the newest readable generation by itself, keeps the damaged file, and raises a visible incident; it never serves 500s until a human repairs it.
4. A test that kills the writer between write and rename, and one that feeds a NUL-filled file, against the production store module.

## Related
The restart also minted placeholder cards for old stage conversations whose task links were lost, and twelve identical 'Seat tick wake unresolved under its original key' inbox tasks at 14:31Z; both are covered in the follow-up issue on restart recovery.

Latand:
[...]

PR #1901, opened 2026-09-19
The board lives in SQLite, imported once from board.json (#1870, slice 3)

Latand:
Slice 3 of #1870, following `docs/design/state-sqlite-migration.md` §6.2 and §8. The board store moves from `board.json` into the `board` collection of `state.sqlite`, one row per project, reusing slice 1's import helper as it is.

## Why

`board.json` was rewritten whole on every pin, hidden group and view-mode change — 2.5 MB across 63 projects through a temp file and a rename. That is the write pattern that came back zero-filled after the 2026-09-19 crash. A board write now commits only the row it changed, in one WAL transaction with `synchronous=FULL`.

## What changed

- **`src/lib/board/store.ts`** keeps every exported signature. The first read imports `board.json` under the legacy file's own write lock, verifies row count and digest inside the transaction, keeps the file as `board.json.imported-<release>` and leaves a tombstone directory in its place, so an old-release writer fails with `EISDIR` instead of writing a board nothing reads. Bytes that are not JSON at all import as a recorded gap with an incident and the board serves empty; valid JSON whose project state the board cannot mean still refuses the import and leaves the file untouched.
- **The board's `board.json.write-lock` queue is gone.** Writers serialize on the collection lease, and a lease whose owner was killed mid-write is reclaimed.
- **Rollback.** The demotion checkpoint writes `board.json` back from one SQLite revision and records its digest. The mirror carries that revision in a reserved project entry, which an older release carries through its own writes, so roll-forward can tell a project the rollback release deleted from one a crash left out of the file. The merge decides who changed a project from row revisions against the mirror, keeps a project SQLite moved further, and raises the incident with the counts.
- **`src/lib/state/legacyCollections.ts`** gains the one-line board entry, so the activation import and the demotion mirror pick it up.
- **`scripts/deploy-staging.ts`** fingerprints `state.sqlite` in place of `board.json`.

## Tests

`src/lib/board/store.sqlite.test.ts` runs the design's failure matrix against the production store module in mkdtemp state directories: a writer killed mid-transaction (a), a NUL-filled legacy file (b), the three half-finished import crash points (c), two processes importing at once (d), an old-release writer hitting the tombstone (e), rollback then roll-forward (f), and an injected digest mismatch (g). Beside them, the two the delivery plan names for this slice: a per-project write leaves every other project's row byte-identical and unmoved, and the retired-key history cap still holds across a reload from SQLite.

`src/lib/board/store.legacyFile.test.ts` drives the legacy-file cases through the store API that predates the move, so they fail on behaviour at the merge base.

## Verification

- Touched test files by path, in isolated state directories: board (85), legacy collections (2), deploy-staging (9), durable migration, MCP bindings (71), board route, coordinator, reaper runtime, scanner discover. Failures outside the board are at merge-base parity, compared file by file against the old store.
- `bun run build` — clean, including the MCP bundle.
- Privacy publication gate — PASS.
- Realism, against a copy of the live board taken read-only (63 projects, 2.48 MB): imported in 33 ms with a matching row digest and source sha, every project read back with the same placements, one project's write left every other row byte-identical, and the demotion mirror round-tripped with no incident.

## Review round 2

Two behaviour regressions against the `board.json` store, both proved through the exported store API in mkdtemp state directories, head against merge base.

- **A chained migration folds through again.** A catalog plan can hold both `A -> B` and `B -> C` — `migrationPlan` stops walking a chain at an intermediate that has conversations of its own, so that intermediate is a target and a source in one map, and durable alias candidates merge into the same map afterwards. The rewrite accumulated each group's edits into `changed` and `deleted` collections the next group's snapshot never saw: one shape named a project as both a written row and a deleted one, which the collection rejects, so a plain `Error` escaped the store and the scanner turned it into `boardHealed=false` on every scan; the other stranded the source's placements on a project key the catalog had just migrated away from, while reporting success. Every source now reaches the end of its chain before the groups are formed, so the outcome no longer depends on the map's iteration order either. A cycle names no final target: its members are left intact and the call reports incomplete.
- **The import-marker probe retries when the file lock is held.** On the first touch of the board in a process the probe runs on a read-only connection opened with `busy_timeout = 0` and no retry loop, so a writer holding the lock made it raise `SQLITE_BUSY`. That is neither a `BoardStoreError` nor a `FileTransactionBusyError`, so it escaped `boardFor`, `patchBoard` and `mutateBoard` untranslated and the write was lost outright, where the `board.json` store queued on its own write lock and never dropped it. The probe now retries on the same bounded schedule the collection's own writes use, in application code with the connection's timeout left at zero, and a `SQLITE_BUSY` that still escapes is reported as the store's own busy failure. No second import mechanism.

New cases in `store.sqlite.test.ts`: a chained migration with and without a row on the intermediate, a migration cycle, a write whose probe meets a lock another process holds and releases, and a lock nobody releases. Neutering the retry and the translation turns the last two red; the chained-migration cases are red at the branch head before this round.

Re-run: `bunx tsc --noEmit --incremental false` clean, `bun run build` clean, privacy gate PASS with the fingerprint catalog loaded. Board test files green (68). The 40-writer concurrency case ran 20 consecutive times green. `scanner/discover`, `api/board/route` and `accounts/migration/coordinator` are at merge-base parity, compared run for run in a worktree at the merge base.

## A file outside this slice's fence

`src/lib/accounts/migration/coordinator.test.ts` belongs to slice 7. One assertion there observed the board's old `board.json.write-lock` queue, which this slice removes, so it now observes the state database signature instead, and one fixture that relied on unparseable bytes refusing a read now uses valid JSON the board cannot mean. `origin/main` carries no slice 7 commit as of this head, so the change merges clean; worth re-checking immediately before the merge.

## Downgrading

A release older than this one fails on the `board.json` directory, naming the path; upgrading again recovers. A deployed release rolled back through the release fence gets its `board.json` written for it. The CHANGELOG carries this note.


Latand:
Review round 2 addressed at `304ebfc6a00e5aad446af40fd7e4ca61fe7a2726`.

**P1 — chained project migrations.** Every source now resolves to the end of its chain before the groups are formed (`resolvedMigrations`), so `A -> B` plus `B -> C` folds A and B into C in one patch, no project is ever named as both a written row and a deleted one, and the outcome no longer depends on the migration map's iteration order. A cycle names no final target, so its members are left intact and the call returns `false` rather than throwing. Cases: `a chained migration folds through to the final target, with and without a row on the intermediate` (asserts the final target carries both sources' placements, both sources are gone from `state_rows`, and an untouched project's row is byte-identical with its revision unmoved) and `a migration cycle leaves both projects intact and reports incomplete`. Both are red at the previous head with `invalid board row deletion`.

**P1 — busy import probe.** The import-marker probe retries on the same bounded schedule the collection's own writes use (6000 × 5 ms, `Atomics.wait`, connection timeout left at zero), and `asBoardStoreError` now reports a `SQLITE_BUSY` that still escapes as the store's own busy failure. Cases: `a write whose import probe meets a held file lock waits for it instead of failing` and `a probe that never gets the file lock reports the store's busy failure, not a raw SQLite error` — the second asserts a `BoardStoreError` whose cause matches `database is locked`. A child takes the lock with `PRAGMA locking_mode = EXCLUSIVE` plus `BEGIN IMMEDIATE`, which makes a read-only connection raise deterministically; neutering the retry and the translation turns both red. The 40-writer concurrency case ran 20 consecutive times green. Note the flake did not reproduce here at the previous head in 5 runs, so the mechanism, not the run count, is what the two new cases prove. The identical probe in the task store (slice 1, on main) is filed as #1904.

**P3 — the file outside the fence.** `origin/main` carries no slice 7 commit at this head, so `src/lib/accounts/migration/coordinator.test.ts` merges clean; it is now named in the PR body with what changed and why, for re-checking immediately before the merge.

Gates at this head: `bunx tsc --noEmit --incremental false` exit 0, `bun run build` exit 0, privacy gate PASS with the fingerprint catalog loaded, all PR checks green. `scanner/discover` (8/38), `api/board/route` (43/23) and `accounts/migration/coordinator` (106/9) are at merge-base parity, run for run in a worktree at `fea407ca1`.

Issue #1905, opened 2026-09-20
A lane's production build can migrate the LIVE state directory: builds and tests must never resolve the operator's config dir


Latand:
## Incident (2026-09-20 00:02:42Z, production machine)
A pipeline builder working on the accounts slice of #1870 ran `bun run build` in its worktree. During the build the new first-boot import executed against the operator's real state directory (`~/.config/agent-log-viewer/state`): eight account state files (`claude-accounts.json`, `codex-accounts.json`, `account-project-bindings.json`, `account-project-overrides.json`, `spawn-admission-fences.json`, `account-mutation-revision.json`, `claude-auth-operations.json`, `codex-login-attempts.json`) were imported into the production `state.sqlite` and replaced by README directories. The production Viewer, still on the JSON readers, answered every account selection with 'account-project-bindings.json is unreadable (EISDIR)'. For about seventy minutes no agent could be spawned in any project on the machine; three lanes parked at their review spawn. The seat restored the files from the `.imported-*` copies (nothing had changed since the import), kept the stray directories aside, and accounts read again.

The lane's spec forbade writing to the live config directory and the builder did not intend to: nothing in the build says it touches state. Prose cannot prevent this; the mechanism has to.

## Causes
1. The import runs on a code path reachable while `next build` loads modules or collects page data, instead of only in the serving Viewer's activation.
2. A build or a test that does not set a state directory silently resolves the operator's real one.

## Wanted
- A state-mutating startup step (import, migration, backup, cleanup) runs only when the process is the serving Viewer (or the runtime host) that owns the release fence; at build time (`NEXT_PHASE=phase-production-build`), under `bun test`, and in scripts it refuses with a clear error.
- `bun run build`, `bun test` and the capture drivers set an isolated throw-away config/state root by default; resolving the real one needs an explicit opt-in that only the launcher and the deploy adapter pass.
- The Viewer's spawn boundary gives every pipeline stage an isolated `XDG_CONFIG_HOME`/state root for its own builds and tests, while the agent's MCP link keeps talking to the real Viewer.
- A test: importing the instrumentation and store modules with the build-phase env set, and with no state dir set, performs zero writes outside a temp dir (assert with an fs spy or a read-only directory).

## Leftover to clean before slice 7 ships
[...]

PR #1909, opened 2026-09-20
A turn that died unauthorized says so, instead of reading "Task completed"

Latand:
## What the operator saw

A first turn ended 1.3 s after the mandate reached the engine, with no
assistant message at all, and the window read: Launched · First message
delivered · Task completed. Nothing said the turn had failed.

The turn-end record said it plainly — no agent message, an `unauthorized`
error, an expired sign-in — and the feed threw that half of the record away:
every Codex `task_complete` rendered as the completion note, whether the turn
produced an answer or never ran.

## The change

`src/components/feed/parse.ts` reads the failure a turn-end record carries and
emits its own row for it; a clean completion has no error field at all, so a
turn that really completed is untouched. The row is the danger-toned alert
anatomy the feed already uses for a question that needs the operator: what
failed, the provider's own sentence through the feed's existing
`redactSecrets`, and — when the sign-in expired — the line that says to sign in
again or move the conversation to another account. English and Ukrainian.

What deliberately did not change: the delivery receipt. It said the message was
delivered because it was; acceptance and inference are separate facts, and this
makes the second one visible without touching the first. The success terminal
keeps its quiet note and a cancelled turn stays a cancellation.

## Evidence at this head

Commit `d70683f40f470d28288591d6553936fde7e5321c`.

- `bun test src/components/feed/parse.test.ts src/components/feed/FeedItem.mobile.dom.test.tsx src/lib/i18n/i18n.test.ts` — 246 pass, 6 fail; the six are phone cases that are red at the merge base too (see the limitation below).
- Red-before-green: the four new parser assertions fail on the merge base (`expected a failed-turn row`) and pass here; the completion and cancellation controls pass on both sides.
- `bun run build` — clean, MCP bundle included.
- `bun scripts/privacy-publication-gate.ts --base origin/main` — PASS.
- Rendered evidence: a case on the conversation window's existing driver (`LLV_CONVERSATION_BROWSER_TEST=1 CHROME_BIN=/usr/bin/google-chrome-stable`), at 390 px and 1280 px in English and Ukrainian, failed and clean terminals. Geometry committed at `evidence/issue-1846-auth-terminal/terminal.json`; frames were inspected as pixels, not only as assertions.

## Limitations, stated rather than implied

- **This is presentation only.** It makes a failed first turn visible. It does
  not propagate the host diagnostic, does not invalidate an account's
  availability on observed unauthorized evidence, and fixes nothing about the
  migration stall those probes were chasing. Please do not read it as any of
  those.
- **Scope of the classifier.** Only the Codex turn-end record type observed in
  the incident is read for a failure. Other turn-end shapes still render as
  they did.
- **Six pre-existing red tests** in `FeedItem.mobile.dom.test.tsx` (and more
  across the feed's other `*.mobile.dom.test.tsx` files) fail at the merge base
  as well: their `matchMedia` stub still keys on a width the layout stopped
  using, so `useIsMobile` never reports the phone. The new cases here drive the
  production media query themselves instead of widening this PR to repair a
  stale harness; that repair also surfaces one genuinely stale assertion, which
  belongs in its own change.
- Issue #1846 is closed; this is a follow-up to the operator's recurrence on it,
  and nothing on that issue or on #1854 was reopened or altered.


Latand:
## Review round 1, both P1 findings fixed

Both were right, and the second one was worse than it read: the fallback value was neither sanitized nor bounded, so an error code with a payload behind it went to the screen whole.

**What changed.** Every value the row can display — the provider message and the `codex_error_info` fallback — now goes through `redactTranscriptText`, the redactor the rest of the feed already uses, and is bounded afterwards. `redactSecrets` alone only recognises a `key: value` shape, which is why it walked past `"access_token": "…"` inside a JSON body and past `Bearer …` in a quoted header.

The order matters and is now fixed in one helper: redact, then cut. Cut first and a credential sitting before the cut survives intact while one straddling it is merely halved; redact first and a cut can only ever fall through `[redacted]`.

**One thing the sentinels surfaced.** Classification still reads the raw record, since what a turn died of is not displayed — and reading it showed that `unauthorized Bearer …` classified as a generic failure rather than the expired sign-in it plainly is, because the code had to be the entire value. The code leads the value now.

**Raw records and delivery facts are untouched.** The receipt still says delivered.

**The explanation still reads.** It names two credential words in prose, and prose is not a `key: value`, so redaction leaves it word for word — asserted on the mounted row and visible in the re-captured frames.

## Evidence at `ffd2ea5748ecf66d815ac897824c49ea42cf8904`

- Five parser sentinels (JSON body, `Bearer` header, fallback sanitized, fallback bounded, cut-after-redaction) and three rendered DOM sentinels on the mounted row, plus one asserting the expired-sign-in sentence survives intact. All eight fail at `d70683f40` and pass here.
- `bun test src/components/feed/parse.test.ts src/components/feed/FeedItem.mobile.dom.test.tsx` — 236 pass, 6 fail; the six are the phone cases red at the merge base (#1910), unchanged by this commit.
- `bunx tsc --noEmit --incremental false` clean; `bun run build` clean; privacy gate PASS.
- Rendered evidence re-captured at 390 px and 1280 px in both languages and inspected as pixels; `evidence/issue-1846-auth-terminal/terminal.json` is unchanged, which is itself the point — sanitizing changed nothing an operator reads.

Build and capture ran under a throw-away `XDG_CONFIG_HOME`/`LLV_STATE_DIR` under `/var/tmp` (#1905).

Latand:
## Review round 2 — the P1 is fixed by changing the trust boundary, not the patterns

The finding was right twice over: the specific shapes leaked, and the approach was wrong. I reproduced it before changing anything — of five shapes a probe tried against `ffd2ea57`, four were still painted:

| shape | at `ffd2ea57` |
|---|---|
| `{'refresh_token': '…'}` (single quotes) | leaked |
| escaped `{\"access_token\": \"…\"}` | leaked |
| `{"a":{"b":{"refresh_token":"…"}}}` (nested) | redacted |
| `{"id_token": "…"}` | leaked |
| code fallback `unauthorized {'id_token': '…'}` | leaked |

Chasing that with more regexes is a losing game, so the row stops echoing the provider at all.

**The new contract.** A failed turn row is keyed on values the parser *recognizes* — a small vocabulary of error codes — and rendered entirely from strings the Viewer authored:

- the title and an explanation of what happened;
- for an expired sign-in, the action to take;
- the recognized code, printed as the vocabulary's own constant rather than as bytes from the record, so a payload glued behind a code cannot ride in inside it. An unrecognized code is not printed;
- one plain line saying the provider's own message is withheld because it can quote credentials — the row is honest about the choice instead of silently dropping it.

Unknown forms therefore land on the safe fallback by construction: there is no path from provider text to the screen.

**Nothing is lost from the artifact.** The record's text is still in the transcript record rows under the same bounded, key-aware redaction every other raw record gets. Raw records and delivery identity are untouched.

## P2, same review

What a turn died of now comes from the code the provider set, not from prose it happened to mention:

- `auth-service-unavailable` → **not** an expired sign-in. The credentials are fine; sending the operator to sign in again would send them after the wrong thing.
- explicit `stream_error` stays a stream error even when its message mentions a successful token refresh.
- recognized auth codes (`unauthorized`, `unauthenticated`, `refresh_token_expired`, …) still classify as auth.
- with no code at all, prose is all there is to go on, so the heuristic remains for that case only.
- an unrecognized code that *leads* with a known one still classifies, without printing anything.

## Evidence at `af438fff7700a8db4e333bb88faf89ba6c338e43`

- Parser: every shape above runs twice — once as the message, once as the fallback — plus a bare credential with no key, a stringified body inside a string, an unrecognized code, and a size case asserting the whole row serializes under 200 bytes because there is no provider string on it to bound.
- Mounted DOM: the same six shapes read off the painted row, plus "only a recognized code is printed", plus EN/UK guidance.
- Old-red / new-green: at `ffd2ea57` the four leaking shapes fail at both layers.
- Pixels: the fixture's failing record now *carries* a single-quoted credential, so the re-captured 390 px and 1280 px frames in both languages prove the absence rather than asserting it. Inspected by eye; `evidence/issue-1846-auth-terminal/terminal.json` updated.
- `bun test src/components/feed/parse.test.ts src/components/feed/FeedItem.mobile.dom.test.tsx src/lib/i18n/i18n.test.ts` — 271 pass, 6 fail; the six are the merge-base-red phone cases (#1910).
- `bunx tsc --noEmit --incremental false` clean; `bun run build` clean; privacy gate PASS.

All app loads ran under throw-away `HOME`/`XDG_CONFIG_HOME`/`XDG_CACHE_HOME`/`LLV_STATE_DIR`/`LLV_CLAUDE_HOME`/`TMPDIR` under `/var/tmp` (#1905).

Issue #1761, opened 2026-09-18
CI audit: where merge time goes, which tests are flaky or redundant, and a cut list


Latand:
Operator request, 2026-09-18: CI costs too much time per merge and the suite holds tests nobody needs. Find out where the time goes and cut it.

## What a merge cost today

Three approved PRs (#1752, #1755, #1759) were merged in sequence. `main` requires branches to be up to date (`strict: true`), so every merge put the next PR BEHIND and forced a branch update plus a full CI round. On #1755 that round went red twice on tests the PR does not touch, each a wall-clock timeout, a different test each run:

- `bun-runtime` — "an ordinary queued send mapped to native keeps its operation through compaction and restart…" (timed out after 5000 ms), then on the rerun "bounded canonical history > bounds pages, enforces one deadline even if RPC never settles…".
- `Windows (native)` — "the CLI names a missing Bun prerequisite before starting the Viewer" (timed out after 10000 ms).

Known trackers for these flakes: #1723, #1715, #1663. The repository holds 1111 test files.

## Wanted: an audit with a cut list, then the cuts

1. Per workflow and per step: duration over the last 50 runs, failure rate, and how many failures went green on a rerun with no code change (flake rate). Name the ten slowest tests and the ten flakiest.
2. Classify tests for deletion: tests that assert prompt or copy wording, tests that re-assert what another test already proves through the same production path, tests of deleted features, tests whose only subject is their own fixture, and timing tests that race a wall clock instead of a fake clock or an injected deadline.
3. Propose the merge-path policy: whether `strict` up-to-date is worth a full CI round per queued PR, or a merge queue / batching several approved PRs into one update does the same job.
4. Deliver the cut list as a comment here with file paths and the reason per file, then a PR that deletes and fixes accordingly. No test is weakened to make it pass; a flaky timing test is either rewritten onto a controlled clock or deleted with the reason stated.

Latand:
# CI audit: where merge time goes, what flakes, and what to cut

**Originating requirement** (operator, 2026-09-18, issue #1761):

> Operator request, 2026-09-18: CI costs too much time per merge and the suite holds tests nobody needs. Find out where the time goes and cut it.
>
> 1. Per workflow and per step: duration over the last 50 runs, failure rate, and how many failures went green on a rerun with no code change (flake rate). Name the ten slowest tests and the ten flakiest.
[...]

Issue #674, opened 2026-07-25
Conversation feed pins an old message as the newest item, below fresher tool cards


Latand:
## Symptom (operator, prod, 2026-07-25)

In a live conversation pane, the **first assistant message of the session** renders as the last item in the feed — below a tool card from minutes ago — and stays there as new records arrive. The operator sees a stale line permanently pinned at the bottom, with the "live tail" affordance offered underneath it.

## What is verified

The transcript is not at fault. For the affected conversation:

- The pinned text exists exactly once, at record 10, timestamped at the start of the session.
- The newest assistant records carry correct, monotonic timestamps minutes later.
- No duplicate of the old record exists anywhere in the file.

So the record ordering is produced by the reader or renderer, not by the writer.

## A concrete suspect

The same transcript contains **133 records with no `timestamp` field at all**: 49 `ai-title`, 49 `last-prompt`, 35 `mode`. If feed ordering sorts by timestamp and treats a missing one as zero or undefined, any comparison involving those records is undefined-ordered, and a stable-sort fallback can drag an early item to the end. This wants confirming or ruling out before anything else.

## Required behaviour

1. The feed orders strictly by the transcript's own record sequence, so an item can never appear after a record that was written later.
2. Records without a timestamp — `ai-title`, `last-prompt`, `mode` — never participate in ordering as if they were dated, and never displace dated items.
3. The newest item is the newest item: with the pane following the tail, the bottom of the feed is the most recent record, not a session-opening line.

## Validation

1. Open a long-running conversation whose transcript contains `ai-title` / `last-prompt` / `mode` records interleaved throughout, and confirm the bottom item is the newest record.
2. Append new records while the pane is open and confirm they land at the bottom in write order.
3. Construct a transcript where an early message and several untimestamped records interleave, and assert ordering in a test rather than by eye.

PR #1951, opened 2026-09-20
Bound the live turn overlay to its in-flight tail

### LINKED ISSUE #1955: McpCallCard squeezes its action title to 1.27 px at phone width
## What happens

At 390 px the canonical `McpCallCard` summary line gives its action title
almost none of the row. Measured in a real browser over an invented sixty-call
turn (`src/components/conversation/liveTurnLongTurn.fixture.ts`), dark, en and
uk alike, on a 330 px row:

| entity chips on the call | title width |
| --- | --- |
| 2 (`link_task_to_pipeline`) | **1.27 px** |
| 1 (`update_task`, `get_pipeline`) | 20.5 – 42.3 px |
| 0 (`create_pipeline`) | 129.5 px |

The readings are committed in `evidence/live-turn-rows/after.json` under
`cases.current.canonicalMcp`, recorded by the existing driver case
("live turn rows on a phone" in
`src/components/kanban/kanbanBoard.browser.test.tsx`).

## Why

`src/components/runtime/McpCallCard.tsx` lays the summary out as one
`flex-wrap` line: the icon, the `MCP · <server>` badge, the title
(`min-w-0 flex-1 truncate`), the entity chips, the outcome mark, the duration
and the timestamp. Everything but the title is `shrink-0`, and `flex-1` gives
the title a flex basis of 0 — so when the line is short the title is the only
item that can give way, and it gives way to nothing. Wrapping never happens
because there is always a zero-width title to shrink instead.

## What was already done next door

`src/components/conversation/LiveTurnRows.tsx` had the same construction in its
live MCP row and was fixed in #1959: the title carries a flex basis
(`grow basis-[10rem]`) so it claims the line, and chips that no longer fit wrap
to the next line of the same block, whole and tappable. The same row now
measures 200 px of 330 in both languages. That file's fence stopped at the
overlay, so the card it hands over to still squeezes — a call's line gets
*harder* to read the moment the transcript replaces the live row, which is the
opposite of the handoff both surfaces are built for.

## Suggested fix

Give the card's title the same basis and let the trailing chrome wrap, then
extend the browser case above to assert the canonical readings the way it
already asserts the live ones (`MIN_MCP_TITLE_PX`, `MIN_MCP_TITLE_SHARE`,
`MIN_CHIP_PX`) rather than only recording them. Both surfaces should land on
one shared summary-line layout so they cannot drift apart again.

Latand:
## The defect

In a long Claude turn the live overlay painted dozens of rows under the
canonical cards: tool rows reduced to `<tool> · arguments omitted`, prose
reduced to `… (+N characters)`, only the newest rows keeping any detail. The
cards below already said the same things properly, so the overlay read as a
wall of duplicated junk.

Reproduced at 390 px over an invented sixty-call turn
(`src/components/conversation/liveTurnLongTurn.fixture.ts`: long Viewer MCP
names, failures, shed arguments, one call still running).

| | before | after |
| --- | --- | --- |
| live rows painted | 63 | 8 |
| overlay height | 1468 px | 221 px |
| rows reading only "arguments omitted" | 48 | 0 |
| collapsed lines | 0 | 1 (`55 earlier steps` / `55 попередніх кроків`) |
| MCP rows with the canonical card's chips | 0 | every MCP row in the tail |
| live MCP action title at 390 px | n/a | 200 px of a 330 px row |
| canonical MCP action title, two chips | 1.27 px | 222 px of a 330 px row |
| live rows repeating a canonical row | 0 | 0 |

Every phase is committed under `evidence/live-turn-rows/`, and each file
records the revision whose tree rendered it: `before.json` the merge base,
`round1.json` the first round's head, `after.json` this branch's latest code
commit. Each was captured from an exported checkout of that revision, driven by
the same driver in the same units, so the phases cannot drift apart again — the
first round shipped before PNGs that were byte-identical to its round-1 ones.
Rasters are not committed; the driver writes them to the directory named by
`LIVE_ROWS_PNG_DIR`.

## What stalls the transcript window — what is established, and what is not

The report came with a banner reading "Runtime degraded · polling". Three
explanations were tested; the honest summary is that the production cause is
**not proven**, and the bound does not depend on which one it was.

**(b) a claim identity mismatch for Claude tool rows — NOT the defect.** A
canonical Claude window built over the same 60 calls claims all of them:
`visibleRuntimeLiveTurnItems` returns `[]`. The parser folds runs into
`cmd-group` rows whose `ids` carry every call id and leaves MCP calls as
standalone `tool` rows with their own id; `toolIdsOf` reads both. Kept as the
first case in `liveTurnOverlayBound.dom.test.tsx`, which also asserts the
window really contains both kinds so the case cannot go vacuous.

**(a) "degraded/polling stalls the canonical feed" — did not hold in the
harness.** `liveTurnStallPath.dom.test.tsx` drives a pane through both buses:
the runtime bus on an injected clock, failing its stream until it settles into
the degraded snapshot poll, and the transcript tail on the log bus's own
polling fallback against a growing file. Held in `degraded` the whole time, the
transcript window still advances by itself and every live row is claimed. The
two buses share no transport, no timer and no pause. What that establishes is
narrow and stated as such in the file: **runtime degradation alone did not
stall a healthy mocked transcript transport.** The transcript transport there
is this test's own `/api/logs`, which answers every poll, so nothing here says
what the operator's transport was doing.

**A path that does reproduce the stall, end to end.**
`liveTurnPaneVisibility.dom.test.tsx` mounts `BranchPane` whole — its own
IntersectionObserver, its own `LogFeed`, the real `useLogTail` subscription —
and drives the pane's visibility rather than a prop:

- a pane that has just mounted is offscreen until its observer says otherwise,
  so it fetches nothing at all before its first sight;
- the observer reports it on screen: the tail subscribes, the window catches
  up, the canonical rows claim every live row, the overlay paints nothing;
- the observer reports it away (past the 256 px margin, or zoomed into the
  board's dormant band): not one byte is asked for while it is away, and the
  turn runs on to sixty calls in the store regardless;
- **the observer reports it back on screen, and the pane paints before its tail
  can fetch anything.** That resume frame — on screen, painting, with the
  window it held while it was away and forty calls unclaimed — is one the test
  constructs. It is shaped like the wall the report carries; nothing here
  proves the operator's wall opened this way. What is on screen in it is the
  bound — a readable tail, one counted line, nothing else — and the test goes
  red when the bound is removed (14 rows painted).
- a moment later the catch-up lands and the overlay retires itself.

So: **pause/resume is one reproduced path; runtime degradation on its own did
not stall a healthy mocked transcript transport; which path produced the
reported window is not established.** Neither file claims the pane's pause is
the only way a window can fall behind, and the fix is deliberately independent
of the answer.

**(c) overflow items reach the renderer — confirmed, and it sets the scale.**
`runtimeLiveTurnItems` concatenates `overflow` with `items`, so the renderer
can receive 544 descriptors (32 active + 512 overflow); 28 of the fixture's 60
rows arrive from `overflow`. `src/lib/runtime/liveTurn.ts` is **unchanged** —
that buffer is what lets a claim retire an old row later and what keeps the
descriptors durable across a journal replay. Dropping it would trade one defect
for another; the bound belongs where the painting happens.

## The fix

The overlay never had a bound of its own — its only bound was the canonical
claim, a condition outside the component. It is now the in-flight tail it was
always meant to be:

- **At most `LIVE_TURN_VISIBLE_ROWS` (8) rows, the newest ones.** At 390 px a
  quiet row is ~20 px, so eight are the ~160 px that fit between the last
  transcript card and the composer without scrolling, and one assistant message
  issues at most a handful of parallel calls, so a whole in-flight batch still
  fits. The window is taken from the end *first* and filtered afterwards:
  pulling an older row forward to fill a slot would paint an old step under
  fresher ones, the ordering defect #674 fixed in the claim path.
- **One quiet counted line for the rest** — `N earlier steps`, localized en+uk
  with plural forms. It carries no `data-live-turn`: it is the absence of rows.
  An explicit omission descriptor folds into that count instead of printing its
  own notice inside a prose row.
- **A call whose arguments the window's bound shed is counted, never listed.**
  It has nothing left but its own name, and a list of those is the defect.
- **Prose the window's text bound shed entirely is counted too** (round 4).
  The window trims prose from the START to honour its 64 KiB text bound, so an
  older message can be left with `text: ""` and its whole length in
  `omittedChars`. Such a row is rightly kept out of the list — there is nothing
  to read on it — but it counted as *zero* steps, so two 64 KiB messages
  painted one row and claimed nothing had happened earlier. A descriptor with
  shed characters and no folded-items count of its own is one step now. A
  streaming placeholder no character has reached yet still counts nothing, and
  an aggregate `omittedItems` descriptor still stands for exactly the items it
  folded, never one more for itself.
- **A Viewer MCP row reads through the same `describeMcpCall` the canonical
  `McpCallCard` uses** — the `MCP · <server>` mark, the call's meaning and the
  entity chips — with no result passed, which is the card's own state while the
  call is still out.
- **An unknown outcome is never a success.** The MCP row mapped every status
  that was not `run` or `err` to success, so a call whose result the runtime
  journal's bound dropped got a green check and `aria-label="success"`. That
  status means the outcome was lost and may have been a failure. It is now its
  own neutral state, with the localized `outcome omitted` in place of every
  mark that carries a meaning.
- **The action title gets the line, on both surfaces.** Badge, chips and
  outcome mark are all intrinsically sized, so on one wrapping flex line the
  title was the only item able to shrink: at 390 px with two entity chips it
  held 1.27 px of a 330 px row. A flex basis makes it claim the line instead
  and lets a chip that no longer fits wrap whole and tappable — 200 px of 330
  on the live row, 222 px on the canonical card, chips unchanged at 82×24, in
  both languages.

## Scope

**`src/components/runtime/McpCallCard.tsx` is in this lane's scope by the
manager seat's decision.** It sits outside the fence this lane started with —
one class and one data attribute on the card's title span, plus the driver
assertions that hold it — and the round-3 review asked for it to be coordinated
or decided rather than left recorded. The manager seat decided it **stays
here** rather than moving to a lane of its own, so the live row and the
canonical card it hands over to are fixed together, in the same shape, gated by
the same numbers, and the handoff geometry gate in the browser driver stays
with them. Acceptance 3 asks that a live MCP row not change look when the
transcript row replaces it, and that cannot be met on the live side alone: with
the live row fixed and the card left alone, a call's line got *harder* to read
the moment it was claimed. **Closes #1955.**

`LogFeed.tsx` is **untouched** — the bound sits inside the overlay, so the
status bar keeps reading the unbounded list and still names a running tool
whose row is collapsed. `TmuxComposer.tsx`, `conversation/outbox.ts`,
`OutboxBubbles.tsx`, `feed/FeedItem.tsx` and `src/lib/runtime/liveTurn.ts` are
untouched.

Left deliberately:

- `feed.liveOmittedItems` has no consumer now that the descriptor is the
  collapsed count. Locale edits in this lane are append-only, so the key stays;
  removing it is a one-line follow-up.
- The MCP chip markup is duplicated from `McpCallCard`'s `LinkChip` rather than
  imported. The parity tests compare the two renderings, so a drift fails —
  which is how the card's own selector change in this round was caught.

## Tests

- `liveTurnPaneVisibility.dom.test.tsx`: the mounted `BranchPane`, its own
  IntersectionObserver, and the resume frame it constructs.
- `liveTurnStallPath.dom.test.tsx`: the two transports as they really are, and
  the narrow thing that follows from holding one of them degraded.
- `liveTurnOverlayBound.dom.test.tsx`: zero live rows against a canonical
  window carrying the same ids; the bounded tail plus one collapsed line
  against a stale one; no "arguments omitted" in either language; a tail of
  shed calls painting only the count; the no-pull-forward invariant; **the
  fully trimmed prose regression** — two 64 KiB messages through the real
  `normalizeRuntimeLiveTurn`, asserting the descriptor shape the bound produced
  (`text: ""`, `omittedChars > 0`, no `omittedItems`), then one painted row and
  `1 earlier step` rendered in en and uk; and the streaming placeholder that
  must keep counting nothing. Against the previous head the regression fails
  with `earlier: 0`.
- `liveTurnToolRows.dom.test.tsx`: three MCP parity cases against a rendered
  `McpCallCard`, both titles held to the same basis, and the four MCP outcome
  states in en and uk.
- `liveTurnMarkdown.dom.test.tsx`: the folded-items notice on the collapsed
  line instead of inside the row.
- `kanbanBoard.browser.test.tsx`: one `describe` block over
  `liveTurnRowsEvidence.fixture.tsx`, gated by `LLV_KANBAN_BROWSER_TEST=1` +
  `CHROME_BIN` like its neighbours. No new driver file (#1761). It asserts the
  live row's and the canonical card's title geometry and chip size in both
  languages, the outcome marks, the bound and the collapsed line, and it
  records the revision it rendered.

## Gates

- `bunx tsc --noEmit` — exit 0.
- Focused tests, one file per invocation, under an isolated config root:
  `liveTurnOverlayBound` 8, `liveTurnToolRows` 16, `liveTurnMarkdown` 23,
  `liveTurnStallPath` 1, `liveTurnPaneVisibility` 1, `McpCallCard.dom` 6,
  `mcpIdentity.parse` 4 — all green.
- `NODE_OPTIONS=--max-old-space-size=8192 bun run build` — exit 0.
- Local privacy publication gate over the whole diff from the merge base, with
  the known-value fingerprints — exit 0, PASS.
- The browser evidence case in both languages plus the neighbouring
  `codex-phone-rows` case (the other consumer of `McpCallCard`'s row geometry)
  — green; the re-captured `after.json` is byte-identical to the previous one
  except for the source revision it records, which is what a change with no
  geometry of its own should produce.
- Every 390 px PNG of the after phase was opened and looked at.


Latand:
Round 3 — the three requested changes, each with what was actually established.

**1. The attribution of the stale window.** The previous round asserted the
pane's pause through a replacement pane and stated it as *the* trigger. Both
have changed. `src/components/conversation/liveTurnPaneVisibility.dom.test.tsx`
mounts `BranchPane` whole — its own IntersectionObserver with the 256 px
margin, its own `LogFeed`, the real `useLogTail` subscription — and drives
visibility rather than a prop: mounted-and-unseen (no bytes fetched at all),
on screen (window current, overlay empty), away (not one request while away,
the turn runs to sixty calls anyway), and back on screen, where the pane paints
before its tail can fetch anything. That last frame is the observable one: on
screen, painting, forty calls unclaimed, and the bound is what is visible in
it. Removing the bound makes it red at 14 rows. `liveTurnStallPath` keeps the
transport refutation and now says plainly that it cannot source the `paused`
input itself; the exclusivity claim is gone from the PR, from
`LiveTurnRows.tsx` and from both test headers.

**2. The before captures.** Confirmed as reported: the delivered
`stale-390-dark-*-before.png` were byte-identical to the round-1 ones. Every
phase has been re-rendered from an exported checkout of the revision it
measures, and each JSON now records that revision in a `source` field — merge
base for `before`, the first round's head for `round1`, this branch's code
commit for `after`. The before frames now show the reported defect: 63 rows,
1468 px, 48 reading only "arguments omitted", and the canonical two-chip card
at 1.27 px of title. `round1.json` is kept separately and unchanged in
substance. No rasters are committed.

**3. The canonical handoff.** Fixed here rather than recorded, and declared:
`src/components/runtime/McpCallCard.tsx` is one file outside the stated fence,
changed by one class and one data attribute on its title span. Acceptance 3
cannot be met on the live side alone — with the card left as it was, a call's
line got harder to read the moment the transcript claimed it. Both titles now
carry the same flex basis, the driver asserts canonical geometry with the same
`MIN_MCP_TITLE_PX` / `MIN_MCP_TITLE_SHARE` / `MIN_CHIP_PX` bounds it already
applied to the live row, and the two-chip canonical title measures 222.47 px of
a 330 px row (67 %) in en and uk, with chips whole at 82×24. Closes #1955. If
the fence is meant to hold absolutely, the fix is one commit to drop and the
acceptance goes back to unmet — say so and I will split it out.

Gates at this head: `bunx tsc --noEmit` exit 0; focused tests one file per
invocation under an isolated config root (6 / 16 / 23 / 1 / 1, plus
`McpCallCard.dom` 6 and `mcpIdentity.parse` 4) all green;
`NODE_OPTIONS=--max-old-space-size=8192 bun run build` exit 0; local privacy
gate with the known-value fingerprints PASS; the browser evidence case in both
languages for all three phases, and the neighbouring `codex-phone-rows` case
re-run green and byte-identical to its committed evidence. Every 390 px PNG was
opened and looked at.

https://github.com/Latand/live-log-viewer-next/pull/1951

PR #1953, opened 2026-09-20
Board view switch: hold the compact two-segment shape and its selected paint

### LINKED ISSUE #1801: Board: the board/conversations switch is bloated; the orchestrator panel needs a side or top placement and a collapse button in both
Filed by the orchestrator seat, 2026-09-19, from an operator report with a desktop board screenshot.

## 1. Bug — the board / conversations switch is odd and bloated

In the board toolbar (right of "Find a task" and "Hidden") the two-way view switch renders as one wide control reading `▥ board ☰ conversations`: both labels and both icons sit inside a single outlined pill with no visible selected state, so it looks like one strange button and takes more width than any other control in the row. Expected: a compact two-segment switch of the same height as its neighbours, the active segment visibly selected, icons with a text label only where the row has room.

Lead: the view tab strings are `kanban.viewTab` and its sibling in `src/lib/i18n/en.ts`; the control is rendered from the board toolbar in `src/components/ProjectDashboard.tsx` / `src/components/kanban`.

## 2. Feature — where the orchestrator sits, and a way to put it away

Today the orchestrator's conversation (`src/components/kanban/KanbanSeat.tsx`) slides down from the top of the board and stays there, taking the upper half of the screen from the columns. The operator's words: "show the orchestrator either at the side as before, or sliding out from the top as now, but so that it is convenient to collapse it with a button and not show it, both vertically and horizontally."

Wanted:
- Two placements for the seat panel, chosen by the operator and remembered per browser: docked at the side (a vertical panel, as the board had before) or docked at the top (as now).
- In either placement one visible button collapses the panel to a slim handle that keeps the seat's state readable (working / needs you, unread reply marker) and one click brings it back; collapsed, the board gets the space.
- The existing drag handle between the panel and the board keeps resizing the panel in its placement.
- Keyboard: the same shortcut toggles collapse in both placements.
- Phone keeps its own sheet (`MobileOrchestratorSheet.tsx`); nothing changes there.
- Rendered evidence at a desktop viewport: top expanded, top collapsed, side expanded, side collapsed, and the toolbar switch in both states.

Latand:
Closes #1801.

## What I found on main first

Section 1 of the issue — "one wide outlined pill with both labels and no selected state" — no longer describes the code. The one-header-bar work already made the switch a two-segment group, so nothing in the component needed redoing. Measured in a real browser on a production build of this branch's base, on a home seeded to look like production:

| viewport | tier | switch group | pressed fill | quiet fill |
| --- | --- | --- | --- | --- |
| 1280, en, dark | narrow | 66 × 32 px, icon only, name in the title | accent at 10% | transparent |
| 1280, uk, dark | narrow | 66 × 32 px, icon only, name in the title | accent at 10% | transparent |
| 2540, en, dark | wide | 201.9 × 32 px, one label per segment | accent at 10% | transparent |
| 2540, uk, dark | wide | 180.6 × 32 px, one label per segment | accent at 10% | transparent |

32 px is the height every other control in the bar has, the segments are ordinary buttons, and the switch keeps its x when the view changes. At 390 px the phone draws its own bar and no desktop switch, as it always has.

## What was actually missing: nothing held that shape

The existing cases asserted that the pressed side is *announced* (`aria-pressed`) and that the narrow tier drops the words. Neither they nor the browser driver bounded the group's width or looked at what the selected segment *paints* — the driver records the group's box but only checks that its x does not move between views. So a reset that painted both sides alike, or a tier that kept its labels, would have passed everything.

This PR adds the cases that hold it:

- the group is one bordered 32 px box that cannot take the row's spare width (no `w-full` / `flex-1` / `grow`);
- the pressed segment is painted apart from the quiet one, and the paint follows the view when the other side is clicked;
- both segments are focusable `<button>`s with a focus ring and no `tabindex`;
- **width bound**, narrow: each segment is a fixed 32 px square carrying its icon, its name in the `title` and no text at all, so neither locale can widen it — asserted in English and in Ukrainian, bounding the group at the 66 px the browser measures;
- width bound, wide: each segment carries its own label exactly once, so the group can never read again as a single control holding both names.

happy-dom lays nothing out, so the bound is read off the boxes the bar asks for; the pixel numbers above come from the browser run.

Each new case was seen red twice: against a component with the selected paint removed, and against one that ignores the narrow tier. No product source, no locale keys: the strings the switch needs are already there in both locales.

## Rendered evidence

The existing board-geometry driver's `header` case, run against a production build of this branch, passed at 2540, 1850 and 1280 (en and uk, light and dark, Tasks closed and open, a 21-character project name) and at 390 (en and uk, light and dark). Frames and measurements were written to the run directory; no rasters are committed.

## The rest of the issue

Sections 2 to 4 shipped before this lane: the seat's side/top placements with a collapse control, one wide column at a time with a pin, and the bar's task search with `/` focus and per-column match counts. I re-verified section 1 in a render and the others in the code, so this closes the issue.

One thing the captures show that belongs elsewhere: on Conversations the bar reads "22 branches running" where the Board reads "5 working" — two vocabularies for the same corner of the header. Filed separately rather than smuggled into this lane.

## Gates

- `bunx tsc --noEmit` → exit 0
- `bun test src/components/ProjectDashboard.headerBar.dom.test.tsx src/components/ProjectDashboard.viewSwitch.dom.test.tsx` → 27 pass, 0 fail
- `NODE_OPTIONS=--max-old-space-size=8192 bun run build` → exit 0
- local privacy publication gate from the merge base, with commit checking → PASS


Latand:
Follow-up filed for the header wording noted above: #1954.

PR #1981, opened 2026-09-21
Phone: a command card's copy controls stay apart, and the feed never rests on a cut line

### LINKED ISSUE #1978: Phone: a command card's two copy controls overlap, and the sticky task chip clips the first row
Seen while reading the README's phone capture (390 px, dark, a Claude conversation with a Bash tool card; the capture is reproducible from the README capture driver and its seeded demo home).

1. On a command tool card the two copy controls (copy the command, copy the output) sit in the top-right corner stacked with no gap: their rounded boxes touch and partly overlap, and together they cover about a fifth of the card's width over the first output lines. On desktop the same two controls are separated and small.
2. With the sticky task chip at the top of the conversation, the card's status line ("exit 0" and the time range) is drawn half under the chip when the card is the first visible item, so the line is cut horizontally.

Expected: at 390 px the two copy controls do not touch or overlap and do not cover output text (one control with a menu, or the second one moved next to the command line); the sticky chip never clips a row, the scroll position snaps so the first visible row starts below it.

Acceptance: measured at 390 px on a production build with the README demo home: zero overlap between the two control boxes, no output glyph under a control, and the first visible row's box starts at or below the chip's bottom edge. Re-capture the README phone image afterwards.

Latand:
Closes #1978

## What the operator sees

**Copy controls (390 px).** A shell card's command copy control no longer lands on the output's copy control. The command block had one ~20 px line and a 44 px coarse-pointer control pinned 6 px down, so the control hung 30 px into the output block, right over the output's own control. On a coarse pointer the command block is now at least 50 px tall (the 6 px inset plus the 44 px control). The output's control starts 12 px below the command's and no glyph sits under either one. The fine pointer's 22 px controls are unchanged, and the desktop at 1280 measures the same positions before and after.

**The task strip no longer cuts a row (phone).** The feed's scroller starts right at the bottom edge of the task strip, so the line straddling that edge at rest was drawn sliced. On the phone the feed now comes to rest on a line boundary:
- **Following the tail:** a spacer under the last row grows by the part of the cut line still showing (at most one line or one control), so the cut line scrolls off whole.
- **After a drag released the tail:** once the finger is off the screen and momentum ends, the feed moves the shorter way to the nearest boundary. That is a smooth scroll, or an instant jump under reduced motion.

"Line" is ink: the text rects under the edge, plus the box of a row-sized button, so a sliced copy control counts as well. `src/components/feedTopEdge.ts` probes seven columns at the edge. The desktop scroll is untouched.

## Proof

- **Browser test:** a `#1978` case in the phone driver `src/components/mobile/issue1671Evidence.browser.test.tsx`, with the fixture's new `?toolcard=1` mode: a Bash call with output and a task assigned to the conversation, so the strip is up. It measures ink (text rects clipped by every overflow ancestor), independently of the product's probe. It checks 390 px dark and light, plus the 1280 px controls:
  - control overlap area is 0 and the controls don't touch;
  - no ink sits under a control;
  - the controls are 44 px on the phone and 22 px on the desktop;
  - no text line or button straddles the feed's top edge while following, after the card opens, or at rest after three drags of different lengths.
- **Red at the merge base:** with only `ToolCard.tsx` and `LogFeed.tsx` reverted, the test fails on both defects: controls overlapping by 704 px², the output's control starting 16 px above the command's bottom, and lines cut while following and at rest. Green here. Readings are in `evidence/issue-1978/phone.json`.
  - Run: `LLV_SWIPE_BROWSER_TEST=1 CHROME_BIN=google-chrome-stable bun test src/components/mobile/issue1671Evidence.browser.test.tsx -t "#1978"`
- **README demo home on a production build:** before and after frames at 390 and 1280, dark and light, are kept outside the repository for the seat to read.
- **README phone image:** re-captured with `scripts/capture-readme-media.ts --only=phone-conversation`. That mode does not rewrite the provenance record, so the image's `sha256`/`rasterTiles` entry in `docs/media/readme/provenance.json` was updated the way the driver computes them. The driver and seed are unchanged, so their recorded hashes still match.

## Gates

- `tsc --noEmit`: exit 0.
- `bun run build`: exit 0.
- Privacy gate from the merge base, with the known-value fingerprints: PASS.
- Touched test file by path: the `#1978` case passes (the file's other cases stay gated). I also ran the neighbouring LogFeed and ToolCard DOM suites one file each. Three files have failures (`LogFeed.deliveryUncertainty`, `LogFeed.startingWindow`, `ToolCard.imageOutput`), with identical counts at the merge base.

## Trade-off to know

While the feed follows the tail on the phone, up to one line (or one 44 px control) of blank space can sit under the last row. That space is what lets the top edge land on a line boundary without hiding the newest line.


Latand:
## Round 2: the first visible row rests whole

The first round aligned ink, so a row's box could still start above the strip (the review measured 83.75 px against a 110 px edge in the README frame). The edge now settles on **rows**: a feed row, or a row inside one (a run's numbered call, a list item, a tool line). The first row crossing the edge that fits in three quarters of the screen starts at the edge, outer before inner. Ink is only the fallback inside a row too tall to fit.

- **Following the tail:** the feed steps back into the blank under the last row when that covers the cut (at most 40 px). Otherwise a spacer lets the cut row leave whole, together with the frame of the run it closes. The newest line always stays in view; the cost is blank space under the tail, up to one row.
- **Released:** the feed moves the shorter way to a row edge after momentum ends, and again when rows above it resize.
- **Pixel rounding:** Chrome keeps `scrollTop` on whole CSS pixels (100.5 → 101 at DPR 2), while rows sit on half pixels. So moves back round up and moves forward round down. A predecessor showing one pixel or less isn't counted as a visible row, and an edge that falls in a row gap counts as on a boundary. Without these three rules the feed stepped a pixel back and forth forever.
- **One more copy-control case:** a one-line output's 44 px control hung 24 px past its call, into the next call's header. On a coarse pointer, each non-empty output is now framed to hold its control. The desktop is unchanged.

**Proof**
- The browser case now asserts, over a fixture run shaped like the README's:
  - `firstVisibleRow.top >= strip.bottom`, strictly, when following, after opening the run, after three drags, and after closing to the board and coming back;
  - no row shows only a frame sliver under the edge;
  - no call's copy control reaches past its call (phone only; the desktop's 22 px control overhangs a one-line output by 2.6 px today, and the desktop look is kept).
- The case fails every phone reading when run against the reviewed head's feed code, and fails the overhang check without the output frame. It passes here, twice in a row.
- On a production build with the README demo home, dark and light, the first visible row lands at 110–114.75 px (edge 110) with no slivers: following, in the README shot state, after two drags, and after reopening.

**README image:** the phone shot now scrolls its test run to the top of the feed before the frame. Following the tail would otherwise open on the final answer with blank space below, which isn't what the shot describes. Because the shot is off the tail, the phone's existing "down" pill shows in the frame. `provenance.json` carries the new frame hash and the new driver hash. The other four images are unchanged; the only driver change is this shot's `prepare`.

**Gates:** tsc exit 0, `bun run build` exit 0, privacy gate PASS. The neighbouring DOM suites show the same three pre-existing failures as the merge base.

Latand:
## Round 3: the feed settles once inside a result taller than the screen

**Finding (P2, oscillation):** inside a row too tall to align, the text fallback only saw what crossed the edge where the feed stood. Clearing the copy control landed on a text line, and aligning that line cut the control again, so the feed alternated between two positions forever.

**Fix**
- **One clear target:** the fallback now reads every text line and row-sized control in a 96 px band around the edge, and picks the shortest move back and forward that clears all of them at once. Moves stay inside that band and inside the row being read, so the position reached is clear by construction and the next reading finds nothing to do.
  - It only walks subtrees whose boxes reach the band.
  - Each line is trimmed to the overflow containers it sits in, so the hidden overflow of a capped long output (`max-h-[420px] overflow-auto`) never counts.
- **Backstop in LogFeed:** an automatic settle never reverses the previous one, and stops after three in a row. A touch, wheel, key or scrollbar press, or a content resize, starts afresh.

**Proof**
- The browser case opens the fixture's new 60-line result with "show all output", then wheels its first lines under the strip at eight offsets (2 to 55 px). Every stop must come to rest with no line or control crossing the edge.
  - Against the reviewed head (36450cc) it never comes to rest.
  - Here it passes twice in a row, and it still passes with the backstop switched off.
- The test's ink check now clips a line by the element holding it too, as the product does. Before, it flagged the long output's invisible overflow.

**Gates:** tsc exit 0, `bun run build` exit 0, privacy gate PASS. The neighbouring DOM suites show the same three pre-existing failures as the merge base.

